PR #8, opened 2021-09-21
One side of the "alert relation": Loki <- client

Abuelodelanada:


lib/charms/loki_k8s/v0/loki.py
```diff
@@ -4,9 +4,168 @@
...
+    > 0.05
+for: 10m
+labels:
+    severity: page
+annotations:
+    summary: High request latency
+
+```
+
+It is **very important** to note the `%%juju_topology%%` filter in the
```
sed-i, line 131:
```suggestion
It is **critical** to use the `%%juju_topology%%` filter in the
```

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -4,9 +4,168 @@
...
+expression for the alert rule shown above. This filter is a stub that
+is automatically replaced by the `LokiConsumer` following Loki Clients's Juju
```
sed-i, line 133:
```suggestion
is automatically replaced by the `LokiConsumer` following Loki Client's Juju
```

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -4,9 +4,168 @@
...
+rules submitted by one provider charm will also result in
+corresponding alerts for other provider charms. Hence every alert rule
+expression must include such a topology filter stub.
+
+Gathering alert rules and generating rule files within the Loki
+charm is easily done using the `alerts()` method of
+`LokiProvider`. Alerts generated by Loki will
+automatically include Juju topology labels in the alerts. These labels
+indicate the source of the alert. The following lables are
+automatically included with each alert
```
sed-i, line 147:
```suggestion
automatically added to every alert
```

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -4,9 +4,168 @@
...
+
+- `juju_model`
+- `juju_model_uuid`
+- `juju_application`
+
+## Relation Data
+
+The Loki charm uses both application and unit relation data to
+obtain information regarding it Loki Push API and alert rules.
```
sed-i, line 156: stray "it"?

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -4,9 +4,168 @@
...
+
+Units of consumer charm sends their alert rules over app relation
```
sed-i, line 158:
```suggestion
Units of consumer charm send their alert rules over app relation
```

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -20,6 +179,8 @@
 # to 0 if you are raising the major API version
 LIBPATCH = 3
 
+RULES_DIR = "/loki/rules"
```
sed-i, line 182: consider adding docstring that this is in the workload container

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -54,24 +213,45 @@ def __init__(self, charm, relation_name: str):
...
+        into the relation data.
+        Besides Loki generates alert rules files based what
+        consumer charms forwards,
+
+        Args:
+            event: a `CharmEvent` in response to which the Loki
+                charm must update its relation data.
+        """
+
+        if event.relation.data[self.charm.unit].get("data") is None:
```
sed-i, line 235:
```suggestion
        if "data" not in event.relation.data[self.charm.unit].get("data"):
```
although I think it should be safe to just
```python
event.relation.data[self.charm.unit].update({"data": self._loki_push_api})
```
because, iirc, relation_changed won't be fired if the data hasn't changed.

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -54,24 +213,45 @@ def __init__(self, charm, relation_name: str):
...
+            event.relation.data[self.charm.unit]["data"] = self._loki_push_api
+            logger.debug("Saving Loki url in relation data %s", self._loki_push_api)
+
+        if event.relation.data.get(event.relation.app) is not None:
```
sed-i, line 239: Consider documenting rationale.
Should probably be more specific with the `.get`: might add unrelated things to app data in the future that should/shouldn't trigger this.

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -90,6 +270,84 @@ def unit_ip(self) -> str:
             return str(bind_address)
         return ""
 
+    def _remove_alert_rules_files(self, container) -> None:
+        """Remove alert rules files"""
+
+        for f in container.list_files(RULES_DIR):
+            container.remove_path(f.path, recursive=True)
+            logger.debug("Alert rule file '%s' deleted", f.path)
```
sed-i, line 278: Could also be a folder?

```suggestion
            logger.debug("Deleting alert rule item '%s'", f.path)
            container.remove_path(f.path, recursive=True)
```

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -115,22 +375,98 @@ def __init__(self, charm: CharmBase, relation_name: str):
...
                 the Loki charmed service.
+            alert_rules_path: an optional path for the location of alert rules
+                files.  Defaults to "src/loki_alert_rules" at the top level
+                of the charm repository.
         """
         super().__init__(charm, relation_name)
         self._stored.set_default(loki_push_api=None)
         self._charm = charm
         self._relation_name = relation_name
+        self._ALERT_RULES_PATH = alert_rules_path
```
sed-i, line 386: why all caps?

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -115,22 +375,98 @@ def __init__(self, charm: CharmBase, relation_name: str):
...
+                charm must update its relation data.
+        """
         if event.unit is None:
             # Workaround: Seems this is a Juju bug that sends event.unit == None
             return
 
         if data := event.relation.data[event.unit].get("data"):
             self._stored.loki_push_api = json.loads(data)["loki_push_api"]
 
+        event.relation.data[self._charm.app]["metadata"] = json.dumps(self._scrape_metadata)
```
sed-i, line 407: leader guard?

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -115,22 +375,98 @@ def __init__(self, charm: CharmBase, relation_name: str):
...
+        self._set_alert_rules(event)
+
+    def _set_alert_rules(self, event):
+        """Set alert rules into relation data.
+
+        Args:
+            event: a `CharmEvent` in response to which the consumer
+                charm must update its relation data.
+        """
+        event.relation.data[self._charm.unit]["loki_host"] = str(
```
sed-i, line 417: Could you please document (if haven't already) what goes into app data and what into unit data?

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -115,22 +375,98 @@ def __init__(self, charm: CharmBase, relation_name: str):
...
+        metadata = self._scrape_metadata
+        topology = 'juju_model="{}", juju_model_uuid="{}", juju_application="{}"'.format(
+            metadata["model"], metadata["model_uuid"], metadata["application"]
+        )
+
+        if expr := rule.get("expr", None):
+            expr = expr.replace("%%juju_topology%%", topology)
+            rule["expr"] = expr
+        else:
+            logger.error("Invalid alert expression in %s", rule.get("alert"))
```
sed-i, line 466: what should be returned in this case?

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -139,3 +475,55 @@ def loki_push_api(self):
...
+            logger.debug("Reading alert rule from %s", path)
+            with path.open() as rule_file:
+                # Load a list of rules from file then add labels and filters
+                try:
+                    rule = yaml.safe_load(rule_file)
+                    rule = self._label_alert_topology(rule)
+                    rule = self._label_alert_expression(rule)
+                    alerts.append(rule)
+                except Exception as e:
+                    logger.error("Failed to read alert rules from %s: %s", path.name, str(e))
```
sed-i, line 505: Not sure if that is what you meant:
```suggestion
                try:
                    rule = yaml.safe_load(rule_file)
                except Exception as e:
                    logger.error("Failed to read alert rules from %s: %s", path.name, str(e))
                else:
                    rule = self._label_alert_topology(rule)
                    rule = self._label_alert_expression(rule)
                    alerts.append(rule)
```

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -139,3 +475,55 @@ def loki_push_api(self):
...
+            metadata = self._scrape_metadata
+            group = {
+                "name": "{model}_{model_uuid}_{application}_alerts".format(**metadata),
+                "rules": alerts,
+            }
+            groups.append(group)
+        return groups
+
+    @property
+    def _scrape_metadata(self) -> dict:
```
sed-i, line 518: we should probably come up with a `class` for topology

sed-i (commented):
I don't fully understand the architecture so just some superficial comments

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -4,9 +4,168 @@
...
+            super().__init__(*args)
+            ...
+            self._provide_loki()
+            ...
+
+        def _provide_loki(self):
+            loki_server = LokiServer()
+
+            if loki_server.is_ready:
+                version = loki_server.version
```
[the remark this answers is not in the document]
Abuelodelanada: Sure, I forgot that!

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -54,24 +213,45 @@ def __init__(self, charm, relation_name: str):
...
+        """Removes alert rules files when consumer charms
+        left the relation with Loki
+
+        Args:
+            event: a `CharmEvent` in response to which the Loki
+                charm must update its relation data.
         """
-        event.relation.data[self.charm.unit]["data"] = self._loki_push_api
-        logger.debug("Saving Loki url in relation data %s", self._loki_push_api)
+        if event.relation.data.get(event.relation.app) is not None:
```
balbirthomas, line 252: Are there values for relation data that can be false in a boolean context but are not `None` ? If there are no such values then you could choose to remove the `is not None` clause.

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -90,6 +270,84 @@ def unit_ip(self) -> str:
             return str(bind_address)
         return ""
 
+    def _remove_alert_rules_files(self, container) -> None:
+        """Remove alert rules files"""
+
+        for f in container.list_files(RULES_DIR):
```
balbirthomas, line 276: As discussed on Mattermost this iteration to remove each file is only necessary because you are using a volume mount. So I presume you have made an explicit decision to do so. Is there a specific reason for this ? Not data is already persisted in the relation.

balbirthomas (approved):

lib/charms/loki_k8s/v0/loki.py
Abuelodelanada: No should be folders there. Fixed

lib/charms/loki_k8s/v0/loki.py
Abuelodelanada: Just because inside the object is a constant

lib/charms/loki_k8s/v0/loki.py
Abuelodelanada: Good catch!

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -4,9 +4,168 @@
...
+  its clients. This relation must confirm to the
+  `loki_push_api` interface.
+
+
+## Alerting Rules
+
+This charm library also supports gathering alerting rules from all
+related Loki clients charms and enabling corresponding alerts within the
+Loki charm. Alert rules are automatically gathered by `LokiConsumer` object
+from a directory conventionally named `loki_alert_rules`.
```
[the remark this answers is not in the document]
Abuelodelanada: What are you thinking about @sed-i ?
Having a shared module for this?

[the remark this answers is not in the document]
sed-i: I propose to organize all the rules in a subdir to reduce clutter in the charm root folder. We're probably going to have more charms with more rules, so might be handy to group them all under e.g. `rules/whatever`.

lib/charms/loki_k8s/v0/loki.py
Abuelodelanada: This relation data is not really necessary. Removed.

PR #9, opened 2021-09-21
Patch the Kubernetes service on install

mmanciop:
Ensure that the `loki` service is correctly reporting port `3100/TCP` instead of the default `65535/TCP`.


Abuelodelanada:
Push to PR #8

PR #11, opened 2021-09-24
Implement relation between loki and alertmanager

Abuelodelanada:
This PR is a WIP.

Tasks:

- [x] Add relation with Alertmanager
- [x] Refactor charm code adding a catch_all pattern
- [x] Build loki config using `/etc/loki/local-config.yaml` as a base
- [x] Modify documentation
- [x] Remove "sharp edges"
- [x] Add proper docstrings 
- [x] Fix failing tests
- [x] Add unit tests

### DIFF
--- a/INTEGRATING.md
+++ b/INTEGRATING.md
@@ -1,3 +1,34 @@
 ## Integrating Loki
 
-Loki integrates with any charm that supports the `logging` interface. This integration is done using the Loki Charm library. The documentation of the charm library, available through [Charmhub](https://charmhub.io/loki-k8s) provides further details on how this integration may be done.
+Loki integrates with the following:
+
+
+1. Any charm that supports the `loki_push_api` interface.
+This integration is done using the Loki Charm library. The documentation of
+the charm library, available through [Charmhub](https://charmhub.io/loki-k8s)
+provides further details. The information
+exchanged through the `loki_push` interface can be broken down
+into two parts
+
+
+- Loki charm provides an [endpoint URL](https://grafana.com/docs/loki/latest/api/#post-lokiapiv1push)
+to receive log from [Loki clients](https://grafana.com/docs/loki/latest/clients/) that
+relates with this charm.
+
+- Loki may receive alert rules which tell when to raise alerts. These
+  rules are read from a directory named `loki_alert_rules`, if
+  present at the top level, within the client charm's source
+  (`src`) directory.
+
+Any time new relations are made, or existing ones changed, between the
+Loki charm and any scrape target charm, a new Loki
+configuration is generated.
+
+2. Loki integrates with
+[Grafana](https://charmhub.io/grafana-k8s) which provides a dashboard
+for viewing logs aggregated by Loki. These dasboards may be
+customised by charms that relate to Grafana.
+
+3. Loki forwards alerts to one or more
+[Alertmanagers](https://charmhub.io/alertmanager-k8s) that are related
+to it.
--- a/INTEGRATING.md
+++ b/INTEGRATING.md
@@ -4,7 +4,7 @@ Loki integrates with the following:
 
 
 1. Any charm that supports the `loki_push_api` interface.
-This integration is done using the Loki Charm library. The documentation of
+You can integrate it using the Loki Charm library. The documentation of
 the charm library, available through [Charmhub](https://charmhub.io/loki-k8s)
 provides further details. The information
 exchanged through the `loki_push` interface can be broken down
--- a/README.md
+++ b/README.md
@@ -31,7 +31,18 @@ Note the `--destroy-storage` will delete any data stored by MySQL in its persist
 
 ## Relations
 
-This charm provides a `loki-push-api` relation so you can integrate this charm with others charms that requires a Loki logging stack.
+Currently supported relations are:
+
+- [Grafana](https://github.com/canonical/grafana-operator) aggregates
+  logs obtained by Loki and provides a versatile dashboard to
+  view these logs in configurable ways. Loki relates to
+  Grafana over the `grafana_datasource` interface.
+- [Alertmanager](https://github.com/canonical/alertmanager-operator)
+  receives alerts from Loki, aggregates and deduplicates them,
+  then forwards them to specified targets. Loki relates to
+  Alertmanager over the `alertmanager` interface.
+- In addition, this Loki charm allows relations with any
+  charm that supports the `loki_push_api` relation.
 
 
 ## OCI Images
--- a/README.md
+++ b/README.md
@@ -40,9 +40,9 @@ Currently supported relations are:
 - [Alertmanager](https://github.com/canonical/alertmanager-operator)
   receives alerts from Loki, aggregates and deduplicates them,
   then forwards them to specified targets. Loki relates to
-  Alertmanager over the `alertmanager` interface.
+  Alertmanager over the `alertmanager_dispatch` interface.
 - In addition, this Loki charm allows relations with any
-  charm that supports the `loki_push_api` relation.
+  charm that supports the `loki_push_api` relation interface.
 
 
 ## OCI Images
--- a/lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
+++ b/lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
@@ -1,0 +1,257 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""# Alertmanager library.
+
+This library is designed to be used by a charm consuming or providing the `alerting` relation.
+
+This library is published as part of the
+[Alertmanager charm](https://charmhub.io/alertmanager-k8s).
+
+You can file bugs [here](https://github.com/canonical/alertmanager-operator/issues)!
+
+A typical example of including this library might be:
+
+```python
+# ...
+from charms.alertmanager_k8s.v0.alertmanager_dispatch import AlertmanagerConsumer
+
+class SomeApplication(CharmBase):
+  def __init__(self, *args):
+    # ...
+    self.alertmanager_consumer = AlertmanagerConsumer(self, relation_name="alertmanager")
+    # ...
+```
+"""
+
+import logging
+from typing import List
+
+import ops
+from ops.charm import CharmBase, RelationEvent, RelationJoinedEvent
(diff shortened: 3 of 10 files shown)


README.md
```diff
@@ -31,7 +31,18 @@ Note the `--destroy-storage` will delete any data stored by MySQL in its persist
...
+Currently supported relations are:
+
+- [Grafana](https://github.com/canonical/grafana-operator) aggregates
+  logs obtained by Loki and provides a versatile dashboard to
+  view these logs in configurable ways. Loki relates to
+  Grafana over the `grafana_datasource` interface.
+- [Alertmanager](https://github.com/canonical/alertmanager-operator)
+  receives alerts from Loki, aggregates and deduplicates them,
+  then forwards them to specified targets. Loki relates to
+  Alertmanager over the `alertmanager` interface.
```
sed-i, line 43: If you really mean relation _interface_ then 
```suggestion
  Alertmanager over the `alertmanager_dispach` interface.
```

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -45,13 +45,15 @@ def __init__(self, *args):
...
         def _provide_loki(self):
-            loki_server = LokiServer()
-
-            if loki_server.is_ready:
-                version = loki_server.version
+            try:
+                version = self._loki_server.version
                 self.loki_provider = LokiProvider(self, "logging")
-                self.unit.status = ActiveStatus()
+                self._stored.provider_ready = True
```
sed-i, line 51: Not sure why you need this if it is equivalent to `if self._loki_server.version` or similar.

src/charm.py
```diff
@@ -57,44 +67,90 @@ def _on_install(self, _):
...
     def _on_loki_pebble_ready(self, event):
-        """Define and start a workload using the Pebble API."""
-        # Get a reference the container attribute on the PebbleReadyEvent
-        container = event.workload
-        # Define an initial Pebble layer configuration
+        self._configure(event)
+
+    def _on_alertmanager_change(self, event):
+        self._configure(event)
+
```
sed-i, line 78: on upgrade should probably also call _configure, and possibly manipulate self._stored.config as well.

src/charm.py
```diff
@@ -57,44 +67,90 @@ def _on_install(self, _):
...
-        container.add_layer("loki", pebble_layer, combine=True)
-        # Autostart any services that were defined with startup: enabled
-        container.autostart()
+
+        return pebble_layer
 
     ##############################################
     #             UTILITY METHODS                #
     ##############################################
     def _provide_loki(self):
```
sed-i, line 143: seems to overlap `_provide_loki` in `loki.py`

lib/charms/loki_k8s/v0/loki.py
Abuelodelanada: No, it's no necessary, it's statement from previous refactor logic.

src/charm.py
```diff
@@ -57,44 +67,90 @@ def _on_install(self, _):
...
+        new_layer = self._build_pebble_layer
+
+        if current_layer != new_layer:
+            restart = True
+
+        config = self._loki_config()
+
+        if yaml.safe_load(self._stored.config) != config:
+            self._stored.config = yaml.dump(config)
+            self._container.push(LOKI_CONFIG, self._stored.config)
```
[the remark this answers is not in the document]
Abuelodelanada: Good catch!

src/charm.py
sed-i: I see it here:
https://github.com/canonical/loki-operator/blob/73a363604231f951aee35eb452d9c6089e7b8d85/lib/charms/loki_k8s/v0/loki.py#L47

INTEGRATING.md
```diff
@@ -1,3 +1,34 @@
 ## Integrating Loki
 
-Loki integrates with any charm that supports the `logging` interface. This integration is done using the Loki Charm library. The documentation of the charm library, available through [Charmhub](https://charmhub.io/loki-k8s) provides further details on how this integration may be done.
+Loki integrates with the following:
```
balbirthomas, line 3: Nice writeup.

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
```diff
@@ -0,0 +1,257 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""# Alertmanager library.
+
+This library is designed to be used by a charm consuming or providing the `alerting` relation.
```
balbirthomas, line 6: Is that "alerting" relation or "alerting" interface. Libraries define the behaviour of relation interfaces. The relation names themselves are arbitrary.

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
```diff
@@ -0,0 +1,257 @@
...
+        )
+        self.framework.observe(
+            self.charm.on[self.name].relation_departed,
+            self._on_relation_departed,
+        )
+        self.framework.observe(self.charm.on[self.name].relation_broken, self._on_relation_broken)
+
+    def _on_relation_changed(self, event: ops.charm.RelationChangedEvent):
+        """This hook notifies the charm that there may have been changes to the cluster."""
+        if event.unit:  # event.unit may be `None` in the case of app data change
```
balbirthomas, line 138: Is there any reason not to emit a cluster changed event when application data for the cluster changes ?

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
```diff
@@ -0,0 +1,257 @@
...
+        This function can be used in two different ways:
+        - update relation data bag of a given event (e.g. a newly joined relation);
+        - update relation data for all relations
+
+        Args:
+            event: The event whose data bag needs to be updated. If it is None, update data bags of
+            all relations.
+        """
+        if event is None:
+            # update all existing relation data
```
balbirthomas, line 248: I am not sure why you need to do these two checks `event is None` and `self.name in self.charm.model.relations`. Could you not just iterate over relations of a specific name as in `for relation in self.charm.model.relations[<relation-name>]` ?

src/charm.py
```diff
@@ -57,44 +68,93 @@ def _on_install(self, _):
...
+        current_layer = self._container.get_plan().services
+        new_layer = self._build_pebble_layer
+
+        if current_layer != new_layer:
+            restart = True
+
+        config = self._loki_config()
+
+        try:
+            if yaml.safe_load(self._stored.config) != config:
```
balbirthomas, line 97: I am not sure why you bother to check if the current config is different from the saved one. Why not just always push the current config and simplify your code flow ?

src/charm.py
```diff
@@ -57,44 +68,93 @@ def _on_install(self, _):
...
+                self._stored.config = yaml.dump(config)
+                self._container.push(LOKI_CONFIG, self._stored.config)
+                logger.info("Pushed new configuration")
+                restart = True
+        except (ProtocolError, PathError) as e:
+            self.unit.status = BlockedStatus(str(e))
+            return False
+
+        if restart:
```
balbirthomas, line 106: Do you always need to restart. Does loki allow reload ?

balbirthomas (approved):

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
Abuelodelanada: @sed-i seems the Alertmanager library docstring is outdated.

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
Abuelodelanada: @sed-i 👆

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
Abuelodelanada: @sed-i 👆

src/charm.py
Abuelodelanada: I am checking whether the current config is different from the saved one to restart the container by setting `restart = True`

Do you think there is a better approach here?

src/charm.py
Abuelodelanada: Loki do not has a way to **on-demand** reload its config.

It has the concept of [`Runtime Configuration file`](https://grafana.com/docs/loki/latest/configuration/#runtime-configuration-file) in which you can place certain configs that Loki will reload while running.

src/charm.py
Abuelodelanada: It's the docstring ;-)

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
sed-i: Originally I was indeed referring to the relation name. [Changed](https://github.com/canonical/alertmanager-operator/pull/34/commits/58e733fafa1ccf2ec9e0ba3406f57035cfccb2c5) to interface name.

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
sed-i: In the general case we could. In the context of alertmanager, "cluster changed" is only sensible for changes in unit data. At least for now. The `if event.unit` reduces false positives / event storm. The `if event.unit` is not intended for the entire `_on_relation_changed` hook, but rather for `on.cluster_changed.emit()` only.

lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py
sed-i: iirc the `if self.name in relations` is there because `self.charm.model.relations[self.name]` on its own occasionally raised KeyError. Can't think of another reason I would have had it there for.

src/charm.py
sed-i: I was probably up for too long at the time of writing :D

src/charm.py
Abuelodelanada: See #13

src/charm.py
Abuelodelanada: See #14

src/charm.py
Abuelodelanada: See #15

PR #18, opened 2021-10-04
Implement upgrade-charm event handler

Abuelodelanada:


src/charm.py
```diff
@@ -71,6 +72,10 @@ def _on_install(self, _):
     def _on_config_changed(self, event):
         self._configure(event)
 
+    def _on_upgrade_charm(self, event):
+        self._patch_k8s_service()
+        self._configure(event)
```
sed-i, line 77: Note that this hook is simple thanks to pulling the config every time.
In alertmanager I store the hash and because of that my upgrade hook is more involved.
Just pointing out :)

sed-i (approved):

PR #19, opened 2021-10-05
Adding pyproject.toml, new tox settings (and linting everything)

Abuelodelanada:


lib/charms/loki_k8s/v0/loki.py
```diff
@@ -4,7 +4,7 @@
 #
 # Learn more at: https://juju.is/docs/sdk
 
-"""## Overview.
+r"""## Overview.
```
sed-i, line 7: Why the `r`? Interesting!

lib/charms/loki_k8s/v0/loki.py
```diff
@@ -417,15 +414,13 @@ def alerts(self) -> dict:
...
-    Loki Consumer class
-    """
+    """Loki Consumer class."""
 
     _stored = StoredState()
     _ALERT_RULES_PATH: str
 
     def __init__(
-        self, charm: CharmBase, relation_name: str, alert_rules_path="src/loki_alert_rules"
+        self, charm: CharmBase, relation_name: str, alert_rules_path: str = "src/loki_alert_rules"
```
sed-i, line 423: consider adding a default for `relation_name`

sed-i (approved):
See also changes I made in https://github.com/canonical/alertmanager-operator/pull/39

lib/charms/loki_k8s/v0/loki.py
Abuelodelanada: Because we are using backslashes in the docstring:
```
lib/charms/loki_k8s/v0/loki.py:7:1: D301 Use r""" if any backslashes in a docstring
```

lib/charms/loki_k8s/v0/loki.py
Abuelodelanada: ack

PR #24, opened 2021-10-07
Firsts Integration tests

Abuelodelanada:
Based on Leon's Alertmanaget tests

### DIFF
--- a/tests/integration/test_upgrade_charm.py
+++ b/tests/integration/test_upgrade_charm.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+
+import json
+import logging
+import urllib.request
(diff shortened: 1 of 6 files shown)


tests/integration/test_upgrade_charm.py
```diff
@@ -0,0 +1,55 @@
...
+
+import pytest
+import yaml
+from helpers import get_unit_address  # type: ignore[import]
+
+log = logging.getLogger(__name__)
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+# app_name = "am"
+app_name = METADATA["name"]
```
sed-i, line 19: app_name is optional in the args below; it will be loki-k8s by default.
Here you can provide a custom name
```suggestion
app_name = "loki"
```

sed-i (approved):

dstathis (approved):

PR #25, opened 2021-10-12
Make relation_name optional in lib, improve alert dir resolution

Abuelodelanada:
Fixing merge conflics in @mmanciop #17, lint and static analysis.

Align behavior of the Loki library to that of Prometheus, see https://github.com/canonical/prometheus-operator/pull/135

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -708,6 +708,8 @@ class LokiPushApiConsumer(RelationManagerBase):
 
         if event.unit is None:
             # Workaround: Seems this is a Juju bug that sends event.unit == None
+            # Remove this if when this issue is closed:
+            # https://github.com/canonical/loki-operator/issues/3
             return
 
         if data := event.relation.data[event.unit].get("data"):


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -258,15 +360,42 @@ def __init__(self, message="Alert rule is not well formed"):
 class LokiProvider(RelationManagerBase):
     """A LokiProvider class."""
 
-    def __init__(self, charm, relation_name: str):
+    def __init__(self, charm, relation_name: Optional[str] = None):
```
mmanciop, line 363: Better to add the default relation name as argument default and have the argument type of relation_name to be `str`, easier to understand

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -417,13 +546,13 @@ class LokiConsumer(RelationManagerBase):
...
     _stored = StoredState()
-    _ALERT_RULES_PATH: str
+    _alert_rules_path: Optional[str]
 
     def __init__(
         self,
         charm: CharmBase,
-        relation_name: str = "logging",
-        alert_rules_path: str = "src/loki_alert_rules",
+        relation_name: Optional[str] = None,
```
mmanciop, line 554: use default argument values rather than Optional[str]

mmanciop (changes requested):
The pattern evolved a bit, check out the latest Grafana and Prometheus libs. Also, check the content of the Grafana docu, I like that a lot in terms of which info we provide.

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i, line 363: I would add a default for the relation and remove `_get_single_relation_by_interface`.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -231,6 +241,98 @@ def __init__(self, *args):
...
 logger = logging.getLogger(__name__)
 
+RELATION_INTERFACE_NAME = "loki_push_api"
+
+
+class RelationDirection(Enum):
+    """Relations can be either provided or required by a charm."""
+
+    PROVIDED = 1
+    REQUIRED = 2
```
sed-i, line 251: This semantic already exists in `ops.charm.RelationRole`.
See alertmanager impl for reference
https://github.com/canonical/alertmanager-operator/blob/e7b4634b15b568fa96d2c07141d107ab801fce79/lib/charms/alertmanager_k8s/v0/alertmanager_dispatch.py#L81

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -437,20 +566,46 @@ def __init__(
...
+                )
+            except MultipleRelationsWithInterfaceFoundError as e:
+                raise ModelError(
+                    f"Multiple required relations with the '{RELATION_INTERFACE_NAME}' interface found: "
+                    f"{e.relations}; you must specify which relation should be managed by this "
+                    f"{type(self)} by providing the `relation_name` argument."
+                )
+
+        if not alert_rules_path:
+            alert_rules_path = resolve_dir_against_main_path("loki_alert_rules")
```
sed-i, line 602: - I would have "loki_alert_rules" as the default instead of None.
- `resolve_dir_against_main_path` should probably be called regardless. See also https://github.com/canonical/prometheus-operator/pull/141

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -574,35 +729,35 @@ def _labeled_alert_groups(self) -> list:
         """
         alerts = []
 
-        for path in Path(self._ALERT_RULES_PATH).glob("*.rule"):
-            if not path.is_file():
-                continue
+        if self._alert_rules_path:
+            for path in Path(self._alert_rules_path).glob("*.rule"):
```
sed-i, line 733: I would consider delegating this concern to a separate class, for example
```python
alerts = list(map(LokiAlertRule.from_file, Path(self._alert_rules_path).glob("*.rule")))
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -574,35 +729,35 @@ def _labeled_alert_groups(self) -> list:
...
             metadata = self._scrape_metadata
-            group = {
-                "name": "{model}_{model_uuid}_{application}_alerts".format(**metadata),
-                "rules": alerts,
-            }
-            groups.append(group)
-        return groups
+            return [
+                {
+                    "name": "{model}_{model_uuid}_{application}_alerts".format(**metadata),
```
sed-i, line 755: Consider generating topology info with a dedicated function/class.
```python
f"{JujuTopology(charm).as_str()}_alerts"
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -611,9 +766,8 @@ def _scrape_metadata(self) -> dict:
         Returns:
             Scrape configutation metadata for this logging provider charm.
         """
-        metadata = {
+        return {
```
sed-i, line 769: Consider generating topology info with a dedicated function/class.
```python
return JujuTopology(charm).as_dict()  # TypedDict
```

sed-i (changes requested):
Looks good. Should use `ops.charm.RelationRole`.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -224,13 +232,194 @@ def __init__(self, *args):
...
+LIBPATCH = 4
 
 # Alert rules directory in workload container
 RULES_DIR = "/loki/rules"
 
 logger = logging.getLogger(__name__)
 
+RELATION_INTERFACE_NAME = "loki_push_api"
+DEFAULT_RELATION_NAME = "logging"
+DEFAULT_ALERT_RULES_RELATIVE_PATH = "./src/loki_alert_rules"
```
sed-i, line 244: Add docstring explaining this is rel to the charm's root?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -224,13 +232,194 @@ def __init__(self, *args):
...
+
+
+class RelationNotFoundError(Exception):
```
sed-i, line 247: Consider inheriting from ValueError or RuntimeError. Applies to the ones below as well.

sed-i (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -24,9 +24,13 @@
...
-- Name of the relation that the Loki charm uses to interact with
-  its clients. This relation name must match the one used in `metadata.yaml`
-  for the `loki_push_api` interface.
+- Optionally, the name of the relation that the Loki charm uses to interact
+  with its clients. If provided, this relation name must match a provided
+  relation in metadata.yaml with the `loki_push_api` interface.
+  This argument is not required if your metadata.yaml has precisely one
+  provided relation in metadata.yaml with the `loki_push_api` interface, as the
+  lib will automatically resolve the relation name inspecting the using the
+  meta information of the charm.
```
mmanciop, line 33: Need to show what to put in the metadata.yaml

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -134,26 +138,30 @@ class LokiClientCharm(CharmBase):
         def __init__(self, *args):
             super().__init__(*args)
             ...
-            self.loki = LokiConsumer(self, "logging")
```
mmanciop, line 137: Why `loki` and not `loki_consumer`?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -258,15 +448,32 @@ def __init__(self, message="Alert rule is not well formed"):
 class LokiProvider(RelationManagerBase):
```
mmanciop, line 448: This needs to be updated to `LokiPushApiProvider`

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -417,13 +624,13 @@ class LokiConsumer(RelationManagerBase):
     """Loki Consumer class."""
```
mmanciop, line 624: I think this needs to be updated to `LokiPushApiConsumer`

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -432,25 +639,51 @@ def __init__(
...
+            alert_rules_path = _resolve_dir_against_charm_path(charm, alert_rules_path)
+        except InvalidAlertRuleFolderPathError as e:
+            logger.warning(
+                "Invalid Prometheus alert rules folder at %s: %s",
+                e.alert_rules_absolute_path,
+                e.message,
+            )
+
         super().__init__(charm, relation_name)
         self._stored.set_default(loki_push_api=None)
```
mmanciop, line 683: If possible, I'd avoid using storage inside a library.

mmanciop (changes requested):

PR #30, opened 2021-10-19
Custom events created

Abuelodelanada:
In this PR we created 
- `LokiDepartedEvent`: Fired when the relation is removed.
- `LokiPushApiEndpointJoined`: Fired when the relation is fully established

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -485,13 +485,9 @@ class AlertRuleError(Exception):
 class LokiPushApiEndpointDeparted(EventBase):
     """Event emitted when Loki departed."""
 
-    pass
-
 
 class LokiPushApiEndpointJoined(EventBase):
     """Event emitted when Loki joined."""
-
-    pass
 
 
 class LoggingEvents(ObjectEvents):


mmanciop:
@Abuelodelanada I am wondering why we have the Departed event and not also its corresponding Joined event. What is the rationale?

mmanciop (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -457,6 +482,25 @@ def __init__(self, message="Alert rule is not well formed"):
         super().__init__(self.message)
 
 
+class LokiPushApiEndpointDeparted(EventBase):
+    """Event emitted when Loki departed."""
+
+    pass
```
sed-i, line 488: `pass` is redundant when you already have a docstring.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -457,6 +482,25 @@ def __init__(self, message="Alert rule is not well formed"):
...
+
+
+class LokiPushApiEndpointJoined(EventBase):
+    """Event emitted when Loki joined."""
+
+    pass
```
sed-i, line 494: `pass` is redundant when you already have a docstring.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -166,6 +166,31 @@ def __init__(self, *args):
   lib will automatically resolve the relation name inspecting the using the
   meta information of the charm
 
+Anytime the relation between a Loki provider charm and a Loki consumer charm is
+established a `LokiPushApiEndpointJoined` event is fired. In the consumer side
+is it possible to observe this event with:
+
```
sed-i, line 172: I'm curious why should charm code care about these events. Maybe worth adding to docstring?

sed-i (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: Thanks, I forgot to deleted!

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: Thanks, I forgot to deleted!

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: We need to add or remove configuration in the consumer side (grafana-agent-operator).
Obviously docstrings can be improved!
Take a look at OP-369

PR #31, opened 2021-11-08
foundations for free-standing rules

sed-i:
This PR adds the necessary functionality needed by [lma-rules-operator](https://github.com/canonical/lma-rules-operator) to forward free standing rules.
Similar changes have been [made to prometheus-operator](https://github.com/canonical/prometheus-operator/pull/149).

- [x] add JujuTopology class
- [x] add generic load_alert_rules_from_dir

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -556,11 +556,12 @@ def load_alert_rules_from_dir(
         relpath = os.path.relpath(os.path.dirname(path), dir_path)
 
         # Generate group name:
-        #  - prefix, from the relative path of the rule file;
         #  - name, from juju topology
+        #  - suffix, from the relative path of the rule file;
         return (
+            f"{topology.identifier}_"
             f"{'' if relpath == '.' else relpath.replace(os.path.sep, '_') + '_'}"
-            f"{topology.identifier}_alerts"
+            "alerts"
         )
 
     invalid_files = []
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -334,20 +334,6 @@ class RelationRoleMismatchError(Exception):
         super().__init__(self.message)
 
 
-class InvalidAlertRuleFolderPathError(Exception):
-    """Raised if the alert rules folder cannot be found or is otherwise invalid."""
-
-    def __init__(
-        self,
-        alert_rules_absolute_path: str,
-        message: str,
-    ):
-        self.alert_rules_absolute_path = alert_rules_absolute_path
-        self.message = message
-
-        super().__init__(self.message)
-
-
 def _validate_relation_by_interface_and_direction(
     charm: CharmBase,
     relation_name: str,
@@ -585,8 +571,8 @@ def _resolve_dir_against_charm_path(charm: CharmBase, *path_elements: str) -> st
 
     Look up the directory of the main .py file being executed. This is normally
     going to be the charm.py file of the charm including this library. Then, resolve
-    the provided path elements and, if the result path exists and is a directory,
-    return its absolute path; otherwise, return `None`.
+    the provided path elements and return its absolute path, without checking for existence or
+     validity.
     """
     charm_dir = Path(charm.charm_dir)
     if not charm_dir.exists() or not charm_dir.is_dir():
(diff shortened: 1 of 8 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -402,6 +403,109 @@ def _validate_relation_by_interface_and_direction(
...
+    def scrape_identifier(self):
+        """Format the topology information into a scrape identifier."""
+        return "juju_{}_{}_{}".format(
+            self.model,
+            self.short_model_uuid,
+            self.application,
+        )
+
+    @property
+    def promql_labels(self) -> str:
```
Abuelodelanada, line 482: Promql?
In Loki we have [LogQL](https://grafana.com/docs/loki/latest/logql/) ;-)

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -402,6 +403,109 @@ def _validate_relation_by_interface_and_direction(
...
+        )
+
+    def as_dict(self, short_uuid=False) -> dict:
+        """Format the topology information into a dict."""
+        as_dict = dataclasses.asdict(self)
+        if short_uuid:
+            as_dict["model_uuid"] = self.short_model_uuid
+        return as_dict
+
+    def as_dict_with_promql_labels(self):
```
Abuelodelanada, line 495: Same here: LogQL ;-)

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -402,6 +403,109 @@ def _validate_relation_by_interface_and_direction(
...
+        return {
+            "juju_model": self.model,
+            "juju_model_uuid": self.model_uuid,
+            "juju_application": self.application,
+            "juju_charm": self.charm_name,
+        }
+
+    def render(self, template: str):
+        """Render a juju-topology template string with topology info."""
+        return template.replace("%%juju_topology%%", self.promql_labels)
```
Abuelodelanada, line 506: promql_labels -> logql_labels

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -798,11 +901,8 @@ def _label_alert_topology(self, rule) -> dict:
             a dictionary representing Loki alert rule with Juju
             topology labels.
         """
-        metadata = self._scrape_metadata
         labels = rule.get("labels", {})
-        labels["juju_model"] = metadata["model"]
-        labels["juju_model_uuid"] = metadata["model_uuid"]
-        labels["juju_application"] = metadata["application"]
+        labels.update(self.topology.as_dict_with_promql_labels())
```
Abuelodelanada, line 905: as_dict_with_promql_labels -> as_dict_with_logql_labels

Abuelodelanada (approved):
I like it!

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i: Good catch! I think we should probably rename this method both here and in prometheus_scrape: the labels are identical here and [there](https://github.com/canonical/prometheus-operator/blob/ea351276712fd54c45635181eb5296e39274bff1/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L533).

Maybe this `juju_...` concept should be named `juju_labels` instead of `promql_labels`? @mmanciop @balbirthomas

mmanciop: in this lib it could be `logql_labels`

mmanciop (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
simskij: While really similar, the duplication is incidental, meaning that the code might have different reasons to change in the future. I'd prefer us to use a descriptive, clear name over finding one that works in both projects.

sed-i: The individual keys `juju_model`, `juju_model_uuid` etc. are probably generic and not prom/log ql specific.
Do you mean that the composite syntax "key1=val1, key2=val2" is coincidentally the same between promql and logql? And that the concern of a future change in that syntax justifies having two different concept names here?

mmanciop: Scrap the probably: we _want_ them to stay stable, so that we can build links and dashboards to go from a data silo to another (e.g., logs to metrics to distributed traces).

sed-i: Good. So in the context of the JujuTopology class, do we need to differentiate between prom and loki as far as the composite string of those labels is called? Any reason we shouldn't rename here and in prom to simply `ql_labels`?

simskij: > we want them to be entirely consistent across out telemetry

Definitely, the labels themselves, or the output, should be consistent across all our data silos.

> Any reason we shouldn't rename here and in prom to simply `ql_labels`

What is considered a `ql`? Is SQL a `ql`? NoSQL? Does it add any additional meaning to prefix it with something so vague? It kind of makes sense to be specific, either using the full `ql` it is intended for, or by describing what it is (ie. `juju_topology_labels`) don't you think?

sed-i: What it really is, is `comma_separated_labels` :)

Abuelodelanada (approved):

PR #32, opened 2021-11-12
push full config instead of pulling and updating

sed-i:
This PR proposes to:
1. Provide the minimal loki config explicitly, instead of pulling the one that is already in the container and updating it.
2. Have the rules dir passed to loki lib from the charm.
3. Add the "fake" subdir to the rules dir ([ref](https://grafana.com/docs/loki/latest/operations/storage/filesystem/)).
4. Update `_stored.config` only after `push` succeeds.


simskij (approved):

Abuelodelanada (approved):
Based on what we talked yesterday looks good to me.
Did you deploy the charm using Loki 2.4.1? If everything goes fine, we should upgrade the image resource in charmhub

PR #34, opened 2021-11-15
fix rules file path + upkeep

sed-i:
This is a bugfix + upkeep PR:
- bugfix: add missing `.identifier` to rules filename. Before this change, filename was e.g. `JujuTopology(model='lma', model_uuid='ea95a2e', application='lma-rules-k8s', charm_name='lma-rules-k8s')_rel_6_alert.rules`. Bug introduced in my previous PR.
- add version tag to loki image
- create "tenant" dir and patch unittests that end up calling pebble's make_dir, until https://github.com/canonical/operator/pull/645 is merged.

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -674,6 +674,8 @@ class LokiPushApiProvider(RelationManagerBase):
                 deploying your charm will have a consistent experience with all
                 other charms that consume metrics endpoints.
 
+            rules_dir: path in workload container where rule files are to be stored.
+
         Raises:
             RelationNotFoundError: If there is no relation in the charm's metadata.yaml
                 with the same name as provided via `relation_name` argument.
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -272,7 +272,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 5
+LIBPATCH = 6
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -690,14 +690,17 @@ def __init__(
         super().__init__(charm, relation_name)
         self.charm = charm
         self._relation_name = relation_name
+        self.container = self.charm.unit.get_container("loki")
```
Abuelodelanada, line 693: In the charm.py `__init__` method you have:

```python
self._container = self.unit.get_container(self._name)
```

So you can access it from the `LokiPushApiProvider` object with:

```python
self.charm._container = self.unit.get_container(self._name)
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -690,14 +690,17 @@ def __init__(
...
 
         # If Loki is run in single-tenant mode, all the chunks are put in a folder named "fake"
         # https://grafana.com/docs/loki/latest/operations/storage/filesystem/
         # https://grafana.com/docs/loki/latest/rules/#ruler-storage
         tenant_id = "fake"
         self._rules_dir = os.path.join(rules_dir, tenant_id)
+        # create tenant dir so that the /loki/api/v1/rules endpoint returns "no rule groups found"
+        # instead of "unable to read rule dir /loki/rules/fake: no such file or directory"
+        self.container.make_dir(self._rules_dir, make_parents=True)
```
Abuelodelanada, line 702: Same here, you can use:

```python
self.charm._container.make_dir(self._rules_dir, make_parents=True)
```

Abuelodelanada (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i: Both `self.charm._container` and `self.charm.unit.get_container("loki")` make inconvenient assumptions (charm must have a `_container` field; workload container name must be "loki").
I'll go with `self.charm._container` for now.

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i: Changed about to set `self.container` - loki lib code currently uses `self.container` in several places.

PR #41, opened 2021-11-24
Fix loki_server version property

Abuelodelanada:
We were hardcoding Loki version because a Loki bug. This bug was solved.

### DIFF
--- a/src/loki_server.py
+++ b/src/loki_server.py
@@ -65,16 +65,8 @@ class LokiServer:
             exception is raised
         """
         try:
-            # Un-comment the following lines once new Loki 2.3.1 version is released.
-            #
-            # We are hardcoding version here because there is and bug in last stable Loki version
-            # (2.3.0) that do not return its version: https://github.com/grafana/loki/issues/4133
-            # This bug it was already addressed in PR: https://github.com/grafana/loki/pull/4135
-            # but it's note released yet.
-            #  info = self._build_info()
-            #  return = info.get("version", None)
-            _ = self._build_info()
-            return "2.3.0"
+            info = self._build_info()
+            return info.get("version", None)
         except requests.exceptions.ConnectionError as e:
             raise LokiServerNotReadyError(str(e))
         except requests.exceptions.HTTPError as e:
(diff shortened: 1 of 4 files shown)


src/loki_server.py
```diff
@@ -65,16 +65,8 @@ def version(self) -> str:
...
-            # We are hardcoding version here because there is and bug in last stable Loki version
-            # (2.3.0) that do not return its version: https://github.com/grafana/loki/issues/4133
-            # This bug it was already addressed in PR: https://github.com/grafana/loki/pull/4135
-            # but it's note released yet.
-            #  info = self._build_info()
-            #  return = info.get("version", None)
-            _ = self._build_info()
-            return "2.3.0"
+            info = self._build_info()
+            return info.get("version", None)
```
sed-i, line 69: Given the `None` is preferred to a KeyError, maybe move the return below the try-except block?

sed-i (approved):

src/loki_server.py
Abuelodelanada: Yes, absolutely

src/loki_server.py
```diff
@@ -65,21 +65,17 @@ def version(self) -> str:
...
-            # This bug it was already addressed in PR: https://github.com/grafana/loki/pull/4135
-            # but it's note released yet.
-            #  info = self._build_info()
-            #  return = info.get("version", None)
-            _ = self._build_info()
-            return "2.3.0"
+            info = self._build_info()
+            version = info.get("version", None)
+            if not version:
+                raise LokiServerNotReadyError("Loki version could not be retrieved.")
```
balbirthomas, line 71: Only reason I can think that the version will not be available is that the Loki server is not yet responding to queries. If this is true then I would expect that the Loki charm sets it status to waiting status or defer and event handler depending on the use case for which version is being retrieved. I am not too familiar with the whole charm so I am leaving this to your better judgement.

balbirthomas (approved):

src/loki_server.py
Abuelodelanada: Yes, that's why we raise LokiServerNotReadyError

PR #45, opened 2021-12-02
Emit custom event if there is an error with alert rules.

Abuelodelanada:
Before this PR we were setting (in the library) `BlockedStatus` in the unit if there was an error with alert rules.

Now whether alert rules files does not contain the keys `alert` or `expr` or
there is no alert rules file in `alert_rules_path` a `loki_push_api_alert_rules_error` event
is emitted.

For this, in this PR we:
- created a custom event: `LokiPushApiAlertRulesError`
- refactor some methods

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -979,6 +979,12 @@ class LokiPushApiConsumer(RelationManagerBase):
         Args:
             event: a `CharmEvent` in response to which the consumer
                 charm must update its relation data.
+
+        Emits:
+            loki_push_api_endpoint_joined: Once the relation is established, this event is emitted.
+            loki_push_api_alert_rules_error: This event is emitted whether alert rules files does
+                not contain the keys `alert` or `expr` or there is no alert rules file in
+                `alert_rules_path`.
         """
         if not self._charm.unit.is_leader():
             return
@@ -988,17 +994,22 @@ class LokiPushApiConsumer(RelationManagerBase):
             # Remove this if when this issue is closed:
             # https://github.com/canonical/loki-operator/issues/3
             return
-        data = event.relation.data[event.unit].get("data")
 
-        if data:
-            self._stored.loki_push_api = json.loads(data)["loki_push_api"]
+        self._store_loki_push_api(event)
+
+        alert_groups, invalid_files = load_alert_rules_from_dir(
+            self._alert_rules_path,
+            self.topology,
+            recursive=False,
+            allow_free_standing=self.allow_free_standing_rules,
+        )
+        alert_rules_error_message = self._check_alert_rules(alert_groups, invalid_files)
+
+        if alert_rules_error_message:
+            self.on.loki_push_api_alert_rules_error.emit(alert_rules_error_message)
 
         event.relation.data[self._charm.app]["metadata"] = json.dumps(self.topology.as_dict())
-        try:
-            self._set_alert_rules(event)
-        except AlertRulesError as e:
-            self.on.loki_push_api_alert_rules_error.emit(message=str(e))
-
+        event.relation.data[self._charm.app]["alert_rules"] = json.dumps({"groups": alert_groups})
         self.on.loki_push_api_endpoint_joined.emit()
 
     def _on_logging_relation_departed(self, _):
@@ -1010,20 +1021,38 @@ class LokiPushApiConsumer(RelationManagerBase):
         """
         self.on.loki_push_api_endpoint_departed.emit()
 
-    def _set_alert_rules(self, event):
-        """Set alert rules into relation data.
+    def _store_loki_push_api(self, event) -> None:
+        """Stores loki_push_api in StoredState."""
+        data = event.relation.data[event.unit].get("data")
+
+        if data:
+            self._stored.loki_push_api = json.loads(data)["loki_push_api"]
+
+    def _check_alert_rules(self, alert_groups, invalid_files) -> str:
+        """Check alert rules.
 
         Args:
-            event: a `CharmEvent` in response to which the consumer
-                charm must update its relation data.
-        """
-        alert_groups = self._labeled_alert_groups
+            alert_groups: a list of prometheus alert rule groups.
+            invalid_files: a list of invalid rules files.
 
-        if alert_groups:
-            event.relation.data[self._charm.app]["alert_rules"] = json.dumps(
-                {"groups": alert_groups}
-            )
-        # TODO: else json.dumps({}) ?
+        Returns:
+            A string with the validation message. The message is not empty whether there are
+            invalid alert rules files or there are no alert rules groups.
+        """
+        message = ""
+
+        if invalid_files:
+            must_contain = ["'alert'", "'expr'"]
+            if self.allow_free_standing_rules:
+                must_contain.append("'%%juju_topology%%'")
+
+            message = "Failed to read alert rules (must contain {}): ".format(
+                ", ".join(must_contain)
+            ) + ", ".join(map(str, invalid_files))
+        elif not alert_groups:
+            message = "No alert rules found in {}".format(self._alert_rules_path)
+
+        return message
 
     def _label_alert_topology(self, rule) -> dict:
         """Insert juju topology labels into an alert rule.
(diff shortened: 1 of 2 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1005,12 +1063,11 @@ def _labeled_alert_groups(self) -> list:
             message = "Failed to read alert rules (must contain {}): ".format(
                 ", ".join(must_contain)
             ) + ", ".join(map(str, invalid_files))
-            self._charm.model.unit.status = BlockedStatus(message)
+            raise AlertRulesError(message)
```
sed-i, line 1066: Note that this is a potentially significant change of contract:
Before, the function would have returned the valid alert groups even if some are invalid; now nothing would be returned.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -246,6 +247,30 @@ def __init__(self, *args):
...
+To handle these situations the event must be observed in the `LokiClientCharm` charm.py file:
+
+
+```python
+class LokiClientCharm(CharmBase):
+
+    def __init__(self, *args):
+        super().__init__(*args)
+        ...
+        self._loki_consumer = LokiPushApiConsumer(self)
```
sed-i, line 263: I wonder if, in line with google docstring style, LokiPushApiConsumer should have an `Emits:` section listing all the custom events.

sed-i (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
simskij: cc @mmanciop

mmanciop: We need to keep the alert groups but append to them the juju topology to avoid clashes with multiple deployments of the same charm

lib/charms/loki_k8s/v0/loki_push_api.py
simskij: Sounds reasonable to me. 👍🏽

simskij (changes requested):

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: Good catch @sed-i !

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: @simskij @sed-i I refactored some methods, keeping the original contract. Now custom events are emitted from the same place (`_on_logging_relation_changed`) so it's easy to follow.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -655,6 +697,7 @@ class LokiPushApiEndpointJoined(EventBase):
 class LoggingEvents(ObjectEvents):
     """Event descriptor for events raised by `LokiPushApiProvider`."""
 
+    loki_push_api_alert_rules_error = EventSource(LokiPushApiAlertRulesError)
```
sed-i, line 700: `loki_push_api_alert_rules_error` is well scoped, and is an instance of the well named `LokiPushApiAlertRulesError`.
I wonder if we should drop the prefix here
```suggestion
    alert_rules_error = EventSource(LokiPushApiAlertRulesError)
```
so the user could just
```python
        self.framework.observe(
            self._loki_consumer.on.alert_rules_error,
            self._alert_rules_error
        )
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -890,6 +934,17 @@ def __init__(
             RelationRoleMismatchError: If the relation with the same name as provided
                 via `relation_name` argument does not have the `RelationRole.provides`
                 role.
+
+        Emits:
+            loki_push_api_endpoint_joined: This event is emitted when the relation between the
+                charmed operator that implements `LokiPushApiProvider` (Loki charm for instance)
+                and the charmed operator that implements `LokiPushApiConsumer` is established.
```
sed-i, line 941: "implements" -> "instantiates" ?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -890,6 +934,17 @@ def __init__(
...
+            loki_push_api_endpoint_departed: This event is emitted when the relation between the
+                charmed operator that implements `LokiPushApiProvider` (Loki charm for instance)
+                and the charmed operator that implements `LokiPushApiConsumer` is removed.
+            loki_push_api_alert_rules_error: This event is emitted whether alert rules files does
+                not contain the keys `alert` or `expr` or there is no alert rules file in
+                `alert_rules_path`.
```
sed-i, line 947:
```suggestion
            loki_push_api_alert_rules_error: This event is emitted when an invalid alert rules
                file is encountered or if `alert_rules_path` is empty.
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -917,6 +972,12 @@ def _on_logging_relation_changed(self, event):
         Args:
             event: a `CharmEvent` in response to which the consumer
                 charm must update its relation data.
+
+        Emits:
+            loki_push_api_endpoint_joined: Once the relation is established, this event is emitted.
+            loki_push_api_alert_rules_error: This event is emitted whether alert rules files does
+                not contain the keys `alert` or `expr` or there is no alert rules file in
+                `alert_rules_path`.
```
sed-i, line 980:
```suggestion
            loki_push_api_alert_rules_error: This event is emitted when an invalid alert rules
                file is encountered or if `alert_rules_path` is empty.
```

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i: Is it emitted every time there's a lifecycle event or just once?

sed-i (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: If we remove the prefix here, we should also remove:

```python
class LoggingEvents(ObjectEvents):
    """Event descriptor for events raised by `LokiPushApiProvider`."""

    loki_push_api_alert_rules_error = EventSource(LokiPushApiAlertRulesError)
    loki_push_api_endpoint_departed = EventSource(LokiPushApiEndpointDeparted)
    loki_push_api_endpoint_joined = EventSource(LokiPushApiEndpointJoined)
``` 

Will create an issue to address it later.

rbarry82: :+1: to this, if only because where it's coming from is already pretty explicit from the library name/instantiated classes.

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: It's only on `relation_changed` that this is checked, unless I missed something.

We don't necessarily have to keep exactly the same pattern, but for comparison, the Grafana Dashboard events I mentioned yesterday have two properties. One is a `bool` indicating whether or not it's valid, and the other is the actual message.

[Usage is here](https://github.com/canonical/grafana-operator/blob/main/lib/charms/grafana_k8s/v0/grafana_dashboard.py#L447), with the actual event definition [here](https://github.com/canonical/grafana-operator/blob/e3b5795cb3ff0da328549b3e266bfb8c1ebb4047/lib/charms/grafana_k8s/v0/grafana_dashboard.py#L249).

It potentially gives a little more flexibility in the future. Not even remotely blocking on this, just a note/example of adding more properties to events.

Abuelodelanada: @rbarry82 You are right, we are using only on `relation_changed`.

The Grafana Dashboard event you mentioned it's a very good example. I'll keep in mind, thanks!!

rbarry82 (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -565,7 +589,9 @@ def _group_name(path) -> str:
 
     invalid_files = []
     for path in filter(Path.is_file, Path(dir_path).glob("**/*.rule" if recursive else "*.rule")):
-        if rule := load_alert_rule_from_file(path, topology, allow_free_standing):
+        rule = load_alert_rule_from_file(path, topology, allow_free_standing)
```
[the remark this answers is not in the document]
Abuelodelanada: Thanks Ryan,

I did not modify this to keep this PR as small as possible. 
Following in https://github.com/canonical/loki-operator/issues/47

sed-i (approved):

simskij (approved):

PR #48, opened 2021-12-06
Fix various issues with the Loki Push API library

mmanciop:
1) Send the endpoint URL over application databags rather than unit (which will not work with distributed Loki)
2) Correctly handle multiple instances of the relation, avoiding concurrent overriding of the endpoint data in the Consumer storage
3) Support the upgrade of charms that include the LokiPushApiConsumer
4) Make the relation databag for the Loki endpoint extensible (using a dict rather than just a string to communicate the URL)

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -976,8 +976,7 @@ class LokiPushApiConsumer(RelationManagerBase):
         self._charm = charm
         self._relation_name = relation_name
         self._alert_rules_path = alert_rules_path
-        limit = self._charm.meta.relations[self._relation_name].limit
-        self._is_multi = True if type(limit) == int and limit > 0 else False
+        self._is_multi = self._charm.meta.relations[self._relation_name].limit != 1
 
         events = self._charm.on[relation_name]
         self.framework.observe(self._charm.on.upgrade_charm, self._on_logging_relation_changed)
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1006,7 +1006,7 @@ class LokiPushApiConsumer(RelationManagerBase):
                 self._process_logging_relation_changed(relation)
 
     def _process_logging_relation_changed(self, relation: Relation):
-        loki_push_api = relation.data[relation.app].get("loki_push_api")
+        loki_push_api_data = relation.data[relation.app].get("loki_push_api")
 
         if loki_push_api_data:
             self._stored.loki_push_api[relation.id] = json.loads(loki_push_api_data)
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1037,14 +1037,31 @@ class LokiPushApiConsumer(RelationManagerBase):
         self._stored.loki_push_api.pop(event.relation.id)
         self.on.loki_push_api_endpoint_departed.emit()
 
-    def _set_alert_rules(self, relation: Relation):
-        """Set alert rules into relation data.
+    def _check_alert_rules(self, alert_groups, invalid_files) -> str:
+        """Check alert rules.
 
         Args:
-            relation: the `Relation` instance for which alert rules must be processed.
+            alert_groups: a list of prometheus alert rule groups.
+            invalid_files: a list of invalid rules files.
+
+        Returns:
+            A string with the validation message. The message is not empty whether there are
+            invalid alert rules files or there are no alert rules groups.
         """
-        if alert_groups := self._labeled_alert_groups:
-            relation.data[self._charm.app]["alert_rules"] = json.dumps({"groups": alert_groups})
+        message = ""
+
+        if invalid_files:
+            must_contain = ["'alert'", "'expr'"]
+            if self.allow_free_standing_rules:
+                must_contain.append("'%%juju_topology%%'")
+
+            message = "Failed to read alert rules (must contain {}): ".format(
+                ", ".join(must_contain)
+            ) + ", ".join(map(str, invalid_files))
+        elif not alert_groups:
+            message = "No alert rules found in {}".format(self._alert_rules_path)
+
+        return message
 
     def _label_alert_topology(self, rule) -> dict:
         """Insert juju topology labels into an alert rule.
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -289,7 +289,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 8
+LIBPATCH = 9
 
 logger = logging.getLogger(__name__)
 
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1063,23 +1063,13 @@ class LokiPushApiConsumer(RelationManagerBase):
         return rule
 
     @property
-    def loki_push_api(self):
-        """Fetch Loki Push API endpoint sent from LokiPushApiProvider through relation data.
+    def loki_push_api(self) -> List[str]:
+        """Fetch Loki Push API endpoints sent from LokiPushApiProvider through relation data.
 
         Returns:
-            The Loki Push API endpoint (or None) if the relation this `LokiPushApiConsumer` tracks
-            has limit 1; with any other relation limit, it returns a list of Loki Push API
-            endpoints, which may be empty in case there are no relation instances.
+            A list with Loki Push API endpoints.
         """
-        loki_endpoints = [  # Convert from the StoredSet data structure to a plain list``
+        return [
             _type_convert_stored(loki_endpoint)
             for loki_endpoint in self._stored.loki_push_api.values()
         ]
-
-        if self._is_multi:
-            return loki_endpoints
-        else:
-            if len(loki_endpoints) == 1:
-                return loki_endpoints[0]
-            else:
-                return None
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -271,7 +271,7 @@ from typing import List, Optional, Tuple
 import yaml
 from ops.charm import CharmBase, RelationEvent, RelationRole
 from ops.framework import EventBase, EventSource, Object, ObjectEvents, StoredState
-from ops.model import BlockedStatus, Relation
+from ops.model import Relation
 
 # The unique Charmhub library identifier, never change it
 LIBID = "bf76f23cdd03464b877c52bd1d2f563e"
@@ -730,10 +730,28 @@ class LokiPushApiProvider(RelationManagerBase):
         self.container.make_dir(self._rules_dir, make_parents=True)
 
         events = self._charm.on[relation_name]
+        self.framework.observe(self._charm.on.upgrade_charm, self._on_logging_relation_changed)
         self.framework.observe(events.relation_changed, self._on_logging_relation_changed)
         self.framework.observe(events.relation_departed, self._on_logging_relation_departed)
 
     def _on_logging_relation_changed(self, event):
+        """Handle changes in related consumers.
+
+        Anytime there are changes in the relation between Loki
+        and its consumers charms.
+
+        Args:
+            event: a `CharmEvent` in response to which the consumer
+                charm must update its relation data.
+        """
+        if isinstance(event, RelationEvent):
+            self._process_logging_relation_changed(event.relation)
+        else:
+            # Upgrade event or other charm-level event
+            for relation in self._charm.model.relations[self._relation_name]:
+                self._process_logging_relation_changed(relation)
+
+    def _process_logging_relation_changed(self, relation: Relation):
         """Handle changes in related consumers.
 
         Anytime there are changes in relations between Loki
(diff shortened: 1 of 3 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1065,6 +1062,18 @@ def loki_push_api(self):
...
+            loki_endpoint for loki_endpoint in self._stored.loki_push_api.values()
+        ]
+
+        if self._is_multi:
+            return loki_endpoints
+        else:
+            if len(loki_endpoints) == 1:
+                return loki_endpoints[0]
+            else:
+                return None
```
Abuelodelanada, line 1079: This property may return:
- A list
- A string
- None

Maybe a better approach will be a property that always returns a **list**:

- A list with many elements
- A list with one element
- An empty list

Abuelodelanada (dismissed):

mmanciop:
> `limit: 0`, which is the default, means that you can have unlimited instances of the relation. If I understand this code right, this commit breaks the logic for the default usecase.

Also why would `limit: 1` be a multi relation?

Abuelodelanada:
> > `limit: 0`, which is the default, means that you can have unlimited instances of the relation. If I understand this code right, this commit breaks the logic for the default usecase.
> 
> Also why would `limit: 1` be a multi relation?

May be I'm missing something, but the `is_multi` is only necessary if `loki_push_api` returns different things. I we always return a `list`, the `is_multi` has no sense.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -77,41 +77,22 @@ def _provide_loki(self):
                 self.unit.status = BlockedStatus(str(e))
 
 
-The `LokiPushApiProvider` object has two main responsibilities:
+The `LokiPushApiProvider` object has several main responsibilities:
```
sed-i, line 80:
```suggestion
The `LokiPushApiProvider` object has several responsibilities:
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -748,28 +750,45 @@ def __init__(
...
         Args:
-            event: a `CharmEvent` in response to which the Loki
-                charm must update its relation data.
+            relation: the `Relation` instance to update.
         """
-        if event.relation.data[self.charm.unit].get("data") is None:
-            event.relation.data[self.charm.unit].update({"data": self._loki_push_api})
+        if self._charm.unit.is_leader():
+            relation.data[self._charm.app].update({"loki_push_api": self._loki_push_api})
             logger.debug("Saving Loki url in relation data %s", self._loki_push_api)
```
sed-i, line 789:
```suggestion
            logger.debug("Saved Loki url in relation data %s", self._loki_push_api)
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -748,28 +750,45 @@ def __init__(
...
 
-        if event.relation.data.get(event.relation.app).get("alert_rules") is not None:
+        if relation.data.get(relation.app).get("alert_rules"):
```
sed-i, line 791: Do we need to tell loki to reload?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -977,48 +998,45 @@ def _on_logging_relation_changed(self, event):
...
-        alert_rules_error_message = self._check_alert_rules(alert_groups, invalid_files)
+        if isinstance(event, RelationEvent):
+            self._process_logging_relation_changed(event.relation)
+        else:
+            # Upgrade event or other charm-level event
+            for relation in self._charm.model.relations[self._relation_name]:
+                self._process_logging_relation_changed(relation)
+
+    def _process_logging_relation_changed(self, relation: Relation):
+        if loki_push_api_data := relation.data[relation.app].get("loki_push_api"):
```
sed-i, line 1009: py3.5 compat?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -977,48 +998,45 @@ def _on_logging_relation_changed(self, event):
...
+            self._stored.loki_push_api[relation.id] = json.loads(loki_push_api_data)
```
sed-i, line 1010: Not sure why duplicate it into `_stored` if it is readily available in relation data.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -977,48 +998,45 @@ def _on_logging_relation_changed(self, event):
...
+                allow_free_standing=self.allow_free_standing_rules,
+            )
+            alert_rules_error_message = self._check_alert_rules(alert_groups, invalid_files)
 
-        if alert_rules_error_message:
-            self.on.loki_push_api_alert_rules_error.emit(alert_rules_error_message)
+            if alert_rules_error_message:
+                self.on.loki_push_api_alert_rules_error.emit(alert_rules_error_message)
+
+            relation.data[self._charm.app]["metadata"] = json.dumps(self.topology.as_dict())
```
sed-i, line 1025: duplication of line 1013

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -977,48 +998,45 @@ def _on_logging_relation_changed(self, event):
...
-    def _on_logging_relation_departed(self, _):
-        """Handle departures in related consumers.
+    def _on_logging_relation_departed(self, event):
+        """Handle departures in related providers.
 
         Anytime there are departures in relations between the consumer charm and Loki
         the consumer charm is informed, through a `LokiPushApiEndpointDeparted` event.
         The consumer charm can then choose to update its configuration.
         """
+        self._stored.loki_push_api[event.relation.id] = None
```
sed-i, line 1037: Is there a "pop" that could be used instead?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1065,6 +1083,19 @@ def loki_push_api(self):
         """Fetch Loki Push API endpoint sent from LokiPushApiProvider through relation data.
 
         Returns:
-            Loki Push API endpoint
+            The Loki Push API endpoint (or None) if the relation this `LokiPushApiConsumer` tracks
+            has limit 1; with any other relation limit, it returns a list of Loki Push API
+            endpoints, which may be empty in case there are no relation instances.
```
sed-i, line 1088: Add return type hint?
Seems like it's `Union[None, str, List[str]]`.
Would it make sense to have a return type of `Optional[List[str]]` or maybe even `List[str]`?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -77,41 +77,22 @@ def _provide_loki(self):
...
 
 
-The `LokiPushApiProvider` object has two main responsibilities:
+The `LokiPushApiProvider` object has several main responsibilities:
 
-1.- Be in charge of setting the Loki Push API Address into
-relation data so clients can use it to send logs. Every time a unit joins
-a relation: `$ juju add-relation loki-k8s loki-client-k8s` the object sets:
+1. Set the URL of the Loki Push API in the provider app data bag; the URL
+   must be unique to all instances (e.g., using a load balancer).
```
sed-i, line 83: This has a data loss potential during the time that no leader is present, e.g. during scale down.
BTW, I did not see any leader events registered.
I do recall this is intended as a single unit, so might worth noting that assumption.
Also, not sure what it means that no peer relation is defined in metadata.

sed-i: I don't understand this:

> the URL must be unique to all instances (e.g., using a load balancer).

mmanciop:
> > > `limit: 0`, which is the default, means that you can have unlimited instances of the relation. If I understand this code right, this commit breaks the logic for the default usecase.
> > 
> > 
> > Also why would `limit: 1` be a multi relation?
> 
> May be I'm missing something, but the `is_multi` is only necessary if `loki_push_api` returns different things. I we always return a `list`, the `is_multi` has no sense.

That is correct.

sed-i (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
mmanciop: > This has a data loss potential during the time that no leader is present, e.g. during scale down.
BTW, I did not see any leader events registered.
I do recall this is intended as a single unit, so might worth noting that assumption.
Also, not sure what it means that no peer relation is defined in metadata.

The API will need to work transparently with Loki charmed in distributed mode, so better to bite that bullet now.

mmanciop: > Also, not sure what it means that no peer relation is defined in metadata.

It means nothing special. We could use a peer relation to observe wether the Loki charm is scaled to more than one unit, and put it in blocked state like we do it with Prometheus + Ingress thou.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -432,6 +421,19 @@ def _is_valid_rule(rule: dict, allow_free_standing: bool) -> bool:
...
+        return list(map(_type_convert_stored, obj))
+    elif isinstance(obj, StoredDict):
+        rdict = {}  # type: Dict[Any, Any]
+        for k in obj.keys():
+            rdict[k] = _type_convert_stored(obj[k])
+        return rdict
+    else:
+        return obj
+
+
```
simskij, line 436: These conversion operations seems like something we probably should bug the operator framework team about providing as part of the Stored* datatypes down the line. Wdyt, @jnsgruk?

jnsgruk: This was discussed at length -- and rejected. Definitely a candidate for `observability-libs`, though.

jnsgruk: (context: https://github.com/canonical/operator/pull/572)

jnsgruk: saying that, the PR I link about was about implementing magic on the Stored* types. Perhaps a more explicit conversion might be more palatable.

mmanciop: I feel `StoredDict` should have a `.to_dict` like we do for Pebble services and plans, etc.

simskij (approved):
Once @sed-i's feedback has been addressed, this hsould be good to go.

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: That's kind of a can of worms.

The initial impetus behind that OF PR came about early in the charming campaign, when Balbir quickly realized that YAML/JSON dumping didn't work on `Stored*` objects, because they don't behave like the types they're named after (in a rather large variety of ways, but in that way particularly).

The PR was principally about hiding this from users by implementing equality methods with transparent conversions, and letting them be converted to JSON/YAML without the use of an explicit YAML/JSON formatter. My feeling then (and now) is that it doesn't feel Pythonic to have `StoredDict.as_dict()`, or to need a formatter for YAML/JSON, since a `FooDict` and `FooList` should just behave like the types they are claiming to be.

In the end, we had a discussion about whether `StoredState` (and the objects underneath it) are/were even appropriate places to keep data. You can't put the cat back in the bag, but certainly "we" at Canonical can write charms in a way which reflects best practice. For resiliency, as your recent comments about `StoredState` being harmful allude to, we may be better off creating peer relation buckets as the default and keeping data there, which is battle-tested, has a well-known API (anything that can be converted to/from a string), etc.

I'd suggest that, instead of going the route of explicit conversions (which will actually be `_type_convert_stored` anyway), a peer relation bucket is added and used here instead. The code block in the middle of [this comment](https://github.com/canonical/operator/pull/572#issuecomment-931529580) trivially demonstrates why. The existing magic conversions in OF already turn a list under a dict into a `StoredList`. Every node under the parent becomes a `Stored*` type, and they all need to be converted back recursively if it needs to work with `json.dumps()` or `yaml.dump()`, so an explicit conversion of `StoredDict.to_dict()` is necessarily more or less identical to `_type_convert_stored`, and the same goes for `StoredList.to_list()`.

Let's add and use a peer relation bucket, which gives anyone reading the code for this when writing a new charm a sense of "best practice" without list of footnotes illustrating the footguns from `StoredState`

simskij: I agree with you @rbarry82, but let's do that in a separate PR/issue 😅

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: Following in this issue #49

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: Let's follow this in #50

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: As far as I understand there is no need to reload Loki: https://grafana.com/docs/loki/latest/rules/

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -748,29 +750,46 @@ def __init__(
...
-            event.relation.data[self.charm.unit].update({"data": self._loki_push_api})
-            logger.debug("Saving Loki url in relation data %s", self._loki_push_api)
+        if self._charm.unit.is_leader():
+            relation.data[self._charm.app].update({"loki_push_api": self._loki_push_api})
+            logger.debug("Saved Loki url in relation data %s", self._loki_push_api)
 
-        if event.relation.data.get(event.relation.app).get("alert_rules") is not None:
-            logger.debug("Saving alerts rules to disk")
+        if relation.data.get(relation.app).get("alert_rules"):
+            logger.debug("Saved alerts rules to disk")
```
sed-i, line 792: If this is in past tense then should probably be under `_generate`?

sed-i (approved):

Abuelodelanada (dismissed):

simskij (approved):

Abuelodelanada (approved):

sed-i (approved):

PR #56, opened 2021-12-08
Syslog

dstathis:


lib/charms/loki_k8s/v0/log_proxy.py
```diff
@@ -586,26 +599,48 @@ def _scrape_configs(self) -> dict:
...
-            "scrape_configs": [
-                {
-                    "job_name": "system",
-                    "static_configs": self._generate_static_configs(config),
-                }
-            ]
+        scrape_configs = []
+
+        # Files config
+        labels = deepcopy(common_labels)
```
sed-i, line 615: Not sure why deepcopy is needed
```suggestion
        labels = common_labels.copy()
```

lib/charms/loki_k8s/v0/log_proxy.py
```diff
@@ -586,26 +599,48 @@ def _scrape_configs(self) -> dict:
...
+                "job_name": "syslog",
+                "syslog": {
+                    "listen_address": "127.0.0.1:{}".format(self._syslog_port),
+                    "label_structured_data": True,
+                    "labels": syslog_labels,
+                },
+            }
+            scrape_configs.append(syslog_config)  # type: ignore
+
+        return {"scrape_configs": scrape_configs}
```
sed-i, line 643: Consider updating the type hint accordingly, `Dict[List[dict]]`

tests/unit/test_log_proxy_consumer.py
```diff
@@ -103,6 +99,41 @@ def test__initial_config_sections(self):
...
+            set(
+                map(
+                    lambda x: x["job_name"],
+                    self.harness.charm._log_proxy._initial_config["scrape_configs"],
+                )
+            ),
+            expected_jobs,
+        )
+
+    def test__initial_config_labels(self):
```
sed-i, line 114: Function name should have a behavioral phrasing, e.g.
- test all scrape jobs have all topology labels
- test all job names are either system or syslog

sed-i (dismissed):

lib/charms/loki_k8s/v0/log_proxy.py
```diff
@@ -586,26 +599,48 @@ def _scrape_configs(self) -> dict:
...
+        labels.update(
+            {
+                "job": job_name,
+                "__path__": "",
+            }
+        )
+        config = {"targets": ["localhost"], "labels": labels}
+        scrape_config = {
+            "job_name": "system",
+            "static_configs": self._generate_static_configs(config),
```
Abuelodelanada, line 625: If the LogProxyConsumer is instantiated without a list of files, the `static_configs` will be empty... the generated config file is going to work in this case?

Abuelodelanada (approved):

lib/charms/loki_k8s/v0/log_proxy.py
sed-i: Gotta put a test on it :muscle:

lib/charms/loki_k8s/v0/log_proxy.py
```diff
@@ -226,17 +233,21 @@ class LogProxyConsumer(RelationManagerBase):
     def __init__(
         self,
         charm,
-        log_files: list,
         container_name: Optional[str],
         relation_name: str = DEFAULT_RELATION_NAME,
+        log_files: list = None,
```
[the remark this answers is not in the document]
dstathis: why?

lib/charms/loki_k8s/v0/log_proxy.py
```diff
@@ -226,17 +233,21 @@ class LogProxyConsumer(RelationManagerBase):
...
         super().__init__(charm, relation_name)
         self._stored.set_default(grafana_agents="{}")
         self._charm = charm
         self._relation_name = relation_name
         self._container_name = container_name
         self._container = self._get_container(container_name)
-        self._log_files = log_files
+        self._log_files = log_files or []
+        self._syslog_port = syslog_port
+        self._is_syslog = syslog
```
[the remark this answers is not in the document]
dstathis: good idea and I think the arg should probably be renamed too

lib/charms/loki_k8s/v0/log_proxy.py
```diff
@@ -366,8 +376,11 @@ def _is_promtail_attached(self) -> bool:
         """
         try:
             resource_path = self._charm.model.resources.fetch("promtail-bin")
-        except ModelError:
-            return False
+        except NameError as e:
+            if "invalid resource name" in str(e):
+                return False
```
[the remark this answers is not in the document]
dstathis: I am not convinced this is the only way we could see a `NameError`

lib/charms/loki_k8s/v0/log_proxy.py
```diff
@@ -586,26 +599,48 @@ def _scrape_configs(self) -> dict:
...
-                "__path__": "",
-            },
+        job_name = "juju_{}_{}_{}".format(
+            self._charm.model.name, self._charm.model.uuid, self._charm.model.app.name
+        )
+        common_labels = {
+            "juju_application": self._charm.app.name,
+            "juju_model": self._charm.model.name,
+            "juju_model_uuid": self._charm.model.uuid,
+            "juju_charm": self._charm.meta.name,
```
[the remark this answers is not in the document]
dstathis: It's in the `JujuTopology` class in Prometheus so I just used it.

lib/charms/loki_k8s/v0/log_proxy.py
dstathis: You're right

lib/charms/loki_k8s/v0/log_proxy.py
```diff
@@ -586,26 +599,48 @@ def _scrape_configs(self) -> dict:
...
+        scrape_config = {
+            "job_name": "system",
+            "static_configs": self._generate_static_configs(config),
         }
+        scrape_configs.append(scrape_config)
+
+        # Syslog config
+        if self._is_syslog:
+            syslog_labels = deepcopy(common_labels)
+            syslog_labels.update({"job": "syslog_{}".format(job_name)})
```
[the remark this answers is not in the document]
dstathis: I thought about this and yes

lib/charms/loki_k8s/v0/log_proxy.py
dstathis: I don't understand where I would put this type hint

tests/unit/test_log_proxy_consumer.py
```diff
@@ -81,7 +77,7 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args)
         self._port = 3100
         self._log_proxy = LogProxyConsumer(
-            charm=self, log_files=LOG_FILES, container_name="consumercharm"
+            charm=self, log_files=LOG_FILES, container_name="consumercharm", syslog=True
```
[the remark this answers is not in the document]
dstathis: Perhaps. Seeing as all previous tests passed with this set to true, It's at least as good as it was. Thoughts @Abuelodelanada

tests/unit/test_log_proxy_consumer.py
dstathis: This is to be consistent with existing test names.

lib/charms/loki_k8s/v0/log_proxy.py
dstathis: This would be an integration test so we should write it after the lib merge. I'm pretty sure it worked in my testing.

lib/charms/loki_k8s/v0/log_proxy.py
sed-i: Function definition currently has just dict.

tests/unit/test_log_proxy_consumer.py
sed-i: Currently this test is quite implicit. Consider breaking it down into
- test all scrape jobs have all topology labels
- test all job names are either system or syslog
- more? not sure what this test tests

lib/charms/loki_k8s/v0/log_proxy.py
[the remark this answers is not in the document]
sed-i: Not sure what you mean. See [PEP 3102](https://peps-aaturner.readthedocs.io/pep-3102/#rationale).
Preventing users from passing these positionally is future proofing for API change.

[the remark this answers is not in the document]
rbarry82: It's worth noting, like other times this has been mentioned, that this PEP is relatively ancient and doesn't have high adoption. It's not a bad thing to do it, but I wouldn't even consider the lack of it a smell.

lib/charms/loki_k8s/v0/log_proxy.py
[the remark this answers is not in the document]
rbarry82:
```python
except NameError as e:
  if "invalid resource name" in str(e):
    return False
  else:
    raise e from None
 ```

File: lib/charms/loki_k8s/v0/log_proxy.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (rbarry82) [reply lib/charms/loki_k8s/v0/log_proxy.py:609]: `juju_unit` can also be used for cross-model relations and aggregations.

`JujuTopology` has a significant patch pending to resolve some labeling errors/bugs. See [here](https://github.com/canonical/prometheus-k8s-operator/blob/c285372fa51c6ef031239d39c4c66c5255ca6a48/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L602). You obviously don't need to implement this exactly, but "either `juju_charm` or `juju_unit` must be present, depending whether it's a K8s/OF charm or  CMR/reactive charm" is a reasonable safeguard to match Prom labeling.

File: lib/charms/loki_k8s/v0/log_proxy.py (hunk 2 above)
Comment (rbarry82) [reply lib/charms/loki_k8s/v0/log_proxy.py:643]: -1 from me here. If we're going to try to "properly" type it, we'd want to define a real type above, since `Dict[List[dict]]` isn't arbitrarily any clearer than just `dict` or `Dict[List]`. I'd go for  a simple `dict` for clarify rather than a middle ground which only partially expresses the datastructure, or define a "real" type which actually expresses the whole thing.

`Dict[List[Dict[str, Union[str, Dict[str, Union[str, List[str]]]]]]` gets unbearable really fast.

File: tests/unit/test_log_proxy_consumer.py (hunk 3 above)
Comment (rbarry82) [reply tests/unit/test_log_proxy_consumer.py:112]: `test_base_config_includes_system_logs` seems to cover it

File: lib/charms/loki_k8s/v0/log_proxy.py (hunk 4 above)
Comment (rbarry82) [reply lib/charms/loki_k8s/v0/log_proxy.py:625]: It will return an empty list, which Prometheus is ok parsing

File: lib/charms/loki_k8s/v0/log_proxy.py (hunk 2 above)
Comment (sed-i) [reply lib/charms/loki_k8s/v0/log_proxy.py:643]: `Dict[List[dict]]` is short and sweet to my eyes, and gives way more insight than `dict`, and very quickly so, into the type.
cf. [tradeoff](https://www.merriam-webster.com/dictionary/trade-off)

File: tests/unit/test_log_proxy_consumer.py (hunk 10 above)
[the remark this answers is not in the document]
Comment (Abuelodelanada) [reply tests/unit/test_log_proxy_consumer.py:80]: I would add some extra test with `syslog=False`

File: lib/charms/loki_k8s/v0/log_proxy.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (dstathis) [reply lib/charms/loki_k8s/v0/log_proxy.py:238]: I'm not going to do this. It's a relatively unknown feature that doesn't really give any benefit.

File: lib/charms/loki_k8s/v0/log_proxy.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (dstathis) [reply lib/charms/loki_k8s/v0/log_proxy.py:381]: @rbarry82 I'm not familiar with this syntax. How does `raise e from None` differ from just `raise`?

File: lib/charms/loki_k8s/v0/log_proxy.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (dstathis) [reply lib/charms/loki_k8s/v0/log_proxy.py:250]: I actually ended up changing just the arg name. Is there a reason you prefer "use" over "is"?

File: lib/charms/loki_k8s/v0/log_proxy.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (dstathis) [reply lib/charms/loki_k8s/v0/log_proxy.py:609]: The `JujuTopology` class is in the rewrite @Abuelodelanada is working on right now and should cover any concerns here.

File: lib/charms/loki_k8s/v0/log_proxy.py (hunk 2 above)
Comment (dstathis) [reply lib/charms/loki_k8s/v0/log_proxy.py:643]: This does not fix the need for `# type: ignore` so I see no point.

File: tests/unit/test_log_proxy_consumer.py (hunk 10 above)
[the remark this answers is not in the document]
Comment (dstathis) [reply tests/unit/test_log_proxy_consumer.py]: Added test__syslog_not_enabled

File: lib/charms/loki_k8s/v0/log_proxy.py [hunk 11]
```diff
@@ -44,7 +47,7 @@ def __init__(self, *args):
            ...
            try:
                self._log_proxy = LogProxyConsumer(
-                   charm=self, log_files=LOG_FILES, container_name=PEER
+                   charm=self, log_files=LOG_FILES, container_name=PEER, syslog=True
```
sed-i, line 50: Seems like this needs an update.

sed-i (approved):

lib/charms/loki_k8s/v0/log_proxy.py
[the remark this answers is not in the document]
dstathis: keeping it as is

Abuelodelanada (approved):

lib/charms/loki_k8s/v0/log_proxy.py
dstathis: I promise I'll open a PR with this after merging

PR #61, opened 2022-01-05
copy itest from prom and streamline confs from alertmanager

sed-i:


Abuelodelanada (approved):
Awesome Leon!

tests/integration/helpers.py
```diff
@@ -1,11 +1,53 @@
...
+    url = f"http://{address}:3100"
+    logger.info("Loki public address: %s", url)
+
+    response = urllib.request.urlopen(
+        f"{url}/loki/api/v1/status/buildinfo", data=None, timeout=2.0
+    )
+    return response.code == 200 and "version" in json.loads(response.read())
+
+
+class IPAddressWorkaround:
```
dstathis, line 29: Once again, I really don't like calling this `IPAddressWorkaround`. This is not at all related to ip addresses. It just shortens the update status interval.

sed-i: Depends how you look at it:
- for the purpose of this test I do not need to change the update status per se, what I need is an IP address
- whatever happens inside is an implementation detail; currently update status does the trick, tomorrow maybe something else. But ideally this bug would be fixed and the whole thing deleted.
In other words I prefer a semantic name over a technical one.

Abuelodelanada (approved):

Abuelodelanada (approved):

rbarry82 (approved):

PR #64, opened 2022-01-07
Simplify charm summary to play nice with charmhub

simskij:


Abuelodelanada (approved):

metadata.yaml
```diff
@@ -9,11 +9,8 @@ display-name: Loki
 description: |
   Loki for Kubernetes cluster
 summary: |
-  Unlike other logging systems, Loki is built around the idea of only indexing metadata
-  about your logs: labels (just like Prometheus labels).
-  Log data itself is then compressed and stored in chunks in object stores such as S3 or GCS,
-  or even locally on the filesystem. A small index and highly compressed chunks simplifies
-  the operation and significantly lowers the cost of Loki.
```
mmanciop, line 16: It does not mention querying. This makes it sound almost like an archiving solution.

how about:

“ Grafana Loki is a multi-tenant log aggregation system that follows the design philosophy of Prometheus, including the indexing of logs by labels and sporting a query language similar to PromQL.“

mmanciop (changes requested):

sed-i (approved):

metadata.yaml
Abuelodelanada: LGTM.

Another alternative from the [Grafana Labs page](https://grafana.com/docs/loki/latest/):

"Grafana Loki is a set of components that can be composed into a fully featured logging stack."

simskij: > It does not mention querying

It does say "consuming", though. I think @Abuelodelanada's suggestion is preferable in this case as it has to be less than 100 characters (even though that's not enforced by the cli yet).

image

metadata.yaml
```diff
@@ -9,11 +9,8 @@ display-name: Loki
...
+  Loki is a cost-efficient tool for compressing, storing and consuming 
+  logs indexed by log metadata.
```
simskij, line 13:
```suggestion
  Loki is a set of components that can be composed into a fully featured logging stack.
```

Issue #63, opened 2022-01-06
Merge log_proxy and loki_push_api libraries


Abuelodelanada:
Now, both libraries uses the same interface name: `loki_push_api` and they do different things:

-  loki_push_api shares Loki Push API endpoint over relation data
-  log_proxy shares Loki Push API endpoint over relation data AND Injects Promtail in workload charm

Abuelodelanada:
# Provider side

## LogProxyProvider 

Observes `_relation_changed` event and stores  `loki_push_api` endpoint and `PROMTAIL_BINARY_ZIP_URL` in data key in unit relation data bucket

## LokiPushApiProvider

Observes: 

- `upgrade_charm`, `relation_changed`: 
  - If the unit is the leader stores loki_push_api endpoint in app relation data bucket.
  - Get alert the rules stored in app relation data bucket by the consumer, generates and store the alert rules files.

- `relation_departed`:
  - Remove alert rules files from disk

# Consumer side

## LogProxyConsumer

Observes:

 - `_relation_created`: Generate intial config for Promtail
 - `_relation_changed`: If there is a `data` object in unit relation data bucket:
   - Obtain Promtail binary
   - Update config for Promtail (add agent to clients section)
   - Update agents_lists in `_stored`
   - Add Pebble layer for Promtail service
   - Restart Promtail service
- ` _relation_departed`:
   - Update config (remove the agent from clients section)
   - Update agents_list in `_stored`
   - If there is no agent in clients section stops Promtail service, else restarts Promtail service.



## LokiPushApiConsumer

Observes:
 - `upgrade_charm` and `relation_changed`:
   - Store loki_push_api in _stored
   - If unit is leader:
     - Load alert rules from a directory and stored them in app relation data bucket
   - Emit loki_push_api_endpoint_joined event

- `relation_departed`:
   - Remove `loki_push_api` from `_stored`
   - Emit `loki_push_api_endpoint_departed` event.

Abuelodelanada:
# Merge Proposal v1

## Provider side

Keep `LokiPushApiProvider` but `PROMTAIL_BINARY_ZIP_URL` should also be in app relation data bucket (we are discussing the new relation data schema.)

## Consumer side 

Keep `LokiPushApiConsumer` adding the `LogProxyConsumer` functionalities.
`LokiPushApiConsumer` should be instantiated with a flag that indicated whether Promtail should be downloaded and configured or not.

## Metadata

Both sides of the relation should be named:
[...]

PR #67, opened 2022-01-12
Merge libs

Abuelodelanada:
Related to #63

This PR merge `log_proxy` lib into `loki_push_api`.

- `LogProxyConsumer` was moved to `loki_push_api.py` and `LogProxyProvider`removed. 
- `LokiPushApiConsumer` now inherits from `ConsumerBase`.
- `ConsumerBase` handles alert rules stuff.
- Relation data now is shared over App relation data bags


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -17,72 +17,81 @@
...
+- `relation_name`: The name of the relation that the charm uses to interact
+  with its clients which implements `LokiPushApiConsumer` or `LogProxyConsumer`.
+  If provided, this relation name must match a provided
   relation in metadata.yaml with the `loki_push_api` interface.
-  This argument is not required if your metadata.yaml has precisely one
-  provided relation in metadata.yaml with the `loki_push_api` interface, as the
-  lib will automatically resolve the relation name inspecting the using the
-  meta information of the charm.
+  Typically `LokiPushApiConsumer` use "logging" as a relation_name and `LogProxyConsumer` use
+  "log_proxy".
```
sed-i, line 45: Not sure if this is the right place or if was already mentioned elsewhere: I'd add an explicit differentiation between push api and logproxy, for example:
push api is for talking directly to loki, whereas log proxy is for sending logs via promtail.
(BTW how does syslog fit in here?)

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -786,6 +935,7 @@ def _process_logging_relation_changed(self, relation: Relation):
         """
         if self._charm.unit.is_leader():
             relation.data[self._charm.app].update({"loki_push_api": self._loki_push_api})
+            relation.data[self._charm.app].update({"data": self._promtail_binary_url})
```
sed-i, line 938: Would it be easy at this point in time to rename `data`?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -907,7 +1062,67 @@ def alerts(self) -> dict:
...
+        Returns:
+            A string with the validation message. The message is not empty whether there are
+            invalid alert rules files or there are no alert rules groups.
+        """
+        message = ""
+
+        if invalid_files:
+            must_contain = ["'alert'", "'expr'"]
+            if self.allow_free_standing_rules:
+                must_contain.append("'%%juju_topology%%'")
```
sed-i, line 1114: Looks familiar but I'm confused: free standing means no juju topology stub, no? i.e. is there a `not` missing?
```suggestion
            if not self.allow_free_standing_rules:
                must_contain.append("'%%juju_topology%%'")
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1039,55 +1231,473 @@ def _on_logging_relation_departed(self, event):
...
+            try:
+                self._obtain_promtail(event)
+                self._update_config(event)
+                self._update_agents_list(event)
+                self._add_pebble_layer()
+                self._container.restart(WORKLOAD_SERVICE_NAME)
+            except HTTPError as e:
+                msg = "Promtail binary couldn't be download - {}".format(str(e))
+                logger.warning(msg)
+                raise PromtailDigestError(msg)
```
sed-i, line 1307: iiuc we can narrow the span of the try:

```suggestion
            try:
                self._obtain_promtail(event)
            except HTTPError as e:
                msg = "Promtail binary couldn't be download - {}".format(str(e))
                logger.warning(msg)
                raise PromtailDigestError(msg)
            else:
                self._update_config(event)
                self._update_agents_list(event)
                self._add_pebble_layer()
                self._container.restart(WORKLOAD_SERVICE_NAME)
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1039,55 +1231,473 @@ def _on_logging_relation_departed(self, event):
...
+        """
+        # TODO: use the JujuTopology object
+        job_name = "juju_{}_{}_{}".format(
+            self._charm.model.name, self._charm.model.uuid, self._charm.model.app.name
+        )
+        common_labels = {
+            "juju_application": self._charm.app.name,
+            "juju_model": self._charm.model.name,
+            "juju_model_uuid": self._charm.model.uuid,
+            "juju_charm": self._charm.meta.name,
```
sed-i, line 1634: Would it make sense to use `class JujuTopology` here?

sed-i (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: We are discussing the relation data schema in a issue. This part is going to change soon.

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: Wasn't it you who added this part? ;-)

In other parts of the code is `if not ...`

What do you think @dstathis ?

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: absolutely! Good catch!

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: @dstathis what do you think about this?

dstathis: Given that this section is for the provider, it should always be logging. You don't provide log_proxy. You only consume it.

dstathis: (syslog is for log_proxy only and is discussed later)

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -177,6 +186,101 @@ def __init__(self, *args):
...
+   ```python
+   from charms.loki_k8s.v0.log_proxy import LogProxyConsumer, PromtailDigestError
+
+   ...
+
+       def __init__(self, *args):
+           ...
+           try:
+               self._log_proxy = LogProxyConsumer(
+                   charm=self, log_files=LOG_FILES, container_name=PEER, syslog=True
```
dstathis, line 212: I think there may have been a bit of a merge issue here. This should say `enable_syslog=True`

mmanciop (changes requested):
I'll look at this again when the alert rules support is in :(

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1039,55 +1231,470 @@ def _on_logging_relation_departed(self, event):
...
+            _type_convert_stored(loki_endpoint)
+            for loki_endpoint in self._stored.loki_push_api.values()
+        ]
+
+
+class PromtailDigestError(Exception):
+    """Raised if there is an error with Promtail binary file."""
+
+
+class LogProxyConsumer(RelationManagerBase):
```
simskij, line 1251: This should be using the consumer base class, shouldn't it? The log proxy is currently lacking support for alerts altogether 😅

simskij (changes requested):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -177,6 +186,101 @@ def __init__(self, *args):
...
+       def __init__(self, *args):
+           ...
+           try:
+               self._log_proxy = LogProxyConsumer(
+                   charm=self, log_files=LOG_FILES, container_name=PEER, enable_syslog=True
+               )
+           except PromtailDigestError as e:
+               msg = str(e)
+               logger.error(msg)
+               self.unit.status = BlockedStatus(msg)
```
mmanciop, line 217: This is IMO not a good pattern: people will copy this, and unknowingly override their status with something else in the first hook that runs. @simskij @jnsgruk do we have better options?

mmanciop:
When I use the library like this:

```
            charm=self,
            enable_syslog=True,
        )
```

I get:

```
TypeError: __init__() missing 2 required positional arguments: 'log_files' and 'container_name'
```

I really do not believe these arguments should be required. The documentation says that `container_name` can be omitted if there is only one workload container. And having to specify an empty array of log files feels wrong.

lib/charms/loki_k8s/v0/loki_push_api.py
simskij: I think the pattern we've discussed using until we come up with something better is to have the lib emit some kind of error event, and have the charm author listen to that and decide themselves whether to block or not.

simskij: I think the pattern we've discussed using until we come up with something better is to have the lib emit some kind of error event, and have the charm author listen to that and decide themselves whether to block or not.

mmanciop (changes requested):
The LogProxyConsumer has a remarkably different API than LogPushApiConsumer, in that is does not emit events, and I really do not like this. I feel very strongly that the charm author experience should be virtually the same.

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: This is a bug in the merge I think. We remove these some weeks ago.

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: You are right, but this is a new feature. 
In this PR I only merge the 2 libs into one. I'll create an issue for this.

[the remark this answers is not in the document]
Abuelodelanada: We created `ConsumerBase` in this PR to let us add alert rules support in a next PR.

lib/charms/loki_k8s/v0/loki_push_api.py
dstathis: Intuitively `not` seems correct but I think I'm not the expert here.

lib/charms/loki_k8s/v0/loki_push_api.py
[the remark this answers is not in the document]
dstathis: It certainly will need to extend ConsumerBas. My understanding is that it will be tested and included in the next PR.

dstathis: It certainly will need to extend ConsumerBase. My understanding is that it will be tested and included in the next PR.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1039,55 +1231,470 @@ def _on_logging_relation_departed(self, event):
...
+    """LogProxyConsumer class."""
+
+    _stored = StoredState()
+
+    def __init__(
+        self,
+        charm,
+        log_files: list,
+        container_name: Optional[str],
+        relation_name: str = "log_proxy",
```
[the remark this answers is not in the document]
dstathis: While not important technically, I would agree that would be nicer.

dstathis (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -17,72 +17,81 @@
 send log to Loki by implementing the consumer side of the `loki_push_api` relation interface.
 For instance a Promtail or Grafana agent charm that needs to send logs to Loki.
 
+- `LogProxyConsumer`: This object can be used by any K8s charmed operator that needs to
```
rbarry82, line 20: 'K8s charmed operator' -> 'Charmed Operator'

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -17,72 +17,81 @@
...
+send logs to Loki through a Log Proxy by implementing the consumer side of the `loki_push_api`
+relation interface.
+Filtering logs in Loki is largely performed on the basis of labels.
+In the Juju ecosystem, Juju topology labels are used to uniquely identify the workload that
```
rbarry82, line 24: 'telemetry, such as logs'

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: Let's follow this in issue #69

Abuelodelanada:
> The LogProxyConsumer has a remarkably different API than LogPushApiConsumer, in that is does not emit events, and I really do not like this. I feel very strongly that the charm author experience should be virtually the same.

LogProxyConsumer works this way since its first implementation in November. Let's follow this change in #70

Abuelodelanada:
> When I use the library like this:
> 
> ```
>         self._log_proxy = LogProxyConsumer(
>             charm=self,
>             enable_syslog=True,
>         )
> ```
> 
> I get:
> 
> ```
> TypeError: __init__() missing 2 required positional arguments: 'log_files' and 'container_name'
> ```
> 
> I really do not believe these arguments should be required. The documentation says that `container_name` can be omitted if there is only one workload container. And having to specify an empty array of log files feels wrong.

Fixed

rbarry82:
Easier for me to just commit a bunch of doc updates than comment on them one by one.

In general, though, the `Provider` and `Consumer` class names all appear to be backwards from our normal conventions unless I missed something.

I'm also a little concerned about the usage of `event.app` as a general pattern -- since it's only the URI and some basic data, it's _probably_ ok, but `event.unit` may be better.

Similarly, the dict keys could be better:

```python
(interface)"loki_push_api" -> relation.data[app]["loki_push_api"]["url"]
```

May be too many "loki_push_api", especially when that's also the name of a `StoredState` variable. Using `loki_push_uri` or similar may be less confusing for readers, but this is just a nit.

It is **very** concerning that the library, by default, tries to fetch Promtail from the internet, and that the documented suggestion "rebuild the charm". For end-users in airgapped environments, this is serious thing, and rebuilding the charm may not be practical (to say nothing of any hashing guarantees we may eventually give on "signed" charms). It would be _far_ preferable to default to using a resource.

mmanciop:
> May be too many "loki_push_api", especially when that's also the name of a `StoredState` variable. Using `loki_push_uri` or similar may be less confusing for readers, but this is just a nit.

It actually _started out_ as the URL being the value of the `loki_push_api`, but I changed it a while back because, when we start working on TLS, we will need a place to put CA and server certs. Having the `loki_push_api` key have a `dict` as a value is a basic form of guarantee of being able to evolve the relation contract in the future.

mmanciop:
> It is very concerning that the library, by default, tries to fetch Promtail from the internet, and that the documented suggestion "rebuild the charm". For end-users in airgapped environments, this is serious thing, and rebuilding the charm may not be practical (to say nothing of any hashing guarantees we may eventually give on "signed" charms). It would be far preferable to default to using a resource.

To my understanding, the library will try to fetch from the internet UNLESS you passed the `promtail` binary as resource to the charm when building it and releasing it. We have had a lot of discussions around this with @jnsgruk and @simskij back in the day, and it is the least of all evils. And yes, it makes for a poor UX with 1st party charms in air-gapped environments. I really wish we could specify resources as dependencies of charm libraries, and have those resources automatically added to the charm at build tine.

mmanciop:
> In general, though, the `Provider` and `Consumer` class names all appear to be backwards from our normal conventions unless I missed something.

No, I think they are fine:

```
- `LogProxyConsumer`: This object can be used by any Charmed Operator which needs to send telemetry, such as logs, to Loki through a Log Proxy by implementing the consumer side of the `loki_push_api` relation interface.
```

The direction of what is consumer and what is producer follows the roles of the relation: `provides` and `requires`. The Provider, because of how `juju expose` works, is the one with the server-socket: in this case, the Loki Push API server. I understand that it is counter-intuitive from the perspective of a log (the app produces the logs, after all), and I am open to discussing whether we should use different names for the Python API objects.

jnsgruk:
> > It is very concerning that the library, by default, tries to fetch Promtail from the internet, and that the documented suggestion "rebuild the charm". For end-users in airgapped environments, this is serious thing, and rebuilding the charm may not be practical (to say nothing of any hashing guarantees we may eventually give on "signed" charms). It would be far preferable to default to using a resource.
> 
> To my understanding, the library will try to fetch from the internet UNLESS you passed the `promtail` binary as resource to the charm when building it and releasing it. We have had a lot of discussions around this with @jnsgruk and @simskij back in the day, and it is the least of all evils. And yes, it makes for a poor UX with 1st party charms in air-gapped environments. I really wish we could specify resources as dependencies of charm libraries, and have those resources automatically added to the charm at build tine.

Indeed - this is a bit of a mess that will eventually be fixed through a combination of subordinates on Kubernetes, and by being able to specify dependencies or resources for charm libraries.

@mmanciop is right though -- there should be no attempts to reach out to Github *if* a resource is attached, and we should make sure the documentation on how to achieve that is first class.

Abuelodelanada:
> 
> Indeed - this is a bit of a mess that will eventually be fixed through a combination of subordinates on Kubernetes, and by being able to specify dependencies or resources for charm libraries.
> 
> @mmanciop is right though -- there should be no attempts to reach out to Github _if_ a resource is attached, and we should make sure the documentation on how to achieve that is first class.

It works that way.

First we check if Promtail is attached. If it's attached we use that file.
If it's not attached, we check whether Promtail must be downloaded or not. We download Promtail if binary it's not in the container or the sha256 mismatch.

```python
    def _obtain_promtail(self, event) -> None:
        """Obtain promtail binary from an attached resource or download it."""
        if self._is_promtail_attached():
            return

        if self._promtail_must_be_downloaded():
            self._download_and_push_promtail_to_workload(event)
        else:
            self._push_binary_to_workload()
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -708,7 +850,12 @@ class LokiPushApiProvider(RelationManagerBase):
     """A LokiPushApiProvider class."""
 
     def __init__(
-        self, charm, relation_name: str = DEFAULT_RELATION_NAME, *, rules_dir="/loki/rules"
+        self,
+        charm,
+        relation_name: str = DEFAULT_RELATION_NAME,
+        port: int = 3100,
+        *,
```
sed-i, line 857: We for sure know that self, charm and relation_name are here to stay, but not so sure about port.
```suggestion
        *,
        port: int = 3100,
```
Although that may break some existing code, if the port is passed positionally.

sed-i (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
mmanciop: In units with multiple workload containers we need to be able to set different ports or the promtails in the various containers will collide and only one will be able to bind the port.

mmanciop: (Nvd from mobile I thought the comment was about the LogProxyConsumer)

sed-i (approved):

dstathis (approved):

PR #71, opened 2022-01-14
Use events for LogProxy, get rid of PromtailError as Exception

### LINKED ISSUE #69: The lib should emit some kind of error event rather than raising an Expection
I think the pattern we've discussed using until we come up with something better is to have the lib emit some kind of error event, and have the charm author listen to that and decide themselves whether to block or not.

_Originally posted by @simskij in https://github.com/canonical/loki-k8s-operator/pull/67#discussion_r784051480_

### LINKED ISSUE #70: LogProxyconsumer and LokiPushApiConsumer should implement the same interface
The LogProxyConsumer has a remarkably different API than LogPushApiConsumer, in that is does not emit events, and I really do not like this. I feel very strongly that the charm author experience should be virtually the same.

_Originally posted by @mmanciop in https://github.com/canonical/loki-k8s-operator/pull/67#pullrequestreview-851854320_

rbarry82:
Send events like LokiPushApi already does so the UX is the same.
    
Instead of raising exceptions for everything in the LogProxy code, forcing consumers to catch them to set state, fire back an event if something happens with the Promtail Binary.
    
Only send a new `ContainerNotFound` Exception if the container name is unset or ambiguous enough that we can't suss one out.
    
Add a bunch of type hints.
    
Update tests.
    
Closes #70 
Closes #69 

Really depends on #67, but the changes on top of that are reasonably small

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1482,7 +1482,7 @@ class LogProxyConsumer(ConsumerBase):
     def _get_container_name(self, container_name: Optional[str] = "") -> str:
         """Gets a container_name.
 
-        If there is more than one container in the Pod a `PromtailDigestError` is emitted.
+        If there is more than one container in the Pod a `ContainerNotFoundError` is raised.
 
         Args:
             container_name: The container name.
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -204,21 +204,49 @@ Adopting this object in a Charmed Operator consist of two steps:
    For example:
 
    ```python
-   from charms.loki_k8s.v0.log_proxy import LogProxyConsumer, PromtailDigestError
+   from charms.loki_k8s.v0.log_proxy import LogProxyConsumer
 
    ...
 
        def __init__(self, *args):
            ...
-           try:
-               self._log_proxy = LogProxyConsumer(
-                   charm=self, log_files=LOG_FILES, container_name=PEER, enable_syslog=True
-               )
-           except PromtailDigestError as e:
-               msg = str(e)
+           self._log_proxy = LogProxyConsumer(
+               charm=self, log_files=LOG_FILES, container_name=PEER, enable_syslog=True
+           )
+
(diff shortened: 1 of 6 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -176,18 +189,151 @@ def __init__(self, *args):
...
+
+Which would then allow operators to deploy the charm this way:
+
+```
+juju deploy \
+    ./your_charm.charm \
+    --resource promtail-bin=/tmp/promtail-linux-amd64
+```
+
+The object can raise a `PromtailDigestError` when:
```
Abuelodelanada, line 304: Should we use `raise`?
You raise an Exception, but now we are emitting an Event. Or am I missing something here?

Abuelodelanada (approved):
Nice work @rbarry82 !!

sed-i (dismissed):

dstathis (approved):

dstathis (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1375,33 +1462,27 @@ def _get_container(self, container_name):
...
-            )
-            raise PromtailDigestError(msg)
+            raise ContainerNotFoundError
 
-    def _get_container_name(self, container_name):
+    def _get_container_name(self, container_name: Optional[str] = "") -> str:
         """Gets a container_name.
 
-        If there is more than one container in the Pod a `PromtailDigestError` is raised.
+        If there is more than one container in the Pod a `PromtailDigestError` is emitted.
```
sed-i, line 1485:
```suggestion
        If there is more than one container in the Pod a `ContainerNotFoundError` is raised.
```

sed-i (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: Events are the "canonical" (not as in the company) way we've chosen for libraries to "talk back" to charms.

`raise` was intentionally taken out so the API is the same, as per the request

sed-i (approved):

dstathis (approved):

PR #72, opened 2022-01-18
LogProxyConsumer should inherit from `ConsumerBase`

rbarry82:
Change the inheritance of `LogProxyConsumer` so Alert Rules can
be pulled to the existing machinery.

Add docstrings for the `LogProxyConsumer` class explaining the
arguments and possible exceptions.


sed-i (approved):

mmanciop:
What, no integration test? :-)

dstathis (approved):

Issue #60, opened 2022-01-05
Log proxy does not install promtail on charm upgrade


dstathis:
Title says it all. This should probably be done at the same time as we merge the two libraries.

dstathis:
This actually happens anytime a pod is redeployed. We should probably fix this by observing the `on_pebble_ready` event in the lib.

jnsgruk:
> This actually happens anytime a pod is redeployed. We should probably fix this by observing the `on_pebble_ready` event in the lib.

We probably should use `on_pebble_ready`, though with a guard so that the code only runs if the relevant relation is established?

simskij:
Solved in #74

PR #74, opened 2022-01-20
Libs fix

dstathis:
install promtail if pebble_ready if needed and move client data to unit buckets

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -18,6 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         language: ['python']
+
     steps:
     - name: Checkout repository
       uses: actions/checkout@v2
@@ -26,9 +27,12 @@ jobs:
       uses: github/codeql-action/init@v1
       with:
         languages: ${{ matrix.language }}
-        
+
     - name: Autobuild
       uses: github/codeql-action/autobuild@v1
-      
+
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v1
+
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@v1
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -31,8 +31,5 @@ jobs:
     - name: Autobuild
       uses: github/codeql-action/autobuild@v1
 
-    - name: Autobuild
-      uses: github/codeql-action/autobuild@v1
-
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@v1
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1224,6 +1224,10 @@ class LokiPushApiProvider(RelationManagerBase):
         return alerts
 
 
+class InvalidAlertRulesError(Exception):
+    """Raised when there are problem encountered in processing alert rules."""
+
+
 class ConsumerBase(RelationManagerBase):
     """Consumer's base class."""
 
@@ -1252,8 +1256,34 @@ class ConsumerBase(RelationManagerBase):
         self._recursive = recursive
 
     def _handle_alert_rules(self, relation):
-        if not self._charm.unit.is_leader():
-            return
+        if self._charm.unit.is_leader():
+            alert_groups, invalid_files = load_alert_rules_from_dir(
+                self._alert_rules_path,
+                self.topology,
+                recursive=False,
+                allow_free_standing=self.allow_free_standing_rules,
+            )
+
+            alert_rules_error_message = self._check_alert_rules(alert_groups, invalid_files)
+
+            relation.data[self._charm.app]["metadata"] = json.dumps(self.topology.as_dict())
+            relation.data[self._charm.app]["alert_rules"] = json.dumps({"groups": alert_groups})
+
+            if alert_rules_error_message:
+                raise InvalidAlertRulesError(alert_rules_error_message)
+
+    def _check_alert_rules(self, alert_groups, invalid_files) -> str:
+        """Check alert rules.
+
+        Args:
+            alert_groups: a list of prometheus alert rule groups.
+            invalid_files: a list of invalid rules files.
+
+        Returns:
+            A string with the validation message. The message is not empty whether there are
+            invalid alert rules files or there are no alert rules groups.
+        """
+        message = ""
 
         alert_rules = AlertRules(self.topology)
         alert_rules.add_path(self._alert_rules_path, recursive=self._recursive)
@@ -1358,7 +1388,11 @@ class LokiPushApiConsumer(ConsumerBase):
             self._handle_alert_rules(relation)
 
     def _process_logging_relation_changed(self, relation: Relation):
-        self._handle_alert_rules(relation)
+        try:
+            self._handle_alert_rules(relation)
+        except InvalidAlertRulesError as e:
+            self.on.loki_push_api_alert_rules_error.emit(str(e))
+
         self.on.loki_push_api_endpoint_joined.emit()
 
     def _on_logging_relation_departed(self, _: RelationEvent):
@@ -1534,12 +1568,17 @@ class LogProxyConsumer(RelationManagerBase):
         if not self._is_promtail_installed():
             self._setup_promtail()
 
-    def _on_relation_changed(self, _: RelationEvent) -> None:
+    def _on_relation_changed(self, event: RelationEvent) -> None:
         """Event handler for `relation_changed`.
 
         Args:
             event: The event object `RelationChangedEvent`.
         """
+        try:
+            self._handle_alert_rules(event.relation)
+        except InvalidAlertRulesError as e:
+            logger.error("A problem was encountered when processing alert rules: {}".format(e))
+
         if not self._container.can_connect():
             return
         if self.model.relations[self._relation_name] and not self._is_promtail_installed():
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1483,15 +1483,14 @@ class LogProxyConsumer(ConsumerBase):
 
     def _on_pebble_ready(self, _: WorkloadEvent):
         """Event handler for `pebble_ready`."""
-        if self.model.relations[self._relation_name] and not self._is_promtail_installed():
+        if self.model.relations[self._relation_name]:
             self._setup_promtail()
 
     def _on_relation_created(self, _: RelationCreatedEvent) -> None:
         """Event handler for `relation_created`."""
         if not self._container.can_connect():
             return
-        if not self._is_promtail_installed():
-            self._setup_promtail()
+        self._setup_promtail()
 
     def _on_relation_changed(self, event: RelationEvent) -> None:
         """Event handler for `relation_changed`.
@@ -1503,7 +1502,7 @@ class LogProxyConsumer(ConsumerBase):
 
         if not self._container.can_connect():
             return
-        if self.model.relations[self._relation_name] and not self._is_promtail_installed():
+        if self.model.relations[self._relation_name]:
             self._setup_promtail()
         else:
             new_config = self._promtail_config
@@ -1844,6 +1843,8 @@ class LogProxyConsumer(ConsumerBase):
         return static_configs
 
     def _setup_promtail(self) -> None:
+        if self._is_promtail_installed():
+            return
         relation = self._charm.model.relations[self._relation_name][0]
         if relation.data[relation.app].get("promtail_binary_zip_url", None) is None:
             return
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -448,6 +448,7 @@ logger = logging.getLogger(__name__)
 RELATION_INTERFACE_NAME = "loki_push_api"
 DEFAULT_RELATION_NAME = "logging"
 DEFAULT_ALERT_RULES_RELATIVE_PATH = "./src/loki_alert_rules"
+DEFAULT_LOG_PROXY_RELATION_NAME = "log-proxy"
 
 PROMTAIL_BINARY_ZIP_URL = (
     "https://github.com/grafana/loki/releases/download/v2.4.1/promtail-linux-amd64.zip"
@@ -1454,7 +1455,7 @@ class LogProxyConsumer(ConsumerBase):
         self,
         charm,
         log_files: list = None,
-        relation_name: str = "log-proxy",
+        relation_name: str = DEFAULT_LOG_PROXY_RELATION_NAME,
         enable_syslog: bool = False,
         syslog_port: int = 1514,
         alert_rules_path: str = DEFAULT_ALERT_RULES_RELATIVE_PATH,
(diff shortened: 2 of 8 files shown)


dstathis:
Setting to draft because there is still a bug in `relation_departed`. Just wanted to put this up so people could start to look at it.

pyproject.toml
```diff
@@ -49,11 +52,12 @@ allow_redefinition = true
...
 # Ignore libraries that do not have type hint nor stubs
 [[tool.mypy.overrides]]
-module = ["ops.*", "pytest_operator.*"]
+module = ["ops.*", "pytest_operator.*", "requests.packages.urllib3.*"]
 ignore_missing_imports = true
+follow_imports = "skip"
 
 [[tool.mypy.overrides]]
-module = ["charms.alertmanager_k8s.v0.*", "charms.grafana_k8s.v0.*"]
+module = ["charms.prometheus_k8s.*"]
```
sed-i, line 60: After @rbarry82 publishes the latest prom lib you could probably remove this override.

tests/unit/test_charm.py
```diff
@@ -116,54 +102,48 @@ def test__alerting_config(self):
...
-        self.harness.update_config(yaml.safe_load(LOKI_CONFIG))
-        self.assertIsInstance(self.harness.charm.unit.status, ActiveStatus)
+        self.harness.set_leader(True)
 
-    @patch("ops.testing._TestingPebbleClient.make_dir")
-    @patch("loki_server.LokiServer.version", new_callable=PropertyMock)
-    def test__provide_loki(self, mock_version, *unused):
-        mock_version.return_value = "3.14159"
+        # Since harness was not started with begin_with_initial_hooks(), this must
+        # be emitted by hand to actually trigger _configure()
```
sed-i, line 124: Is there a reason to keep `begin` and not replace it with `begin_with_initial_hooks`?

tests/unit/test_charm.py
```diff
@@ -116,54 +102,48 @@ def test__alerting_config(self):
...
-        self.harness.update_config(yaml.safe_load(LOKI_CONFIG))
+
+        # Since harness was not started with begin_with_initial_hooks(), this must
+        # be emitted by hand to actually trigger _configure()
+        self.harness.charm.on.config_changed.emit()
         self.assertIsInstance(self.harness.charm.unit.status, WaitingStatus)
 
-    @patch("ops.testing._TestingPebbleClient.push")
+    @patch("ops.testing._TestingPebbleClient.push", Mock())
+    @patch("ops.model.Container.can_connect", MagicMock(return_value=True))
```
sed-i, line 117: Haven't tried myself but I expected that these mocks should not be needed.

tests/unit/test_provider.py
```diff
@@ -99,22 +99,26 @@ def setUp(self):
...
-    @patch("ops.testing._TestingPebbleClient.make_dir")
-    @patch("charms.loki_k8s.v0.loki_push_api.LokiPushApiProvider._generate_alert_rules_files")
-    @patch("charms.loki_k8s.v0.loki_push_api.LokiPushApiProvider._remove_alert_rules_files")
+    @patch("ops.testing._TestingPebbleClient.make_dir", Mock())
+    @patch(
+        "charms.loki_k8s.v0.loki_push_api.LokiPushApiProvider._generate_alert_rules_files", Mock()
+    )
+    @patch(
+        "charms.loki_k8s.v0.loki_push_api.LokiPushApiProvider._remove_alert_rules_files", Mock()
+    )
```
sed-i, line 117: Can the test be updated to work without these mocks?

tests/unit/test_provider.py
```diff
@@ -126,13 +130,13 @@ def test__on_logging_relation_changed(self, mock_unit_ip, *unused):
                 "DEBUG:charms.loki_k8s.v0.loki_push_api:Saved alerts rules to disk",
             )
 
-    @patch("ops.testing._TestingPebbleClient.make_dir")
-    @patch("os.makedirs")
-    @patch("ops.testing._TestingPebbleClient.remove_path")
+    @patch("ops.testing._TestingPebbleClient.make_dir", Mock())
+    @patch("os.makedirs", Mock())
+    @patch("ops.testing._TestingPebbleClient.remove_path", Mock())
```
sed-i, line 135: Can the test be updated to work without these mocks?

tox.ini
```diff
@@ -39,82 +40,63 @@ commands =
...
+    types-dataclasses
     types-PyYAML
-    pytest
-    pytest-operator
-    juju
-    types-setuptools
-    types-toml
     types-requests
+    charm: -r{toxinidir}/requirements.txt
+    lib: git+https://github.com/canonical/operator#egg=ops
```
sed-i, line 70: This is a bit risky because requirements.txt is from pypi. The two should probably match.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,64 @@
+# For most projects, this workflow file will not need changing; you simply need
```
simskij, line 1: Remove the comments.

tox.ini
```diff
@@ -39,82 +40,63 @@ commands =
...
 deps =
+    aiohttp
     git+https://github.com/juju/python-libjuju.git
     #juju
     pytest
     git+https://github.com/charmed-kubernetes/pytest-operator.git
 commands =
-    pytest -v --tb native --log-cli-level=INFO -s {posargs} {toxinidir}/tests/integration
-
-[testenv:integration-lma]
```
simskij, line 97: why are we dropping bundle tests?

simskij (changes requested):

tests/unit/test_charm.py
rbarry82: For one, we need to wait for a [patch to the storage harness to merge](https://github.com/canonical/operator/pull/680) or the tests will blow up since the storage name uses hyphens.

The other is the [sheer amount of stuff happening in __init__](https://github.com/canonical/loki-k8s-operator/blob/829195d9db6e3b02687c170648afc16959992b8f/src/charm.py#L68). It would take a lot of mocking in the initial `setUp` to ensure that we could "pretend" that the k8s service was patched, that the Loki server was available (or mocked), etc. 

It's not impossible to do, just a lot of more changes in mocking instead of manually emitting one event here.

tests/unit/test_charm.py
rbarry82: `.push` mocks were already removed in a previous patch. I'll also remove `.can_connect`

tests/unit/test_provider.py
rbarry82: Yes, except for `makedirs`

tests/unit/test_provider.py
rbarry82: No

tox.ini
rbarry82: The bundle tests in this were more or less directly ripped from Prometheus, and don't even exercise Loki.

"Real" bundle integration tests will come later in a different repo which can tie everything together, as part of the pyramid testing strategy.

simskij (approved):

tox.ini
rbarry82: This also matches the other repos (alert manager, prometheus, etc). If we want to fix it, let's do a unified change later.

simskij:
Why is this PR moving the client data to unit data bags? I remember the conclusion last we discussed this being that it serves no obvious purpose and also breaks the compatibility. Clustered Loki will be done a bit differently to what you might expect (not just a `juju scale-application`) as we'll need it to be able to run in multiple clusters and possibly even on multiple clouds.

dstathis:
What this PR does:
Fixes #60 by installing Promtail in pebble_ready if needed.
Fixes a bug in relation_departed where Promtail does not stop properly.
Changes the relation data schema to properly support HA

What this PR does not do:
Fixes the bug I just found (#78) where LokiPushApiProvider does not correctly handle IP address changes.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1282,31 +1250,35 @@ def __init__(
...
         )
+        self.framework.observe(
+            getattr(self._charm.on, "{}_pebble_ready".format(self._container_name)),
+            self._on_pebble_ready,
+        )
+
+    def _on_pebble_ready(self, event):
+        """Event handler for `pebble_ready`."""
+        if self.model.relations[self._relation_name] and not self._is_promtail_installed():
+            self._setup_promtail()
```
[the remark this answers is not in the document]
dstathis: I agree it would be nice but I don't think it is important enough to change at the moment.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1282,31 +1250,35 @@ def __init__(
...
-            else:
-                self._update_config(event)
-                self._update_agents_list(event)
-                self._add_pebble_layer()
+        if self.model.relations[self._relation_name] and not self._is_promtail_installed():
+            self._setup_promtail()
+        else:
+            new_config = self._promtail_config
+            if new_config != self._current_config:
+                self._container.push(WORKLOAD_CONFIG_PATH, yaml.safe_dump(new_config))
```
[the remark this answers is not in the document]
dstathis: We have one a few lines up.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1315,10 +1287,11 @@ def _on_log_proxy_relation_departed(self, event):
...
         """
-        self._update_config(event)
-        self._update_agents_list(event)
+        new_config = self._promtail_config
+        if new_config != self._current_config:
+            self._container.push(WORKLOAD_CONFIG_PATH, yaml.safe_dump(new_config))
 
-        if len(self._current_config["clients"]) == 0:
+        if not self._charm.model.relations[self._relation_name]:
             self._container.stop(WORKLOAD_SERVICE_NAME)
```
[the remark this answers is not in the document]
dstathis: If we try to restart here, Promtail will crash and we will have an error state so we need to check.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1408,13 +1381,13 @@ def _create_directories(self) -> None:
...
 
-    def _obtain_promtail(self, event) -> None:
+    def _obtain_promtail(self) -> None:
         """Obtain promtail binary from an attached resource or download it."""
         if self._is_promtail_attached():
             return
 
         if self._promtail_must_be_downloaded():
-            self._download_and_push_promtail_to_workload(event)
+            self._download_and_push_promtail_to_workload()
```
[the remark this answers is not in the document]
dstathis: yeah it would be nice

pyproject.toml
dstathis: I'll leave for now and look back in to this after beta release.

tests/unit/test_charm.py
dstathis: Well said Ryan

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1496,81 +1449,86 @@ class LogProxyConsumer(ConsumerBase):
...
     on = LogProxyEvents()
-    _stored = StoredState()
 
     def __init__(
         self,
         charm,
-        log_files: list = [],
-        relation_name: str = "log_proxy",
+        log_files: list = None,
+        relation_name: str = "log-proxy",
```
sed-i, line 1457: Add and use "const" analogous to `DEFAULT_RELATION_NAME`?

sed-i (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1558,75 +1516,30 @@ def _current_config(self) -> dict:
...
-            config = self._remove_client(current_config, agent_url)
-
-        return yaml.safe_dump(config)
-
     @property
-    def _initial_config(self) -> dict:
+    def _promtail_config(self) -> dict:
         """Generates an initial config for Promtail.
 
         This config it's going to be completed with the `client` section
```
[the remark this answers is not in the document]
dstathis: Fixed the docstring. I didn't put the output because I think it would be excessively large for a docstring.

lib/charms/loki_k8s/v0/loki_push_api.py
dstathis: done

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1315,10 +1287,11 @@ def _on_log_proxy_relation_departed(self, event):
         Args:
             event: The event object `RelationDepartedEvent`.
         """
-        self._update_config(event)
-        self._update_agents_list(event)
+        new_config = self._promtail_config
+        if new_config != self._current_config:
+            self._container.push(WORKLOAD_CONFIG_PATH, yaml.safe_dump(new_config))
```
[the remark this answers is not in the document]
dstathis: Very unlikely to come up but you are correct. Fixed.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1282,31 +1250,35 @@ def __init__(
...
     def _on_log_proxy_relation_created(self, event):
         """Event handler for the `log_proxy_relation_created`."""
-        self._create_directories()
-        self._container.push(
-            WORKLOAD_CONFIG_PATH, yaml.safe_dump(self._initial_config), make_dirs=True
-        )
+        if not self._container.can_connect():
+            return
+        if not self._is_promtail_installed():
+            self._setup_promtail()
```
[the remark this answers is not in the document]
dstathis: The can_connect() guard is needed for code outside of _setup_promtail as well so it can't be moved inside. I think _is_installed probably can be though.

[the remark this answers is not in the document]
dstathis: Done for _is_promtail_installed.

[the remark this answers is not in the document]
dstathis: This actually broke things and I'm reverting it as I think it is not the behavior that would be expected from such a method and could break things.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1733,9 +1693,17 @@ def _is_promtail_binary_in_charm(self) -> bool:
...
-        url = json.loads(event.relation.data[event.app].get("data"))["promtail_binary_zip_url"]
+        # Use the first
+        relations = self._charm.model.relations[self._relation_name]
+        if len(relations) > 1:
+            logger.debug(
+                "Multiple log_proxy relations. Getting Promtail from application {}".format(
+                    relations[0].app.name
+                )
+            )
+        url = relations[0].data[relations[0].app].get("promtail_binary_zip_url")
```
sed-i, line 1707: Are the multiple [0] here safe? Looks a bit odd but maybe ok in this context

sed-i (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
dstathis: I agree it looks funny but yeah it's safe think of it as
```
relation = relations[0]
url = relation.data[relation.app].get("promtail_binary_zip_url")
```

dstathis: Which would probably have been better but not worth adding more commits to this PR.

rbarry82 (approved):

PR #82, opened 2022-01-27
Finally, bring loki_push_api in sync with the new topology instead of using its own

rbarry82:
Instead of using its own, merge together the last of the changes
from the PR storm.
    
The one remaining difference is that Loki uses `metadata` as a relation key instead of `scrape_metadata`, which makes sense for naming but must be kept in mind instead of blindly copying and pasting it in the future.

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -627,7 +627,7 @@ class JujuTopology:
         self.unit = unit
 
     @classmethod
-    def from_charm(cls, charm):
+    def from_charm(cls, charm) -> "JujuTopology":
         """Factory method for creating `JujuTopology` children from a given charm.
 
         Args:
@@ -645,7 +645,7 @@ class JujuTopology:
         )
 
     @classmethod
-    def from_relation_data(cls, data: dict):
+    def from_relation_data(cls, data: dict) -> "JujuTopology":
         """Factory method for creating `JujuTopology` children from a dictionary.
 
         Args:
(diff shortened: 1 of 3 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -585,79 +586,194 @@ def _validate_relation_by_interface_and_direction(
...
+        # If a key exists in `rename_keys`, replace the value
+        if rename_keys:
+            ret = OrderedDict(
+                (rename_keys.get(k), v) if rename_keys.get(k) else (k, v) for k, v in ret.items()  # type: ignore
+            )
+
+        return ret
 
-    def as_dict_with_logql_labels(self) -> dict:
+    def as_promql_label_dict(self):
```
sed-i, line 719: promql ain't logql?

```suggestion
    def as_logql_label_dict(self):
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -585,79 +586,194 @@ def _validate_relation_by_interface_and_direction(
...
+            a `AggregatorTopology` object.
+        """
+        return cls(
+            model=model,
+            model_uuid=model_uuid,
+            application=application,
+            unit=unit,
+        )
+
+    def as_promql_label_dict(self):
```
sed-i, line 756:
```suggestion
    def as_logql_label_dict(self):
```

sed-i (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: It's not, but the content is exactly the same.

We should open an issue to just rename this to `as_query_language_label_dict` or similar.

I left it as `as_promql_label_dict` because the difference between `metadata` in the Loki data bags for alerts is difference enough, and "keep all of this stuff in sync across 3 different libraries" (prometheus_remote_write, prometheus_scrape, loki_push_api) is bad enough without needing to remember the caveats for all of them.

simskij (approved):

PR #85, opened 2022-01-28
Move relation data back in to app bucket

dstathis:


balbirthomas (approved):

balbirthomas:
Looks like there are a few mypy errors.

rbarry82:
> Looks like there are a few mypy errors.

Aren't there always...

balbirthomas (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1178,17 +1178,31 @@ def _process_logging_relation_changed(self, relation: Relation):
...
             logger.debug("Saved promtail binary url: %s", self._promtail_binary_url)
+            relation.data[self._charm.app]["endpoints"] = json.dumps(self._endpoints())
+            logger.debug("Saved endpoints in relation data")
 
         if relation.data.get(relation.app).get("alert_rules"):
             logger.debug("Saved alerts rules to disk")
             self._remove_alert_rules_files(self.container)
             self._generate_alert_rules_files(self.container)
 
+    def _endpoints(self):
```
sed-i, line 1192: Nice. I should probably do the same for alertmanager peer relation data.

```suggestion
    def _endpoints(self) -> List[dict]:
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1178,17 +1178,31 @@ def _process_logging_relation_changed(self, relation: Relation):
...
+        """Return a list of Loki Push Api endpoints."""
+        return [{"url": self._url(unit_number=i)} for i in range(self._charm.app.planned_units())]
+
+    def _url(self, unit_number):
```
sed-i, line 1196:
```suggestion
    def _url(self, unit_number) -> str:
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1471,30 +1475,17 @@ def _on_logging_relation_departed(self, _: RelationEvent):
...
     @property
-    def loki_push_api(self) -> List[dict]:
+    def loki_endpoints(self) -> List[dict]:
         """Fetch Loki Push API endpoints sent from LokiPushApiProvider through relation data.
 
         Returns:
             A list with Loki Push API endpoints.
         """
-        endpoints = []
+        endpoints = []  # type: list
```
sed-i, line 1484: What does the type: list comment do? I'd imagine it would need to be identical to the return type.

sed-i (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
dstathis: It makes pypy not annoyed with me

sed-i (approved):

balbirthomas (approved):

PR #89, opened 2022-02-01
Use KubernetesServicePatch from observability_libs

Abuelodelanada:


jnsgruk (dismissed):

src/charm.py
```diff
@@ -48,14 +48,14 @@ def __init__(self, *args):
         super().__init__(*args)
         self._container = self.unit.get_container(self._name)
         self._stored.set_default(k8s_service_patched=False, config="")
+        self.service_patch = KubernetesServicePatch(self, [(f"{self.app.name}", self._port)])
```
sed-i, line 51:
```suggestion
        self.service_patch = KubernetesServicePatch(self, [(self.app.name, self._port)])
```

sed-i (dismissed):

src/charm.py
rbarry82: 👍 

I believe other charms still have this, since there's more or less a 0% chance of them running in a pre-3.8 environment other than the libraries, but still

sed-i (approved):

simskij (approved):

PR #97, opened 2022-02-17
itest upgrade-charm retains rule files

sed-i:
This PR updates the existing integration test to make sure alert rules that were forwarded before an upgrade are still there after an upgrade.

Currently the itest fails because rule files are indeed not persisted across upgrade.

Crossref: OPENG-402


Abuelodelanada:
Following in [this issue ](https://github.com/canonical/loki-k8s-operator/issues/103)

PR #99, opened 2022-02-22
FIX #96 - Only the alert rules of the relations that are deleted are removed.

### LINKED ISSUE #96: `_remove_alert_rules_files` method deletes all alert rules files when a charm departs the relation.
#### Issue

`LokiPushApiProvider` is deleting the entire alert rules directory every time a charm departs the relation. So if we have 2 charms related to Loki and one of them departs the relation, the `_remove_alert_rules_files` deletes everything.

#### Expected Behavior

We need to delete just the files of the charm that departs the relation. 

#### Additional info, logs, etc:

```python
    def _remove_alert_rules_files(self, container: Container) -> None:
        """Remove alert rules files from workload container.

        Args:
            container: Container which has alert rules files to be deleted
        """
        container.remove_path(self._rules_dir, recursive=True)
        logger.debug("Previous Alert rules files deleted")
        # Since container.remove_path deletes the directory itself with its files
        # we should create it again.
        os.makedirs(self._rules_dir, exist_ok=True)
```

#### Possible Fix

In `_remove_alert_rules_files` we should retrieve alert rules from the file system and delete the files of the charm that departs the relation.

#### Steps to Reproduce

Test to verify this works as expected:

1. Deploy Spring Music and Grafana Agent in a separate model
2. Deploy another Spring Music in the same model as Loki
3. Relate the Spring Music instance in the same model as Loki directly to Loki
4. Relate Spring Music to Grafana Agent
5. Relate Grafana Agent to Loki over a CMR
6. Remove the relation
7. Make sure the correct files are deleted

Do the same with a direct relation between Spring Music and Loki. Experiment with different variations.

#### Your Environment

* Charm Version used: `main` branch
* Environment name and version (e.g. Juju, Microk8s, etc):
  *  Juju: 2.9.22
  * microk8s: 1.23/stable
* Operating System and version (e.g Ubuntu 20.04): Ubuntu 20.04

Abuelodelanada:
This PR fixes #96

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1230,8 +1230,8 @@ class LokiPushApiProvider(RelationManagerBase):
             event: a `CharmEvent` in response to which the Loki
                 charm must update its relation data.
         """
-        if event.relation.data.get(event.relation.app):
-            self._remove_alert_rules_files(self.container)
+        if isinstance(event, RelationEvent):
+            self._process_logging_relation_changed(event.relation)
 
     @property
     def _promtail_binary_url(self) -> dict:
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1230,8 +1230,7 @@ class LokiPushApiProvider(RelationManagerBase):
             event: a `CharmEvent` in response to which the Loki
                 charm must update its relation data.
         """
-        if isinstance(event, RelationEvent):
-            self._process_logging_relation_changed(event.relation)
+        self._process_logging_relation_changed(event.relation)
 
     @property
     def _promtail_binary_url(self) -> dict:
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -442,7 +442,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 9
+LIBPATCH = 10
 
 logger = logging.getLogger(__name__)
 
--- a/tests/integration/test_alert_rules.py
+++ b/tests/integration/test_alert_rules.py
@@ -45,6 +45,7 @@ async def test_no_rules_files(ops_test):
     rules_before_relation = await loki_rules(ops_test, app_name)
     assert len(rules_before_relation) == 0
 
+
 @pytest.mark.abort_on_fail
 async def test_first_relation_one_alert_rule(ops_test):
     await asyncio.gather(
@@ -105,6 +106,7 @@ async def test_second_relation_second_alert_rule(ops_test):
 
     rules_after_relation2 = await loki_rules(ops_test, app_name)
     assert len(rules_after_relation2) == 2
+
 
 @pytest.mark.abort_on_fail
 async def test_remove_app_one_alert_rules_is_reteined(ops_test):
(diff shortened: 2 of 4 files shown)


rbarry82 (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1230,8 +1230,8 @@ def _on_logging_relation_departed(self, event: RelationDepartedEvent):
             event: a `CharmEvent` in response to which the Loki
                 charm must update its relation data.
         """
-        if event.relation.data.get(event.relation.app):
-            self._remove_alert_rules_files(self.container)
+        if isinstance(event, RelationEvent):
```
dstathis, line 1233: I'm a bit confused about this. How could it be anything other than a RelationEvent?

tox.ini
```diff
@@ -93,9 +93,8 @@ commands =
 description = Run integration tests
 deps =
     aiohttp
-    git+https://github.com/juju/python-libjuju.git
-    #juju
+    juju
     pytest
-    git+https://github.com/charmed-kubernetes/pytest-operator.git
+    pytest-operator==0.10.1
```
dstathis, line 98: Why are we pinning this?

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: You are right, a missing line from a big refactor that did not survive

rbarry82 (approved):

tests/integration/test_alert_rules.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python3
```
sed-i, line 1: Consider renaming file e.g. `test_multiple_rule_providing_apps`

tests/integration/test_alert_rules.py
```diff
@@ -0,0 +1,117 @@
...
+            apps=[app_name, rules_app], status="active", timeout=1000
+        )
+
+    rules_after_relation = await loki_rules(ops_test, app_name)
+    assert len(rules_after_relation) == 1
+
+
+@pytest.mark.abort_on_fail
+async def test_second_relation_second_alert_rule(ops_test):
+    await asyncio.gather(
```
sed-i, line 82: `async def deploy_rule_providing_app(app_name: str)`?

dstathis (approved):

tests/integration/test_alert_rules.py
```diff
@@ -0,0 +1,117 @@
...
+    assert len(rules_after_relation2) == 2
+
+
+@pytest.mark.abort_on_fail
+async def test_remove_app_one_alert_rules_is_reteined(ops_test):
+    await ops_test.model.applications[rules_app].remove()
+    await ops_test.model.block_until(lambda: rules_app not in ops_test.model.applications)
+
+    rules_after_delete_relation2 = await loki_rules(ops_test, app_name)
+    assert len(rules_after_delete_relation2) == 1
```
sed-i, line 117: If instead you remove `rules_app2` then you could easily have a stronger assertion:
```suggestion
    assert rules_after_delete_relation2 == rules_after_relation
```

sed-i (approved):

tests/integration/test_alert_rules.py
Abuelodelanada: Will do it in: https://github.com/canonical/loki-k8s-operator/issues/103

tests/integration/test_alert_rules.py
Abuelodelanada: Will do it in: https://github.com/canonical/loki-k8s-operator/issues/103

tests/integration/test_alert_rules.py
Abuelodelanada: Will do it in: https://github.com/canonical/loki-k8s-operator/issues/103

PR #105, opened 2022-03-03
FIX #91 - Loki push_api lib should validate alert rules

### LINKED ISSUE #91: Loki push_api lib should validate alert rules have at least one equality matcher
While working on cos-config I found out that [alert rules](https://github.com/canonical/cos-configuration-k8s-operator/tree/main/tests/samples/loki_alert_rules) that I forward to Loki are rejected: Loki sees them but complains about syntax
```
> $ curl 10.152.183.14:3100/loki/api/v1/rules                                                                                                                                                              
failed to list rule group for user fake and namespace 0_cos-configuration-k8s_alert.rules: error parsing /loki/rules/fake/0_cos-configuration-k8s_alert.rules: /loki/rules/fake/0_cos-configuration-k8s_alert.rules: could not parse expression: parse error at line 1, col 1: syntax error: unexpected IDENTIFIER
```
but the only thing in line 1 is `groups:`.

Apparently,

> you need at least one equality matcher
> https://github.com/cortexproject/cortex/issues/3875#issuecomment-790530357

When I replaced the rule with the following, Loki was happy
```yaml
groups:
 - name: example
   rules:
     - alert: HighThroughputLogStreams
       expr: absent_over_time({namespace="dev",job=~".*-logs"}[2m])
       for: 2m
```
Seems like the loki lib needs to check for this so the user could be prompted.

Abuelodelanada:
This PR introduces changes to `LokiPushApiProvider`

- Once a relation is established and alert rules are sent through relation data, the object will check alert rules by querying Loki API. If there are no errors a `loki_push_api_alert_rules_ok` event is emitted. In the other hand if there are errors with alert rules, `loki_push_api_alert_rules_error` event is emitted.
- Loki charm now observes these new events and sets unit status accordingly.
- Refactor in `_remove_alert_rules_files()`. Now we delete one by one alert rule files instead of removing the entire directory.

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -131,8 +131,8 @@ The `LokiPushApiProvider` object has several responsibilities:
 
 Once these alert rules are sent over relation data, the `LokiPushApiProvider` object
 stores these files in the directory `/loki/rules` inside the Loki charm container. After
-storing alert rules files, the object will check alert rules by querying
-Loki API endpoint: [`loki/api/v1/rules`](https://grafana.com/docs/loki/latest/api/#list-rule-groups).
+storing alert rules files, the object will check alert rules by querying Loki API
+endpoint: [`loki/api/v1/rules`](https://grafana.com/docs/loki/latest/api/#list-rule-groups).
 If there are errors with alert rules a `loki_push_api_alert_rules_error` event will
 be emitted, in the other hand if there are no error a `loki_push_api_alert_rules_ok`
 event will be emitted.
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -133,11 +133,11 @@ Once these alert rules are sent over relation data, the `LokiPushApiProvider` ob
 stores these files in the directory `/loki/rules` inside the Loki charm container. After
 storing alert rules files, the object will check alert rules by querying Loki API
 endpoint: [`loki/api/v1/rules`](https://grafana.com/docs/loki/latest/api/#list-rule-groups).
-If there are errors with alert rules a `loki_push_api_alert_rules_error` event will
-be emitted, in the other hand if there are no error a `loki_push_api_alert_rules_ok`
-event will be emitted.
+If there are errors with alert rules a `loki_push_api_alert_rules_changed` event will
+be emitted with `event.error=True`, in the other hand if there are no error a
+`loki_push_api_alert_rules_changed` event will be emitted with `event.error=False`
 
-These events should be observed in the charm that uses `LokiPushApiProvider`:
+This events should be observed in the charm that uses `LokiPushApiProvider`:
 
 ```python
     def __init__(self, *args):
@@ -145,17 +145,17 @@ These events should be observed in the charm that uses `LokiPushApiProvider`:
         ...
         self.loki_provider = LokiPushApiProvider(self)
         self.framework.observe(
-            self.loki_provider.on.loki_push_api_alert_rules_error,
-            self._loki_push_api_alert_rules_error,
-        )
-        self.framework.observe(
-            self.loki_provider.on.loki_push_api_alert_rules_ok, self._loki_push_api_alert_rules_ok
+            self.loki_provider.on.loki_push_api_alert_rules_changed,
+            self._loki_push_api_alert_rules_changed,
         )
 
-    def _loki_push_api_alert_rules_error(self, event):
-        self.unit.status = BlockedStatus("Errors in alert rule groups. Check juju debug-log.")
+    def _loki_push_api_alert_rules_changed(self, event):
+        if event.error:
+            self.unit.status = BlockedStatus(event.message)
+            return
 
-    def _loki_push_api_alert_rules_ok(self, event):
+        ...
+
         self.unit.status = ActiveStatus()
 ```
 
@@ -1129,12 +1129,22 @@ class LokiPushApiEndpointJoined(EventBase):
     """Event emitted when Loki joined."""
 
 
-class LokiPushApiAlertRulesError(EventBase):
+class LokiPushApiAlertRulesChanged(EventBase):
     """Event emitted if there is an error with Alert Rules."""
 
+    def __init__(self, handle, error: bool = False, message: str = ""):
+        super().__init__(handle)
+        self.message = message
+        self.error = error
 
-class LokiPushApiAlertRulesOK(EventBase):
-    """Event emitted if there is no error with Alert Rules."""
+    def snapshot(self) -> Dict:
+        """Save grafana source information."""
+        return {"error": self.error, "message": self.message}
+
+    def restore(self, snapshot):
+        """Restore grafana source information."""
+        self.message = snapshot["message"]
+        self.error = snapshot["error"]
 
 
 class LokiPushApiEvents(ObjectEvents):
@@ -1142,8 +1152,7 @@ class LokiPushApiEvents(ObjectEvents):
 
     loki_push_api_endpoint_departed = EventSource(LokiPushApiEndpointDeparted)
     loki_push_api_endpoint_joined = EventSource(LokiPushApiEndpointJoined)
-    loki_push_api_alert_rules_error = EventSource(LokiPushApiAlertRulesError)
-    loki_push_api_alert_rules_ok = EventSource(LokiPushApiAlertRulesOK)
+    loki_push_api_alert_rules_changed = EventSource(LokiPushApiAlertRulesChanged)
 
 
 class LokiPushApiProvider(RelationManagerBase):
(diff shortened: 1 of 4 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1223,6 +1264,48 @@ def _url(self, unit_number) -> str:
...
+        url = "http://127.0.0.1:{}/loki/api/v1/rules".format(self.port)
+        req = request.Request(url)
+        try:
+            request.urlopen(req)
+        except HTTPError as e:
+            msg = e.msg  # type: ignore
+            if e.code == 404 and "no rule groups found" in msg:
+                logger.debug("Checking alert rules: No rule groups found")
+                self.on.loki_push_api_alert_rules_ok.emit()
+                return True
```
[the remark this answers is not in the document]
dstathis: I think having the return here makes the code a bit easier to read.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1223,6 +1264,48 @@ def _url(self, unit_number) -> str:
...
+            self.on.loki_push_api_alert_rules_error.emit()
+            return False
+        except URLError as e:
+            logger.error("Checking alert rules: %s", e.reason)
+            self.on.loki_push_api_alert_rules_error.emit()
+            return False
+        else:
+            logger.debug("Checking alert rules: Ok")
+            self.on.loki_push_api_alert_rules_ok.emit()
+            return True
```
dstathis, line 1305: While this is certainly good, it could be improved further. We could check that the rules actually match what is supposed to be there. What happens when you give Loki one good and one bad alert rule? I think this check would pass.

src/charm.py
```diff
@@ -81,6 +89,12 @@ def _on_loki_pebble_ready(self, event):
     def _on_alertmanager_change(self, event):
         self._configure(event)
 
+    def _loki_push_api_alert_rules_error(self, event):
+        self.unit.status = BlockedStatus("Errors in alert rule groups. See debug-log.")
+
+    def _loki_push_api_alert_rules_ok(self, event):
+        self.unit.status = ActiveStatus()
```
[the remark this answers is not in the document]
dstathis: Yeah we need to check the current message (is this possible?) and set active only if it is `"Errors in alert rule groups. See debug-log."`

src/charm.py
```diff
@@ -154,7 +168,7 @@ def _provide_loki(self):
         """Gets LokiPushApiProvider instance into `self.loki_provider`."""
         try:
             version = self._loki_server.version
-            self.loki_provider = self.loki_provider or LokiPushApiProvider(self)
+            # self.loki_provider = self.loki_provider or LokiPushApiProvider(self)
```
dstathis, line 171: Just delete this

lib/charms/loki_k8s/v0/loki_push_api.py
dstathis: I am okay merging with this as is, but just wanted to mention it could be improved.

dstathis (changes requested):

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: > What happens when you give Loki one good and one bad alert rule?

Querying the endpoint `/loki/api/v1/rules` will return a `400` with:

```failed to list rule group for user fake and namespace loki_c937c803-173d-4c1f-896d-03ae800a0c82_prom_prom_0_promtail-k8s_alert.rules: error parsing /loki/rules/fake/loki_c937c803-173d-4c1f-896d-03ae800a0c82_prom_prom_0_promtail-k8s_alert.rules: /loki/rules/fake/loki_c937c803-173d-4c1f-896d-03ae800a0c82_prom_prom_0_promtail-k8s_alert.rules: could not parse expression: parse error at line 0, col 11: syntax error: unexpected RANGE, expecting NUMBER or { or (```

As far as I understand when querying `/loki/api/v1/rules` Loki returns alert rules that are stored in `/loki/rules/fake` directory...  @simskij  Do you think it's worth checking again here?

File: lib/charms/loki_k8s/v0/loki_push_api.py (hunk 1 above)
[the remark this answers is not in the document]
Comment (Abuelodelanada) [reply lib/charms/loki_k8s/v0/loki_push_api.py:1287]: We do not have an `else` with this `if`, so the return is necessary ;-)

File: lib/charms/loki_k8s/v0/loki_push_api.py [hunk 5]
```diff
@@ -1101,16 +1129,28 @@ class LokiPushApiEndpointJoined(EventBase):
...
+    """Event emitted if there is no error with Alert Rules."""
+
+
 class LokiPushApiEvents(ObjectEvents):
     """Event descriptor for events raised by `LokiPushApiProvider`."""
 
     loki_push_api_endpoint_departed = EventSource(LokiPushApiEndpointDeparted)
     loki_push_api_endpoint_joined = EventSource(LokiPushApiEndpointJoined)
+    loki_push_api_alert_rules_error = EventSource(LokiPushApiAlertRulesError)
+    loki_push_api_alert_rules_ok = EventSource(LokiPushApiAlertRulesOK)
```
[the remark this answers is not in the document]
Abuelodelanada: I have no strong opinions about this.
Having only one event is handy since charm authors will only observe this event, but in the other hand they will have to add logic in the event handler to decide what to do.

Whatever approach we choose, it would be good to keep it in all charms. 
What do you think @simskij ?

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i: I'd imagine this responsibility belongs to loki and hopefully in the near future there will be api / web ui around invalid config details.

lib/charms/loki_k8s/v0/loki_push_api.py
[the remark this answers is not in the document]
rbarry82: I'm obviously biased, but my perspective is that a single event which contains `errors` or some other indication which can be logged/put into `juju status`, and an attribute for "everything is ok now/in the first place" is a nicer API than listening to `loki_push_api_alert_rules_ok`, which will (hopefully) always be true

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: Grafana is the web API, generally. If it changes in the future, I'd imagine that it will end up like the change to Grafana, which now returns a 200 OK along with a JSON message saying "actually it's not ok."

An integration test is important here.

src/charm.py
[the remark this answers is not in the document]
rbarry82: Nicer to have `"... Check juju debug-log"`, since it may otherwise be unclear which "log" to look at (pod/juju/etc)

lib/charms/loki_k8s/v0/loki_push_api.py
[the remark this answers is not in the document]
Abuelodelanada: I think we should discuss this in a Café

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: Integration test added.

Abuelodelanada:
> Add itest?

Added... now we are getting error alert rules from `error_alerts` branch... but I believe they should be in `main` or faking error alert rules.

tests/integration/test_alert_rules.py
```diff
@@ -115,3 +116,29 @@ async def test_remove_app_one_alert_rules_is_reteined(ops_test):
 
     rules_after_delete_relation2 = await loki_rules(ops_test, app_name)
     assert len(rules_after_delete_relation2) == 1
+
+
+@pytest.mark.abort_on_fail
+async def test_wrong_alert_rule(ops_test):
+    await asyncio.gather(
```
sed-i, line 123: No need for gather here (only one awaitable)

sed-i (dismissed):

sed-i (approved):

rbarry82 (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1223,6 +1273,57 @@ def _url(self, unit_number) -> str:
...
+            loki_push_api_alert_rules_changed: This event is emitted when alert rules are checked.
+        """
+        url = "http://127.0.0.1:{}/loki/api/v1/rules".format(self.port)
+        req = request.Request(url)
+        try:
+            request.urlopen(req)
+        except HTTPError as e:
+            msg = e.read().decode("utf-8")
+
+            if e.code == 404 and "no rule groups found" in msg:
```
sed-i, line 1291: Should we open an Issue on Loki asking to turn this into proper json?
And mention that issue in a comment here?

sed-i (approved):

rbarry82 (approved):

dstathis (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: I'm going to ask in the community chat first

PR #108, opened 2022-03-11
Adding alert rule with a syntax error for testing

Abuelodelanada:
This simple PR adds an alert rule with syntax error for testing purposes

### DIFF
--- a/tests/sample_rule_files/free-standing/error/alert.rule
+++ b/tests/sample_rule_files/free-standing/error/alert.rule
@@ -1,0 +1,20 @@
+groups:
+  - name: should_fire
+    rules:
+      - alert: HighPercentageError
+        expr: |
+          sum(rate({) by (job) > 0.05
+        for: 10m
+        labels:
+            severity: page
+        annotations:
+            summary: High request latency
+  - name: credentials_leak
+    rules:
+      - alert: http-credentials-leaked
+        annotations:
+          message: "{{ $labels.job }} is leaking http basic auth credentials."
+        expr: 'sum by ("http(s?)://(\\w+):(\\w+)@" [5m) > 0)'
+        for: 10m
+        labels:
+          severity: critical


tests/sample_rule_files/free-standing/error/alert.rule
```diff
@@ -0,0 +1,20 @@
+groups:
```
sed-i, line 1: Is the subdir needed? I.e. would `free-standing/alert_with_syntax_error_in_expr.rule` work?

sed-i (approved):
I wonder if these kinds of tests would be more manageable if:
- we keep them under the cos bundle repo
- we use cos-config with rule files we write to disk at runtime (git-sync supports local folders for syncing git repos)

tests/sample_rule_files/free-standing/error/alert.rule
Abuelodelanada: I believe it's better to have them in a separate directory so we can use cos-config in itests:

```
        ops_test.model.deploy(
            "ch:cos-configuration-k8s",
            application_name=rules_app3,
            channel="edge",
            config={
                "git_repo": "https://github.com/canonical/loki-k8s-operator",
                "git_branch": "main",
                "loki_alert_rules_path": "tests/sample_rule_files/free-standing/error",
            },
        ),
    )
```

Abuelodelanada:
> I wonder if these kinds of tests would be more manageable if:
> 
>     * we keep them under the cos bundle repo

I do not know... checking alert rules is something Loki lib does...

> 
>     * we use cos-config with rule files we write to disk at runtime (git-sync supports local folders for syncing git repos)

Oh, if this is possible, it is a better idea. Do we have an example of this?

mmanciop:
By which test are these used? It’s unclear from the PR if they are used

Abuelodelanada:
> By which test are these used? It’s unclear from the PR if they are used

By a test I'm writing in which cos-config clones loki repo, so I need the error alert rule in `main` beforehand.

mmanciop:
Mmm this sounds like a weird setup. Why not putting these data into a different, protected branch of cos-config instead?

dstathis (approved):

simskij:
So am I correct in understanding that this actually is to be used by cos-config, which is then pointed here to get the alert rules?

My initial reaction is that this sounds unnecessarily complicated. Why is cos-config even needed in this case?

simskij:
> Mmm this sounds like a weird setup. Why not putting these data into a different, protected branch of cos-config instead?

I would say it's preferable if the test resources live near the code that uses them. If anything, I would like to avoid involving cos-config at all for these tests, although I'm not sure whether that would be feasible.

simskij (approved):
lgtm. let's get this merged. we can iterate on the specifics later.

PR #110, opened 2022-03-17
Handle empty endpoints list

dstathis:
I got a temporary error state from the loki_push_api relation when the remote unit had not yet published endpoints. That caused an exception to be raised. This PR fixes that.

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -2008,9 +2008,7 @@ class LogProxyConsumer(ConsumerBase):
         """
         clients = []  # type: list
         for relation in self._charm.model.relations.get(self._relation_name, []):
-            endpoints = json.loads(relation.data[relation.app].get("endpoints", ""))
-            if endpoints:
-                clients += endpoints
+            clients = clients + json.loads(relation.data[relation.app].get("endpoints", "[]"))
         return clients
 
     def _server_config(self) -> dict:
--- a/tests/unit/test_log_proxy_consumer.py
+++ b/tests/unit/test_log_proxy_consumer.py
@@ -167,6 +167,12 @@ class TestLogProxyConsumer(unittest.TestCase):
             {x["url"] for x in self.harness.charm._log_proxy._clients_list()}, expected_clients
         )
 
+    def test__empty_clients_list(self):
+        # Ensure we do not raise an exception if the clients list is empty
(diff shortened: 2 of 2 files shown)


tests/unit/test_log_proxy_consumer.py
```diff
@@ -167,6 +167,12 @@ def test__add_client(self):
             {x["url"] for x in self.harness.charm._log_proxy._clients_list()}, expected_clients
         )
 
+    def test__empty_clients_list(self):
+        # Ensure we do not raise an exception if the clients list is empty
+        rel_id = self.harness.add_relation("log-proxy", "agent")
+        self.harness.add_relation_unit(rel_id, "agent/0")
+        self.assertEqual(self.harness.charm._log_proxy._clients_list(), [])
```
sed-i, line 174: Is there a more indirect way of testing this? Prefer blackbox testing over whitebox testing.
For example:
```
class TestRelatedEndpoints:
    def test_no_endpoints
    def test_one_invalid_endpoint
    def test_endpoints_without_scheme
```

sed-i (approved):

tests/unit/test_log_proxy_consumer.py
simskij: I agree with Leon on this. Can we assess this using a public API instead?

simskij (dismissed):

Abuelodelanada (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -2001,7 +2001,7 @@ def _clients_list(self) -> list:
         """
         clients = []  # type: list
         for relation in self._charm.model.relations.get(self._relation_name, []):
-            clients = clients + json.loads(relation.data[relation.app]["endpoints"])
+            clients = clients + json.loads(relation.data[relation.app].get("endpoints", "[]"))
```
balbirthomas, line 2004: Is this change still required ? It seems to be out of sync with current main branch. Please see [this line](https://github.com/canonical/loki-k8s-operator/blob/main/lib/charms/loki_k8s/v0/loki_push_api.py#L2011) on main branch.

mmanciop: The line you linked if bugged out big times. Passing a `""` to JSON parsing and trying to treat it as a list to concatenate to another? :D

rbarry82:
Handled as part of #150

PR #111, opened 2022-03-18
Correctly treat empty rule files

### LINKED ISSUE #98: loki_push_api does not like loki_alert_rules being an empty file.
`_is_official_alert_rule_format(rule_file)`
raises an uncaught `TypeError: argument of type 'NoneType' is not iterable`
when loki_alert_rules is an empty file. 

Is that a valid 'rules' file to begin with?
Either way, looks like we should return `list()` in `_from_file` even if the file is invalid/empty.

sed-i:
`yaml.safe_load` may return a None. This PR makes sure this case is correctly handled.

Resolves #98.

Crossref: OPENG-649

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -907,7 +907,7 @@ class AlertRules:
         with file_path.open() as rf:
             # Load a list of rules from file then add labels and filters
             try:
-                rule_file = yaml.safe_load(rf)
+                rule_file = yaml.safe_load(rf) or {}
 
             except Exception as e:
                 logger.error("Failed to read alert rules from %s: %s", file_path.name, e)
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -921,7 +921,8 @@ class AlertRules:
                 alert_groups = [{"name": file_path.stem, "rules": [rule_file]}]
             else:
                 # invalid/unsupported
-                logger.error("Invalid rules file: %s", file_path.name)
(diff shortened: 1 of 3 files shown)


tests/unit/test_consumer.py
```diff
@@ -9,6 +9,7 @@
 
 import yaml
 from charms.loki_k8s.v0.loki_push_api import LokiPushApiConsumer
+from fs.tempfs import TempFS
 from helpers import TempFolderSandbox
```
sed-i, line 13: If no one objects, I will drop `TempFolderSandbox` in favor of `TempFS` as part of this PR.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -907,7 +907,7 @@ def _from_file(self, root_path: Path, file_path: Path) -> List[dict]:
         with file_path.open() as rf:
             # Load a list of rules from file then add labels and filters
             try:
-                rule_file = yaml.safe_load(rf)
+                rule_file = yaml.safe_load(rf) or {}
```
mmanciop, line 910: Sounds to me like a case in why we would log at least a debug. It is unlikely to be intended to have an empty rule file.

sed-i: An ERROR is logged in this case, line 924.
This behavior (logging an error) is also anchored in the utests.

mmanciop: L924 says that it is invalid. It does not say _why_.

sed-i: Added.

mmanciop (approved):

Abuelodelanada (approved):

PR #119, opened 2022-03-24
Update alert rules to not use "juju_unit" as part of the topology

dstathis:
This makes it so that all units will trigger alerts instead of just the leader.

## Issue
When using the loki_push_api library, "juju_unit" is added to the alert rules. This causes the rule to only apply to the leader and not to the other unit.

## Solution
I used the updated JujuTopology class copied from prometheus_scrape with a minor modification to make it work in loki_push_api. This removes the "juju_unit" key from the alert rules.

## Context
Similar to [this PR](https://github.com/canonical/prometheus-k8s-operator/pull/239)

## Testing Instructions
Deploy a charm that has loki alert rules defined. (cassandra-k8s from the "update_for_cos" branch has this)
Relate to loki-k8s
Use `juju show-unit` and `curl http://address/loki/api/v1/rules` to verify the labels and expression are both correct.

## Release Notes
* Update alert rules to not use "juju_unit" as part of the topology. This makes it so that all units will trigger alerts instead of just the leader.

## Additional
In lieu of an integration test which Balbir is currently writing, here is the output of `show-unit` while using the updated lib
```
loki/0:
  opened-ports: []
  charm: local:focal/loki-k8s-2
  leader: true
  relation-info:
  - endpoint: logging
    related-endpoint: log-proxy
    application-data:
      alert_rules: '{"groups": [{"name": "lma_f6eed17b-0b64-41b4-87b5-c56390663aa9_cass_cassandra-k8s_alert_on_error_alerts",
        "rules": [{"alert": "alert_on_error", "annotations": {"summary": "Logs found
        at ERROR level"}, "expr": "rate({juju_model=\"lma\", juju_model_uuid=\"f6eed17b-0b64-41b4-87b5-c56390663aa9\",
        juju_application=\"cass\", juju_charm=\"cassandra-k8s\"}|=\"ERROR\"[5m]) >
        0", "for": "1m", "labels": {"juju_application": "cass", "juju_charm": "cassandra-k8s",
        "juju_model": "lma", "juju_model_uuid": "f6eed17b-0b64-41b4-87b5-c56390663aa9",
        "severity": "critical"}}]}]}'
      metadata: '{"model": "lma", "model_uuid": "f6eed17b-0b64-41b4-87b5-c56390663aa9",
        "application": "cass", "unit": "cass/0", "charm_name": "cassandra-k8s"}'
    related-units:
      cass/0:
        in-scope: true
        data:
          egress-subnets: 10.152.183.122/32
          ingress-address: 10.152.183.122
          private-address: 10.152.183.122
  - endpoint: logging
    related-endpoint: logging
    application-data:
      alert_rules: '{"groups": [{"name": "lma_f6eed17b-0b64-41b4-87b5-c56390663aa9_spring_spring-music_alert_on_error_alerts",
        "rules": [{"alert": "alert_on_error", "annotations": {"summary": "Logs found
        at ERROR level"}, "expr": "rate({juju_model=\"lma\", juju_model_uuid=\"f6eed17b-0b64-41b4-87b5-c56390663aa9\",
        juju_application=\"spring\", juju_charm=\"spring-music\"}|=\"ERROR\"[5m])
        > 0", "for": "1m", "labels": {"juju_application": "spring", "juju_charm":
        "spring-music", "juju_model": "lma", "juju_model_uuid": "f6eed17b-0b64-41b4-87b5-c56390663aa9",
        "severity": "critical"}}]}]}'
      metadata: '{"model": "lma", "model_uuid": "f6eed17b-0b64-41b4-87b5-c56390663aa9",
        "application": "spring", "unit": "spring/0", "charm_name": "spring-music"}'
    related-units:
      spring/0:
        in-scope: true
        data:
          egress-subnets: 10.152.183.222/32
          ingress-address: 10.152.183.222
          private-address: 10.152.183.222
  provider-id: loki-0
  address: 10.1.121.164
```

### DIFF
--- a/tests/unit/test_consumer.py
+++ b/tests/unit/test_consumer.py
@@ -292,6 +292,7 @@ class TestAlertRuleFormat(unittest.TestCase):
         class ConsumerCharm(CharmBase):
             metadata_yaml = textwrap.dedent(
                 """
+                name: loki-consumer-k8s
                 requires:
                   logging:
                     interface: loki_push_api
@@ -358,3 +359,89 @@ class TestAlertRuleFormat(unittest.TestCase):
         logger_output = "\n".join(logger.output)
         self.assertIn("tab.rule", logger_output)
         self.assertIn("multicolon.rule", logger_output)
+
+    def test_rules_have_correct_labels(self):
+        unlabeled_rule = {
+            "groups": [
+                {
+                    "name": "alert_on_error",
+                    "rules": [
+                        {
+                            "alert": "alert_on_error",
+                            "expr": 'rate({%%juju_topology%%}|="ERROR"[5m]) > 0',
+                            "for": "1m",
+                            "labels": {
+                                "severity": "critical",
+                            },
+                            "annotations": {"summary": "Logs found at ERROR level"},
+                        }
+                    ],
+                }
+            ]
+        }
+        self.sandbox.writetext("error.rules", yaml.dump(unlabeled_rule))
+        self.harness.begin_with_initial_hooks()
+        relation = self.harness.charm.model.get_relation("logging")
+        rules = json.loads(relation.data[self.harness.charm.app].get("alert_rules"))
+        import pprint
+
+        pprint.pprint(rules)
+        expr = rules["groups"][0]["rules"][0]["expr"]
+        self.assertTrue(
+            "juju_model" in expr
+            and "juju_model_uuid" in expr
+            and "juju_application" in expr
+            and "juju_charm" in expr
+            and "juju_unit" not in expr
+        )
+        self.assertEqual(
+            set(rules["groups"][0]["rules"][0]["labels"]),
+            {"juju_application", "juju_charm", "juju_model", "juju_model_uuid", "severity"},
+        )
+
+    def test_rules_have_correct_labels_when_unit_is_set(self):
+        unlabeled_rule = {
+            "groups": [
+                {
+                    "name": "alert_on_error",
+                    "rules": [
+                        {
+                            "alert": "alert_on_error",
+                            "expr": 'rate({%%juju_topology%%, juju_unit="app/0"}|="ERROR"[5m]) > 0',
+                            "for": "1m",
+                            "labels": {
+                                "severity": "critical",
+                                "juju_unit": "app/0",
+                            },
+                            "annotations": {"summary": "Logs found at ERROR level"},
+                        }
+                    ],
+                }
+            ]
+        }
+        self.sandbox.writetext("error.rules", yaml.dump(unlabeled_rule))
+        self.harness.begin_with_initial_hooks()
+        relation = self.harness.charm.model.get_relation("logging")
+        rules = json.loads(relation.data[self.harness.charm.app].get("alert_rules"))
+        import pprint
+
+        pprint.pprint(rules)
+        expr = rules["groups"][0]["rules"][0]["expr"]
+        self.assertTrue(
+            "juju_model" in expr
+            and "juju_model_uuid" in expr
+            and "juju_application" in expr
+            and "juju_charm" in expr
+            and "juju_unit" in expr
+        )
+        self.assertEqual(
+            set(rules["groups"][0]["rules"][0]["labels"]),
+            {
+                "juju_application",
+                "juju_charm",
+                "juju_model",
+                "juju_model_uuid",
+                "severity",
+                "juju_unit",
+            },
+        )
--- a/tests/unit/test_consumer.py
+++ b/tests/unit/test_consumer.py
@@ -383,9 +383,6 @@ class TestAlertRuleFormat(unittest.TestCase):
         self.harness.begin_with_initial_hooks()
         relation = self.harness.charm.model.get_relation("logging")
         rules = json.loads(relation.data[self.harness.charm.app].get("alert_rules"))
-        import pprint
-
-        pprint.pprint(rules)
         expr = rules["groups"][0]["rules"][0]["expr"]
         self.assertTrue(
             "juju_model" in expr
@@ -423,9 +420,6 @@ class TestAlertRuleFormat(unittest.TestCase):
         self.harness.begin_with_initial_hooks()
         relation = self.harness.charm.model.get_relation("logging")
         rules = json.loads(relation.data[self.harness.charm.app].get("alert_rules"))
-        import pprint
-
-        pprint.pprint(rules)
         expr = rules["groups"][0]["rules"][0]["expr"]
         self.assertTrue(
             "juju_model" in expr
--- a/tests/unit/test_consumer.py
+++ b/tests/unit/test_consumer.py
@@ -384,13 +384,11 @@ class TestAlertRuleFormat(unittest.TestCase):
         relation = self.harness.charm.model.get_relation("logging")
         rules = json.loads(relation.data[self.harness.charm.app].get("alert_rules"))
         expr = rules["groups"][0]["rules"][0]["expr"]
-        self.assertTrue(
-            "juju_model" in expr
-            and "juju_model_uuid" in expr
-            and "juju_application" in expr
-            and "juju_charm" in expr
-            and "juju_unit" not in expr
-        )
+        self.assertIn("juju_model", expr)
+        self.assertIn("juju_model_uuid", expr)
+        self.assertIn("juju_application", expr)
+        self.assertIn("juju_charm", expr)
+        self.assertNotIn("juju_unit", expr)
         self.assertEqual(
             set(rules["groups"][0]["rules"][0]["labels"]),
             {"juju_application", "juju_charm", "juju_model", "juju_model_uuid", "severity"},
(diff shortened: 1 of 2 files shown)


mmanciop (dismissed):
I think this PR deserves a unit test or two.

sed-i:
> I think this PR deserves a unit test or two.

@dstathis see https://github.com/canonical/prometheus-k8s-operator/blob/8b8e23b929e27d6ffc909c79dc480cb39726c760/tests/unit/test_endpoint_provider.py#L560

dstathis:
> > I think this PR deserves a unit test or two.
> 
> @dstathis see https://github.com/canonical/prometheus-k8s-operator/blob/8b8e23b929e27d6ffc909c79dc480cb39726c760/tests/unit/test_endpoint_provider.py#L560

Balbir is currently writing integration tests for this. I had assumed we don't need both. What do people think?

mmanciop:
> > > I think this PR deserves a unit test or two.
> > 
> > 
> > @dstathis see https://github.com/canonical/prometheus-k8s-operator/blob/8b8e23b929e27d6ffc909c79dc480cb39726c760/tests/unit/test_endpoint_provider.py#L560
> 
> Balbir is currently writing integration tests for this. I had assumed we don't need both. What do people think?

I think that, for a bug of this severity, there ain't ever enough safeguards :-)

dstathis:
> > > > I think this PR deserves a unit test or two.
> > > 
> > > 
> > > @dstathis see https://github.com/canonical/prometheus-k8s-operator/blob/8b8e23b929e27d6ffc909c79dc480cb39726c760/tests/unit/test_endpoint_provider.py#L560
> > 
> > 
> > Balbir is currently writing integration tests for this. I had assumed we don't need both. What do people think?
> 
> I think that, for a bug of this severity, there ain't ever enough safeguards :-)

Unit tests added.

tests/unit/test_consumer.py
```diff
@@ -358,3 +359,89 @@ def test_rules_files_with_invalid_yaml_are_dropped_and_produce_an_error(self):
...
+                }
+            ]
+        }
+        self.sandbox.writetext("error.rules", yaml.dump(unlabeled_rule))
+        self.harness.begin_with_initial_hooks()
+        relation = self.harness.charm.model.get_relation("logging")
+        rules = json.loads(relation.data[self.harness.charm.app].get("alert_rules"))
+        import pprint
+
+        pprint.pprint(rules)
```
sed-i, line 388: Leftover?

tests/unit/test_consumer.py
```diff
@@ -358,3 +359,89 @@ def test_rules_files_with_invalid_yaml_are_dropped_and_produce_an_error(self):
...
+        expr = rules["groups"][0]["rules"][0]["expr"]
+        self.assertTrue(
+            "juju_model" in expr
+            and "juju_model_uuid" in expr
+            and "juju_application" in expr
+            and "juju_charm" in expr
+            and "juju_unit" not in expr
+        )
```
sed-i, line 396: The error message in case of a failure would be considerably more readable if this is split into multiple `assertIn`s and an `assertNotIn`.

tests/unit/test_consumer.py
```diff
@@ -358,3 +359,89 @@ def test_rules_files_with_invalid_yaml_are_dropped_and_produce_an_error(self):
...
+        self.assertEqual(
```
sed-i, line 397: We can use inequality to reduce fragility:
```suggestion
        self.assertGreaterEqual(
```

tests/unit/test_consumer.py
```diff
@@ -358,3 +359,89 @@ def test_rules_files_with_invalid_yaml_are_dropped_and_produce_an_error(self):
...
+                    "name": "alert_on_error",
+                    "rules": [
+                        {
+                            "alert": "alert_on_error",
+                            "expr": 'rate({%%juju_topology%%}|="ERROR"[5m]) > 0',
+                            "for": "1m",
+                            "labels": {
+                                "severity": "critical",
+                            },
+                            "annotations": {"summary": "Logs found at ERROR level"},
```
sed-i, line 376: Drop annotations for brevity?

tests/unit/test_consumer.py
```diff
@@ -358,3 +359,89 @@ def test_rules_files_with_invalid_yaml_are_dropped_and_produce_an_error(self):
...
+                    "rules": [
+                        {
+                            "alert": "alert_on_error",
+                            "expr": 'rate({%%juju_topology%%, juju_unit="app/0"}|="ERROR"[5m]) > 0',
+                            "for": "1m",
+                            "labels": {
+                                "severity": "critical",
+                                "juju_unit": "app/0",
+                            },
+                            "annotations": {"summary": "Logs found at ERROR level"},
```
sed-i, line 416: Drop annotations for brevity?

tests/unit/test_consumer.py
```diff
@@ -358,3 +359,89 @@ def test_rules_files_with_invalid_yaml_are_dropped_and_produce_an_error(self):
...
+        pprint.pprint(rules)
+        expr = rules["groups"][0]["rules"][0]["expr"]
+        self.assertTrue(
+            "juju_model" in expr
+            and "juju_model_uuid" in expr
+            and "juju_application" in expr
+            and "juju_charm" in expr
+            and "juju_unit" in expr
+        )
+        self.assertEqual(
```
sed-i, line 437:
```suggestion
        self.assertGreaterEqual(
```

tests/unit/test_consumer.py
```diff
@@ -358,3 +359,89 @@ def test_rules_files_with_invalid_yaml_are_dropped_and_produce_an_error(self):
...
+                {
+                    "name": "alert_on_error",
+                    "rules": [
+                        {
+                            "alert": "alert_on_error",
+                            "expr": 'rate({%%juju_topology%%, juju_unit="app/0"}|="ERROR"[5m]) > 0',
+                            "for": "1m",
+                            "labels": {
+                                "severity": "critical",
+                                "juju_unit": "app/0",
```
sed-i, line 414: Do we need two more tests?
- `juju_unit` in expr but not in labels
- `juju_unit` not in expr but in labels

tests/unit/test_consumer.py
```diff
@@ -358,3 +359,89 @@ def test_rules_files_with_invalid_yaml_are_dropped_and_produce_an_error(self):
...
+                }
+            ]
+        }
+        self.sandbox.writetext("error.rules", yaml.dump(unlabeled_rule))
+        self.harness.begin_with_initial_hooks()
+        relation = self.harness.charm.model.get_relation("logging")
+        rules = json.loads(relation.data[self.harness.charm.app].get("alert_rules"))
+        import pprint
+
+        pprint.pprint(rules)
```
sed-i, line 428: Leftover?

tests/unit/test_consumer.py
```diff
@@ -358,3 +359,89 @@ def test_rules_files_with_invalid_yaml_are_dropped_and_produce_an_error(self):
...
+        expr = rules["groups"][0]["rules"][0]["expr"]
+        self.assertTrue(
```
sed-i, line 430: Multiple `assertIn`?

tests/unit/test_consumer.py
dstathis: Yes. Fixed.

tests/unit/test_consumer.py
dstathis: Agreed. Changed.

tests/unit/test_consumer.py
dstathis: I don't understand this suggestion. Could you elaborate?

tests/unit/test_consumer.py
dstathis: I don't understand this suggestion. Could you elaborate?

tests/unit/test_consumer.py
dstathis: I think having it look a bit more "real" is nice here

tests/unit/test_consumer.py
dstathis: I think if one of those tests were to fail, then one of the tests we already have would too. I know that depends on the two functionalities being independent but I think it is probably okay as is.

tests/unit/test_consumer.py
dstathis: Yes. Fixed.

tests/unit/test_consumer.py
dstathis: Agreed. Fixed.

tests/unit/test_consumer.py
sed-i: If the lib at some point in the future adds another label that is independent to this feature, then this test would break because it would no longer match `{"juju_application", "juju_charm", "juju_model", "juju_model_uuid", "severity"}` exactly. For that reason I am proposing to change the assertion to "subset" (⊆)

sed-i (approved):

rbarry82 (approved):

tests/unit/test_consumer.py
dstathis: I think it will be much better to update the test when it changes. These bugs make me think it is best to check for exact labels.

Issue #117, opened 2022-03-21
`_endpoints` methods must not rely on `planned_units`


Abuelodelanada:
### Bug Description

The [following method](https://github.com/canonical/loki-k8s-operator/blob/main/lib/charms/loki_k8s/v0/loki_push_api.py#L1263): 

```python
    def _endpoints(self) -> List[dict]:
        """Return a list of Loki Push Api endpoints."""
        return [{"url": self._url(unit_number=i)} for i in range(self._charm.app.planned_units())]
```

in `LokiPushApiProvider` class relies on planned units numbers to generate Loki endpoints.

But lifecycle reasons mean that at any given point unit numbers may not start with 0, may also have gaps, for instance if unit 5 is faulty I would remove it and add a new one, I'd end up with units 4, 6, 7.

### To Reproduce

There is no need to have steps to reproduce this.

### Environment

- juju 2.9.25

### Relevant log output

```shell
None
```

Abuelodelanada:
@mmanciop said:

> that method should be easy to change based on the remote unit name, but not if it is a CMR
> Because the unit name would not be correct
> We will have to have the requesters put their topology (or at least, unit name) in their own data bags

Abuelodelanada:
Fixed in #120

PR #120, opened 2022-03-24
Integrate Loki with Traefik ingress

### LINKED ISSUE #116: Integrate loki_push_api with Ingress
### Bug Description

Loki charmed in standalone mode must implement an ingress relation using the ingress_per_unit relation interface and the https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_k8s/v0/ingress_per_unit.py  library. The URL received by Loki must be used to produce the URL of the Loki Push API endpoint, and the https://github.com/canonical/loki-k8s-operator/blob/ba1bd4176a83b6849c89c46137f39f1b7694cadc/lib/charms/loki_k8s/v0/loki_push_api.py#L1111. No tight coupling between the ingress and loki_push_api API is desired. LokiPushApiProvider must therefore be updated to enable the Loki charm to specify the host address and the path of the LokiPushApi endpoint.

**Note**: The path of the LokiPushApi endpoint will be affected by an ingress using a path routing rule, see e.g. https://github.com/canonical/traefik-k8s-operator#configurations

**Note**: The future charming of Loki using the distributed mode, will instead use the [https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_k8s/v0/ingress.py](https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_k8s/v0/ingress_per_unit.py) library and the respective ingress relation interface, as the Loki ingester components are designed to be load-balanced, rather than exposed as separate endpoints.

### To Reproduce

None

### Environment

- juju
- microk8s

### Relevant log output

```shell
None
```

### Additional context

_No response_

Abuelodelanada:
## Issue

This PR fixes #116 and #117 (since they are related)

## Solution

This PR integrates `ingress_per_unit` lib with `Loki`

## Context

Loki send over relation data its endpoint (internal cluster IP) so client charms can set up it, but this is not useful for clients that are outside the cluster. That why we are integrating `ingress_per_unit` lib from Traefik charm

## Testing Instructions

1. Deploy Grafana agent charm
2. Deploy Loki charm
3. Deploy Traefik charm
4. Add a relation between Grafana agent and Loki
5. Check in `/etc/agen/agent.yaml` that Loki's endpoint is present.
6. Relate Loki with Traefik
7. Check in `/etc/agen/agent.yaml` that Traefik based endpoint is present.


config.yaml
```diff
@@ -0,0 +1,15 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+options:
+  web_external_url:
```
mmanciop, line 4: This config is neither needed nor desired. We are actually thinking of dropping it entirely from `prometheus-k8s`.

src/charm.py
```diff
@@ -237,6 +260,33 @@ def _loki_config(self) -> dict:
...
+        """Returns the pod ip of this unit."""
+        if bind_address := self.model.get_binding(PEER).network.bind_address:
+            return str(bind_address)
+        return ""
+
+    @property
+    def _external_url(self) -> str:
+        """Return the external hostname to be passed to ingress via the relation."""
+        self.framework.breakpoint()
+        if "web_external_url" in self.model.config:
```
mmanciop, line 274: This block must go :-)

src/charm.py
```diff
@@ -56,12 +59,24 @@ def __init__(self, *args):
         self.grafana_source_provider = GrafanaSourceProvider(
```
mmanciop, line 59: Here we need a similar change as in https://github.com/canonical/prometheus-k8s-operator/pull/253

mmanciop, line 59: Here we need a similar change as in https://github.com/canonical/prometheus-k8s-operator/pull/253

src/charm.py
```diff
@@ -56,12 +59,24 @@ def __init__(self, *args):
...
             charm=self,
             refresh_event=self.on.loki_pebble_ready,
-            source_type="loki",
+            source_type=PEER,
```
mmanciop, line 62: This seems a find/replace mistake. This is the datasource type. It just happens now to have the same name as the peer relation.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1167,6 +1167,10 @@ def __init__(
         *,
         port: int = 3100,
         rules_dir="/loki/rules",
+        endpoint_schema: str = "http",
+        endpoint_address: str = "",
+        endpoint_port: Union[str, int] = 3100,
```
mmanciop, line 1172: Isn't this already the value for `port`?

src/charm.py
```diff
@@ -56,12 +59,24 @@ def __init__(self, *args):
...
+            self, endpoint="ingress-per-unit", port=self._port
+        )
+        external_url = urlparse(self._external_url)
+
+        self.loki_provider = LokiPushApiProvider(
+            self,
+            endpoint_address=external_url.hostname or "",
+            endpoint_port=external_url.port or self._port,
+            endpoint_schema=external_url.scheme,
+            endpoint_path=f"{external_url.path}/loki/api/v1/push",
```
mmanciop, line 78: I believe Loki needs to be started with `-server.path-prefix=`, see https://stackoverflow.com/questions/62575304/how-to-configure-loki-to-accept-requests-with-subpath

mmanciop (changes requested):

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: Yes, it's the value for `_port` ... may be it is redundant.

src/charm.py
Abuelodelanada: What do you think using `http_path_prefix` and place this setting in the config file instead of starting the sever with the `--server.path-prefix` flag?

See: https://grafana.com/docs/loki/latest/configuration/#server_config

src/charm.py
Abuelodelanada: So do you mean rename `self.grafana_source_provider` with  `self.grafana_source_consumer`  and moving the `grafana-source` relation from the `provider` side to the `requires` in `metadata.yaml`?

rbarry82: Actually, it's the `source_uri` parameter that's needed, maybe...

When `web.external-url` is set in Prometheus, Prometheus routes _all_ connections to that path. So if ingress points to `http://1.2.3.4/prometheus-unit-0/`, `web.external-url` is also set to `prometheus-unit-0/`. Ingress is not performing any path rewriting to push it back to `/`. 

In this case, even "in-namespace" requests to  `http://{prometheus_address}:9090` from inside the model/namespace will fail unless the path set in `web.external-url` is appended.

Failing to do so will cause lookups to `v1/api/query/....` to fail, and Grafana doesn't "see" Prometheus data.

If Loki operates the same way with `server.path-prefix= || http_path_prefix` (and it looks like it does), then the full URL/URI ("with" `http://` preferred, but not strictly required) should also be passed to `GrafanaSourceProvider` as the argument to `source_uri` when `web_external_url` has a value.

rbarry82 (changes requested):
Generally ok, but the URI|URL should be passed to `GrafanaSourceProvider` also. Details inline.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1229,12 +1237,43 @@ def _on_logging_relation_changed(self, event: HookEvent):
...
                 self._process_logging_relation_changed(relation)
 
+    def _on_logging_relation_departed(self, event: RelationDepartedEvent):
+        """Removes alert rules files when consumer charms left the relation with Loki.
+
+        Args:
+            event: a `CharmEvent` in response to which the Loki
+                charm must update its relation data.
+        """
+        # self._set_endpoint_on_relation(event.relation)
```
sed-i, line 1255: This commented-out line got me thinking: 
1. do we need a `if event.departing_unit == self.model.unit` guard if the charm is scaled down to zero units? I.e. is app data already handled correctly in that case?
1. do we need a `_set_endpoint_on_relation` or `_process_logging_relation_changed` on leader-elected?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1229,12 +1237,43 @@ def _on_logging_relation_changed(self, event: HookEvent):
...
+        self._process_logging_relation_changed(event.relation)
+
+    def update_endpoint(self, relation: Relation = None) -> None:
```
sed-i, line 1258: To minimize user involvement, would it make sense to have the constructor take an optional arg for extra events to register this handler to?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1250,29 +1289,40 @@ def _process_logging_relation_changed(self, relation: Relation):
...
+        endpoint_url = "{}://{}:{}{}".format(
+            self._endpoint_schema, address, str(self._endpoint_port), path
+        )
+        relation.data[self._charm.unit]["endpoint"] = json.dumps({"url": endpoint_url})
+        logger.debug(
+            "Saved endpoint %s in relation data. ", relation.data[self._charm.unit]["endpoint"]
         )
 
+    def _get_relation_bind_address(self):
+        network_binding = self._charm.model.get_binding(self._relation_name)
```
sed-i, line 1323: Note that this may return None. Not sure if relevant for the circumstances here.
https://bugs.launchpad.net/juju/+bug/1929364

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1602,8 +1643,18 @@ def loki_endpoints(self) -> List[dict]:
...
         endpoints = []  # type: list
+
         for relation in self._charm.model.relations[self._relation_name]:
-            endpoints = endpoints + json.loads(relation.data[relation.app].get("endpoints", "[]"))
+            for unit in relation.units:
+                if unit.app is self._charm.app:
+                    # This is a peer unit
+                    continue
+
+                endpoint = relation.data[unit].get("endpoint")
```
sed-i, line 1653: :+1:

metadata.yaml
```diff
@@ -43,6 +43,13 @@ provides:
 requires:
   alertmanager:
     interface: alertmanager_dispatch
+  ingress-per-unit:
+    interface: ingress_per_unit
+    limit: 1
+
+peers:
+  loki:
+    interface: loki
```
sed-i, line 52: Consider using `peers`, `replicas`, etc.
In juju status,
```
Relation provider
loki:loki
```
seems a bit confusing to me.

requirements.txt
```diff
@@ -3,4 +3,5 @@ kubernetes
 requests
 lightkube
 lightkube-models
-aiohttp
\ No newline at end of file
+aiohttp
+serialized_data_interface >= 0.4.0
```
sed-i, line 7: Should we start commenting inside `requirements.txt`?
```suggestion

# SDI library
# Used by: traefik-k8s
serialized_data_interface >= 0.4.0
```

src/charm.py
```diff
@@ -92,6 +111,10 @@ def _on_alertmanager_change(self, event):
     def _loki_push_api_alert_rules_changed(self, event):
         self._configure(event)
 
+    def _on_ingress_changed(self, event):
+        self._configure(event)
+        self.loki_provider.update_endpoint()
```
sed-i, line 116: If `self.loki_provider.update_endpoint()` is called after `_configure`, then perhaps calling `_configure` is not needed at all?

src/charm.py
```diff
@@ -237,6 +260,33 @@ def _loki_config(self) -> dict:
...
+    @property
+    def _external_url(self) -> str:
+        """Return the external hostname to be passed to ingress via the relation."""
+        self.framework.breakpoint()
+        if "web_external_url" in self.model.config:
+            if web_external_url := self.model.config["web_external_url"]:
+                return web_external_url
+
+        if ingress_url := self.ingress_per_unit.url:
+            logger.info("This unit's ingress URL: %s", ingress_url)
```
sed-i, line 279: Should this be a DEBUG?

src/charm.py
sed-i: Iiuc, this condition is always True. Regardless, it is redundant since you have `if web_external_url := self.model.config["web_external_url"]`, which I would change to `if web_external_url := self.model.config.get("web_external_url")` just in case :D

tests/unit/test_provider.py
```diff
@@ -120,20 +119,19 @@ def test_relation_data(self):
...
-            mock_unit_ip.return_value = "10.1.2.3"
             rel_id = self.harness.add_relation("logging", "promtail")
             self.harness.add_relation_unit(rel_id, "promtail/0")
-
             self.harness.update_relation_data(rel_id, "promtail", {"alert_rules": "ww"})
             self.assertEqual(
                 sorted(logger.output)[0],
                 "DEBUG:charms.loki_k8s.v0.loki_push_api:Saved alerts rules to disk",
             )
+            self.assertEqual(
```
sed-i, line 131: Can we assert against the relation data directly?

src/charm.py
```diff
@@ -15,6 +15,7 @@
 import logging
```
sed-i, line 15: Is an itest in the scope of this PR?

.jujuignore
```diff
@@ -1,3 +1,4 @@
 /venv
 *.py[cod]
 *.charm
+.#*
```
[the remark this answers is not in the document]
Abuelodelanada: emacs backup files start with `.#` and if you didn't save your file and run `charmcraft pack`, the command fails because of this backup file.

requirements.txt
Abuelodelanada: Since `serialized_data_interface` [it's going to be removed](https://github.com/canonical/traefik-k8s-operator/pull/28) soon from Traefik lib it's not necessary.

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: 1.- As far as I understand it's not needed. App data is handled correctly while scaling down to zero and scaling up again to (let's say) 3 units.
2.- No, we do not need it since in the Provider side of the relation we need to set unit endpoint in every relation changed.

src/charm.py
mmanciop: @Abuelodelanada yes

config.yaml
mmanciop: @Abuelodelanada why did you close this discussion? The bogus config is still here.

mmanciop: @Abuelodelanada why did you close this discussion? The bogus config is still here.

mmanciop: @Abuelodelanada why did you close this discussion? The bogus config is still here.

lib/charms/loki_k8s/v0/loki_push_api.py
mmanciop: Then we need to change the naming scheme (and potentially break people) OR get rid of the `endpoint_` prefix and remove the duplicate.

mmanciop (changes requested):

src/charm.py
Abuelodelanada: @rbarry82 are we planning to rename [`source_url`](https://github.com/canonical/grafana-k8s-operator/blob/main/lib/charms/grafana_k8s/v0/grafana_source.py#L323) with `source_uri`?

src/charm.py
mmanciop: If it works, I have no particular preference one way or another. I am used to debug path issues with `ingress` using `kubectl exec ... -- ps aux`, but checking out the file works fine too.

src/charm.py
mmanciop: @Abuelodelanada we did ;-) https://github.com/canonical/grafana-k8s-operator/blob/ffa7b7f750a328bdf75cbbd3da300083bff5bf7f/lib/charms/grafana_k8s/v0/grafana_source.py#L30

src/charm.py
Abuelodelanada: When a relation between Loki and Traefik is established the `server` section in Loki's config file is:

```yaml
server:
  http_listen_address: 0.0.0.0
  http_listen_port: 3100
  http_path_prefix: /cos-loki-0
``` 

imagen


When there is not relation:

```yaml
server:
  http_listen_address: 0.0.0.0
  http_listen_port: 3100
  http_path_prefix: ''
```

imagen

src/charm.py
rbarry82: Way to sneak that one in 🤐 I was against that change, since naming it `source_uri` in the first place was a "hey, please include the schema when you pass this!" hint in the variable name. 

Sure the code checked if it included one, and added `http://` if it did not, but `_uri` was not an accident.

tests/unit/test_provider.py
rbarry82: Even if we can, ensuring that the actual code sees it is probably of value.

rbarry82 (approved):

src/charm.py
mmanciop: URLs must have the scheme too. Every URL is a URI. The difference is that URIs does not have to be resolvable, URLs must.

mmanciop: If the worry is about protocol-relative URLs, those are by definitions valid URIs too. It would make sense to me to validate the URL and ask it to be absolute or raise an exception.

tests/integration/test_rerelate.py
```diff
@@ -42,7 +42,7 @@ async def deploy(self, ops_test: OpsTest):
 
 related_apps = [
     RelatedApp("alertmanager", "ch:alertmanager-k8s", "alertmanager", {}),
-    RelatedApp("cassandra", "ch:cassandra-k8s", "logging", {"heap_size": "1g"}),
+    RelatedApp("grafana-agent", "ch:grafana-agent-k8s", "logging", {}),
 ]
```
sed-i, line 46: Add traefik to this list?

Abuelodelanada: I added a new specific itest for traefik

mmanciop (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -470,7 +470,7 @@ def _alert_rules_error(self, event):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 10
+LIBPATCH = 12
```
mmanciop, line 473: CharmHub won’t allow you to skip versions. Current version is 9

mmanciop (changes requested):

lib/charms/loki_k8s/v0/loki_push_api.py
mmanciop:
```suggested
LIBPATCH = 10
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1234,12 +1243,43 @@ def _on_logging_relation_changed(self, event: HookEvent):
...
                 self._process_logging_relation_changed(relation)
 
+    def _on_logging_relation_departed(self, event: RelationDepartedEvent):
+        """Removes alert rules files when consumer charms left the relation with Loki.
+
+        Args:
+            event: a `CharmEvent` in response to which the Loki
+                charm must update its relation data.
+        """
+        # self._set_endpoint_on_relation(event.relation)
```
mmanciop, line 1261: Get rid of this comment?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1255,27 +1295,32 @@ def _process_logging_relation_changed(self, relation: Relation):
         if self._charm.unit.is_leader():
             relation.data[self._charm.app].update(self._promtail_binary_url)
             logger.debug("Saved promtail binary url: %s", self._promtail_binary_url)
-            relation.data[self._charm.app]["endpoints"] = json.dumps(self._endpoints())
```
mmanciop, line 1258: This is a breaking change, and a breaking change that will NOT work with Loki charmed in distributed mode, where writing in unit data bags makes no sense.

mmanciop (changes requested):
Breaking lib changes, and not future proof for distributed Loki

lib/charms/loki_k8s/v0/loki_push_api.py
mmanciop: Use instead https://github.com/canonical/traefik-k8s-operator/blob/a98fcd300b6c009bc5b6814a55096387f59abde2/lib/charms/traefik_k8s/v0/ingress_per_unit.py#L403 from the Loki leader.

config.yaml
Abuelodelanada: Forgot to delete it. Done.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1183,8 +1186,11 @@ def __init__(
                 It is strongly advised not to change the default, so that people
                 deploying your charm will have a consistent experience with all
                 other charms that consume metrics endpoints.
-
+            port: the port on which Loki is running.
             rules_dir: path in workload container where rule files are to be stored.
+            scheme: whether Loki is running in http or https.
+            address: Loki server address.
```
mmanciop, line 1192: We are not documenting what happens with the default value as empty string, and we should because the behavior is non trivial.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1240,6 +1249,41 @@ def _on_logging_relation_changed(self, event: HookEvent):
...
+            return
+
+        if not relation:
+            if not self._charm.model.get_relation(self._relation_name):
+                return
+
+            relation = self._charm.model.get_relation(self._relation_name)
+
+        relation.data[self._charm.app].update({"endpoints": json.dumps(self._endpoints())})
+        logger.debug("Saved endpoints in relation data")
```
mmanciop, line 1285: I don’t think this debug log is worth having

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1266,7 +1309,21 @@ def _process_logging_relation_changed(self, relation: Relation):
 
     def _endpoints(self) -> List[dict]:
         """Return a list of Loki Push Api endpoints."""
-        return [{"url": self._url(unit_number=i)} for i in range(self._charm.app.planned_units())]
+        endpoints = []
+        path = "/loki/api/v1/push"
+
+        if self._charm.ingress_per_unit.urls:
+            return [{"url": url + path} for url in self._charm.ingress_per_unit.urls.values()]
```
mmanciop, line 1316: Concatenating with something more solid than string concatenation would make me feel safer, e.g., to be sure to avoid double slashes

metadata.yaml
```diff
@@ -43,6 +43,13 @@ provides:
 requires:
   alertmanager:
     interface: alertmanager_dispatch
+  ingress-per-unit:
```
mmanciop, line 46: Please change the relation name to `ingress`, it’s clearer to the Juju admin and will remain consistent with prometheus-k8s and likely with future distributed loki (even though that will use a different interface)

src/charm.py
```diff
@@ -45,23 +49,32 @@ class LokiOperatorCharm(CharmBase):
...
             charm=self,
             refresh_event=self.on.loki_pebble_ready,
-            source_type="loki",
-            source_port=str(self._port),
+            source_type=PEER,
+            source_url=self._external_url,
+        )
+
+        self.loki_provider = LokiPushApiProvider(
+            self,
```
mmanciop, line 72: I don’t understand why we don’t do self.hostname in the else branch

mmanciop (changes requested):
Have not tried the code yet,  but there’s a few comments

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: I think the default value for `address` should be `localhost`

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: `self._charm.ingress_per_unit.urls` always returns an url without trailing slash, and `path` starts with `/`. Anyway I can add a `rstrip` to the url.

src/charm.py
Abuelodelanada: "else branch"? What do you mean?

src/charm.py
```diff
@@ -45,23 +49,32 @@ class LokiOperatorCharm(CharmBase):
...
+            address=external_url.hostname or "",
```
mmanciop, line 73:
```suggestion
address=external_url.hostname or self.hostname
```

src/charm.py
mmanciop: See next comment

lib/charms/loki_k8s/v0/loki_push_api.py
mmanciop: Absolutely not `localhost`. You are going to send `address` over the relation to a different container. The process there will think of its own `localhost`!

Abuelodelanada: No.
We are using `address` only to build the url used to check alert rules in `_check_alert_rules()` method:

```python
url = "{}://{}:{}{}/loki/api/v1/rules".format(self.scheme, self.address, self.port, path)
```

The endpoints we are sending over relation data come from ingress or are built on the fly, note that we are not using `address` here:

```python
    def _endpoints(self) -> List[dict]:
        """Return a list of Loki Push Api endpoints."""
        endpoints = []
        path = "/loki/api/v1/push"

        if self._charm.ingress_per_unit.urls:
            return [
                {"url": url.rstrip("/") + path}
                for url in self._charm.ingress_per_unit.urls.values()
            ]

        peers = [*self._charm.meta.peers.keys()][0]
        units = self._charm.model.get_relation(peers).units
        units.add(self._charm.unit)

        for unit in units:
            unit_number = unit.name.split("/")[-1]
            endpoints.append({"url": self._url(unit_number)})

        return endpoints

    def _url(self, unit_number) -> str:
        """Get the url for a given unit."""
        return "http://{}-{}.{}-endpoints.{}.svc.cluster.local:{}/loki/api/v1/push".format(
            self._charm.app.name,
            unit_number,
            self._charm.app.name,
            self._charm.model.name,
            self.port,
        )
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1250,15 +1252,24 @@ def _on_logging_relation_changed(self, event: HookEvent):
...
 
-    def update_endpoint(self, relation: Relation = None) -> None:
+    def _on_logging_relation_broken(self, event: RelationBrokenEvent):
+        """Handle broken relation in related consumers.
+
+        Args:
+            event: a `CharmEvent` in response to which the Loki
+                charm must update its relation data.
+        """
+        self._process_logging_relation_changed(event.relation)
```
mmanciop, line 1270: @PietroPasotti see evidence for my claim that there is no reason to be so fine-grained in terms of _why_ ingress-provided URLs change.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1268,11 +1279,11 @@ def update_endpoint(self, relation: Relation = None) -> None:
         Ingress after the `logging` relation is established.
 
         Args:
+            url: An optional url value to update relation data.
             relation: An optional instance of `class:ops.model.Relation` to update.
-                If not provided, all instances of the `prometheus_remote_write`
-                relation are updated.
         """
-        if not self._charm.unit.is_leader():
+        if isinstance(relation, (RelationDepartedEvent, RelationBrokenEvent)):
```
mmanciop, line 1285: @Abuelodelanada how can a `Relation` object be ALSO an event? :D

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1298,45 +1314,40 @@ def _process_logging_relation_changed(self, relation: Relation):
...
+        return "http://{}-{}.{}-endpoints.{}.svc.cluster.local:{}".format(
             self._charm.app.name,
-            unit_number,
+            self._charm.unit.name.split("/")[-1],
             self._charm.app.name,
             self._charm.model.name,
             self.port,
         )
 
+    def _endpoint(self, url) -> dict:
```
mmanciop, line 1340: I wish this would be a property rather than a method. Passing the `url` as argument feel unnecessary.

mmanciop (changes requested):
I think the code needs more work, and we most definitely need an integration test or two in this PR.

src/charm.py
```diff
@@ -63,14 +66,24 @@ def __init__(self, *args):
         self.service_patch = KubernetesServicePatch(self, [(self.app.name, self._port)])
 
         self.alertmanager_consumer = AlertmanagerConsumer(self, relation_name="alertmanager")
+        self.ingress_per_unit = IngressPerUnitRequirer(
+            self, relation_name="ingress", port=self._port
+        )
+        external_url = urlparse(self._external_url)
```
rbarry82, line 72: This looks a little confusing, mostly because `_external_url` is defined as a property waaaaay at the bottom, and it's not obvious when reading the constructor.

Especially considering that the parts aren't used until a little below this, it may read better by moving it down to just above `LokiPushApiProvider` and renaming the variable to something which can be distinguished at first-pass (instead of `self._external_url <-> external_url`), like `parsed_external_url` or whatever name you like.

rbarry82 (approved):

rbarry82:
Integration test would be great. The one from Prometheus could probably be _mostly_ reused.

Abuelodelanada:
> Integration test would be great. The one from Prometheus could probably be _mostly_ reused.

Absolutely, I was reading just that one

src/charm.py
Abuelodelanada: Done

src/charm.py
```diff
@@ -98,6 +114,75 @@ def _on_loki_pebble_ready(self, _):
...
+            logger.debug("This unit's ingress URL: %s", ingress_url)
+            return ingress_url
+
+        # If we do not have an ingress, then use the pod ip as hostname.
+        # The reason to prefer this over the pod name (which is the actual
+        # hostname visible from the pod) or a K8s service, is that those
+        # are routable virtually exclusively inside the cluster (as they rely)
+        # on the cluster's DNS service, while the ip address is _sometimes_
+        # routable from the outside, e.g., when deploying on MicroK8s on Linux.
+        return f"http://{self.hostname}:{self._port}"
```
dstathis, line 181: The comment seems to say we us IP address but the value set is hostname. Which one is it supposed to be? I would recommend hostname.

tests/integration/test_ingress_with_traefik_k8s.py
```diff
@@ -0,0 +1,69 @@
...
+            config={
+                "routing_mode": "path",
+                "external_hostname": "foo.bar",
+            },
+        ),
+    )
+
+    apps = [loki_name, traefik_name]
+    await ops_test.model.wait_for_idle(apps=apps, status="active")
+    assert initial_workload_is_ready(ops_test, apps)
```
dstathis, line 45: Asserting `initial_workload_is_ready` seems to be redundant given the previous line.

src/charm.py
Abuelodelanada: You are right, hostname.

tests/integration/test_ingress_with_traefik_k8s.py
Abuelodelanada: good catch!

rbarry82 (dismissed):

tests/integration/test_ingress_with_traefik_k8s.py
```diff
@@ -0,0 +1,104 @@
...
+
+import pytest
+from helpers import is_loki_up, oci_image
+
+logger = logging.getLogger(__name__)
+
+
+class AddressNotFoundError(Exception):
+    def __init__(self, message):
+        super().__init__(message)
```
sed-i, line 17: Seems to be unused.

src/charm.py
```diff
@@ -98,6 +114,77 @@ def _on_loki_pebble_ready(self, _):
     def _on_alertmanager_change(self, _):
         self._configure()
 
+    def _on_ingress_changed(self, _):
+        self._configure()
+        self.loki_provider.update_endpoint(url=self._external_url)
+
+    def _on_logging_relation_changed(self, event):
+        # If there is a change in logging relation, let's update Loki endpoint
+        # We are listening to relation_change to handle the Loki scale down to 0 and scale up again
```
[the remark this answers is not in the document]
Abuelodelanada: Yes, we still need this, specially when scaling down to 0 and scaling up again.

tests/integration/test_ingress_with_traefik_k8s.py
```diff
@@ -0,0 +1,95 @@
...
+@pytest.mark.abort_on_fail
+async def test_ingress_traefik_k8s_upscaling_loki(ops_test, loki_charm):
+    loki_name = "loki"
+    traefik_name = "traefik-ingress"
+
+    await ops_test.model.applications[loki_name].scale(scale=3)
+    await ops_test.model.wait_for_idle(apps=[loki_name], status="active", wait_for_exact_units=3)
+
+    result = await _retrieve_proxied_endpoints(ops_test, traefik_name)
+    scale = 3
```
sed-i, line 65: Just out of kindness for our future selves, move `scale = 3` to the top?

```suggestion
    scale = 3
    await ops_test.model.applications[loki_name].scale(scale=scale)
    await ops_test.model.wait_for_idle(apps=[loki_name], status="active", wait_for_exact_units=scale)

    result = await _retrieve_proxied_endpoints(ops_test, traefik_name)
```

sed-i (approved):

sed-i (approved):

PR #123, opened 2022-03-30
Automatically release statically-compiled promtail binaries

mmanciop:
## Issue

Automatically build, validate and ship statically-compiled `promtail` binaries based on [upstream `grafana/loki` releases](https://github.com/grafana/loki/releases).

## Solution

1. Every 4 hours, the cron-like [`check-promtail-releases`](.github/workflows/check-promtail-releases.yaml) GitHub workflow compares the latest upstream release of Loki with specific `promtail-*` tags in this repository.
For example, the upstream `v2.4.2` tag pointed at by the upstream release, means that the `check-promtail-releases` workflow checks for a `promtail-v2.4.2` tag in this repository.
2. If an upstream release is found that has no matching `promtail-*` tag, the [`build-promtail-release.yaml`](.github/workflows/build-promtail-release.yaml) is triggered, which:
   1. Statically compile `promtail` for different architectures
   2. Try the binary by building [containers with different base images](promtail-build/static/test) and run then with a `docker run` command
   3. Create a shallow tag of the original Loki codebase (without history) and create the `promtail-*` tag from it.
   4. Create a GitHub release with the promtail binaries, pointing at the `promtail-*` tag

Given the fact that we do not yet have `arm64`-based GitHub runners at our disposal, the build process currently builds only for `amd64`.
However, all it should take to build `arm64`, when we have runners for it, is add `arm64` to the `matrix.arch` fields in the GitHub workflows.

## Context

Static linking is necessary due to the various versions of libc (glibc, muslc), or even the absence of any libc implementation, to be found in containers based on container images other than `ubuntu` and the like.
For example, Alpine-based containers ship muslc.
Distroless-based and "from scratch" containers often have no libc at all.
And the `promtail` builds from [upstream](https://github.com/grafana/loki/releases/) effectively work only on a subset of containers, which is a limitation we cannot afford.

## Testing Instructions

Run the build manually from https://github.com/canonical/loki-k8s-operator/actions, wait for the GitHub release to appear in https://github.com/canonical/loki-k8s-operator/releases, grab a `promtail` binary, and try it out :D

## Release Notes

This repository is going to regularly release statically-compiled `promtail` binaries tailing the upstream releases from [upstream `grafana/loki`](https://github.com/grafana/loki/releases/).

### DIFF
--- a/.github/workflows/build-promtail-release.yaml
+++ b/.github/workflows/build-promtail-release.yaml
@@ -1,0 +1,174 @@
+name: Build Promtail releases
+on:
+  # Manual trigger
+  workflow_call:
+    inputs:
+      release:
+        required: true
+        type: string
+
+permissions:
+  contents: write
+
+jobs:
+
+  build:
+    name: Build statically-linked Promtail
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: [amd64] # [amd64, arm64] # GitHub workflows do not support YAML anchors :-(
+    steps:
+      - name: Checkout grafana/loki, tag ${{inputs.release}}
+        uses: actions/checkout@v2
+        with:
+          repository: grafana/loki
+          ref: ${{inputs.release}}
+          fetch-depth: 1
+          path: ./loki-upstream
+      - name: Checkout ${{github.repository}}
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          path: ./loki-k8s-operator
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17.8
+      - name: Build Promtail
+        # We run the make-based build with CGO_ENABLED=0
+        # because we want statically-linked binaries and
+        # cross-build for the specified architecture
+        run: |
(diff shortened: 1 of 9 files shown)


simskij (approved):

Abuelodelanada:
About the testing instructions, is there a way to test these workflows before merging this PR since in https://github.com/canonical/loki-k8s-operator/actions are only the workflows in main?

imagen

mmanciop:
> About the testing instructions, is there a way to test these workflows before merging this PR since in https://github.com/canonical/loki-k8s-operator/actions are only the workflows in main?
> 
> imagen

Yes, but better in your private clone, because that build process _will_ modify this repo by adding a tag and a GitHub release :-)

Abuelodelanada (approved):
Works ok.

imagen

```sh
➜  /tmp ./promtail-amd64 --version
promtail, version 2.4.2 (branch: HEAD, revision: 525040a)
  build user:       runner@fv-az178-772
  build date:       2022-03-30T12:56:28Z
  go version:       go1.17.8
  platform:         linux/amd64
```

PR #129, opened 2022-04-01
Alertmanager Integration

rbarry82:
## Issue
Implement integration between Alertmanager and Loki, so any alerting rules _actually_ fire alerts.

## Solution
Primarily this was simply additions to `metadata.yaml` and small changes to the charm. The bulk of it was the creation of a workload-less tester charm which can generate messages in an arbitrary format through Juju actions.

## Testing Instructions
Deploy Loki and Alertmanager, as well as some charm which includes alerting rules (or `cos-config` with rules which match). See if alerts fire.

## Release Notes
Integrate Loki and Alertmanager, so that alerts fired by Loki based on LogQL alert rules are sent to Alertmanager for routing.


sed-i (dismissed):

tests/integration/conftest.py
```diff
@@ -10,3 +11,19 @@ async def loki_charm(ops_test: OpsTest):
...
     charm = await ops_test.build_charm(".")
     return charm
+
+
+@pytest.fixture(scope="module", autouse=True)
+def copy_loki_library_into_tester_charm(ops_test):
+    """Ensure that the tester charm uses the current Prometheus library."""
+    library_path = "lib/charms/loki_k8s/v0/loki_push_api.py"
+    install_path = "tests/integration/loki-tester/" + library_path
+    shutil.copyfile(library_path, install_path)
```
mmanciop, line 21: I ain’t no @sed-i, but wouldn’t it be possible to simply have the tester charm lib be a relative symlink to the root-level lib folder. Done it this way, it looks plenty brittle.

tests/integration/loki-tester/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
mmanciop, line 1: File not needed, top-level license applies.

tests/integration/conftest.py
balbirthomas: Have not tried it. But certainly worth a try. It would work if charmcraft resolves the symlink into an actual file and copies the file (rather than a symlink) into the built charm when packing. Of the top of my head I do not know if charmcraft actually does that.

tests/integration/helpers.py
```diff
@@ -40,6 +42,43 @@ async def loki_rules(ops_test, app_name) -> dict:
...
+            'severity': 'Low',
+            'status': 'testing'
+        }
+    }
+    """
+    address = await get_unit_address(ops_test, unit_name, unit_num)
+    url = f"http://{address}:9093/api/v2/alerts"
+    while not (alerts := json.loads(urllib.request.urlopen(url, data=None, timeout=2).read())):
+        retries -= 1
+        if retries > 0:
```
balbirthomas, line 74: ATTN @rbarry82 : I appreciate and like the simplicity of this approach. Thank you. I am going to do a couple of runs of this locally and if it works let us go with it. But if github runs start failing sporadically we could consider using exponential backoff here.

tests/integration/loki-tester/LICENSE
balbirthomas: Ack, my bad. Should have removed it after `charmcraft init`.

tests/integration/test_alert_rules_fire.py
```diff
@@ -0,0 +1,70 @@
...
+        apps=[loki_app_name, tester_app_name, alertmanager_app_name], status="active"
+    )
+
+    # Trigger a log message to fire an alert on
+    await ops_test.model.applications[tester_app_name].units[0].run_action(
+        "log-error", message="Error logged!"
+    )
+    alerts = await get_alertmanager_alerts(ops_test, "alertmanager", 0, retries=100)
+    assert all(
+        key in alert["labels"].kieys()
```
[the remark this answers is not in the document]
balbirthomas: Python should have detected `AttributeError` here without the need for mypy :-) .

tests/integration/test_alert_rules_fire.py
```diff
@@ -0,0 +1,70 @@
...
+
+    await ops_test.model.block_until(
+        lambda: (
+            len(ops_test.model.applications[loki_app_name].units) > 0
+            and len(ops_test.model.applications[tester_app_name].units) > 0
+            and len(ops_test.model.applications[alertmanager_app_name].units) > 0
+        )
+    )
+    await asyncio.gather(
+        ops_test.model.add_relation(loki_app_name, tester_app_name),
```
balbirthomas, line 54: We are combining two tests here

1. One to check that Loki charm does raise alerts when it receives logs that trigger one of its alert rules.
2. The other to check that the Loki charm successfully forwards alerts to alertmanager. 

I would have put test 2 into the cos-lite-bundle and leave test 1 in this repository.

balbirthomas (dismissed):

tests/integration/test_alert_rules_fire.py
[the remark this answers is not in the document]
balbirthomas: Locally tested and pushed fix. As predicted the error was
```
FAILED tests/integration/test_alert_rules_fire.py::test_alert_rules_do_raise_fire - AttributeError: 'dict' object has no attribute 'kieys'
```

balbirthomas (dismissed):

tests/integration/test_alert_rules_fire.py
balbirthomas: Brilliant. Thank you.

tests/integration/loki-tester/.flake8
```diff
@@ -0,0 +1,9 @@
+[flake8]
```
sed-i, line 1: Even for tester charms I would use operator-template rather than charmcraft init.

sed-i (approved):

Issue #128, opened 2022-03-31
Loki push API: Add `can_connect` guard in relation events.


Abuelodelanada:
### Bug Description

Sometimes `_on_logging_relation_changed` is fired before `pebble_ready`.

For instance two methods "inside" `_on_logging_relation_changed` we are executing:

```python
        files = container.list_files(self._rules_dir)
``` 
And if Pebble is not ready, and `ops.pebble.ConnectionError` is raise

### To Reproduce

See: https://github.com/canonical/grafana-agent-k8s-operator/runs/5771963079?check_suite_focus=true#step:4:530

### Environment

All versions of loki_push_api lib

### Relevant log output

```shell
Traceback (most recent call last):
  File "./src/charm.py", line 242, in <module>
    main(LokiOperatorCharm)
  File "/var/lib/juju/agents/unit-loki-0/charm/venv/ops/main.py", line 431, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-loki-0/charm/venv/ops/main.py", line 142, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-loki-0/charm/venv/ops/framework.py", line 283, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-loki-0/charm/venv/ops/framework.py", line 743, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-loki-0/charm/venv/ops/framework.py", line 790, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-loki-0/charm/lib/charms/loki_k8s/v0/loki_push_api.py", line 1237, in _on_logging_relation_changed
    self._process_logging_relation_changed(event.relation)
  File "/var/lib/juju/agents/unit-loki-0/charm/lib/charms/loki_k8s/v0/loki_push_api.py", line 1263, in _process_logging_relation_changed
    self._remove_alert_rules_files(self.container)
  File "/var/lib/juju/agents/unit-loki-0/charm/lib/charms/loki_k8s/v0/loki_push_api.py", line 1361, in _remove_alert_rules_files
    files = container.list_files(self._rules_dir)
  File "/var/lib/juju/agents/unit-loki-0/charm/venv/ops/model.py", line 1301, in list_files
    return self._pebble.list_files(path, pattern=pattern, itself=itself)
[...]

Issue #112, opened 2022-03-18
Every time the constructor is called, an HTTP request is sent to Loki asking it for it's version


simskij:
### Bug Description

Every time the constructor is called, an HTTP request is sent to Loki asking it for its version. This information is then logged into the debug logs, spamming them with unneeded log lines.

### To Reproduce

1. juju deploy loki-k8s --channel edge
2. look in the debug logs

### Environment

charm version: latest/edge

### Relevant log output

```shell
N/A
```

balbirthomas:
The `_provide_loki()` method  queries the Loki server version using a HTTP API call and raise different types of exceptions depending on the success or failure of this call. This method is only being used to set a status message by the Loki provider charm depending on if the Loki server is active and reachable or not. The docstring of the `LokiPushApiProvider` object also recommends the use of such a `_provide_loki()` method in this manner. However neither the docstring nor the Loki charm use the exceptions raised by the `_provide_loki()` to conditionally instantiate the `LokiPushApiProvider` object only if the Loki server is active and reachable. The `LokiPushApiProvier` object does use the Loki HTTP API in its `_check_alert_rules` method. This method is invoked in response to any changes in the relation with the consumer object and in response to a pebble ready method (see [PR 132](https://github.com/canonical/loki-k8s-operator/pull/132) ). Hence instantiating the `LokiPushApiProvider` object can lead to false negative result in `_check_alert_rules` just because the Loki HTTP API is not responsive. In view of these observations it is proposed that

1. The `_provide_loki()` method be renamed to something like `_is_loki_active()`. Instead of raising exceptions this method may return a boolean that asserts if the Loki server HTTP API is responsive or not.
2. Instantiation of the `LokiPushApiProvider` be guarded by check for `_is_loki_active()`.

There may be a few concerns here. For example how should such a scenario be handled

- The Loki charm receives a relation changed event but its workload container is not yet alive so this event is ignored (as per [PR 132](https://github.com/canonical/loki-k8s-operator/pull/132) ).
- Subsequently on pebble ready the Loki charm fails to see an active loki server because of the time lag between the workload container becoming active and the time it takes a Loki HTTP API to become responsive. hence the Loki charm does not instantiate the `LokiPushApiProvider` . Hence the pending relation changed event is still not handled. In fact it will never be handled.

[...]

PR #132, opened 2022-04-04
Fix pebble ready handling

balbirthomas:
## Issue
This pull request implements a fix for issue #128. This bug results in a untrapped pebble `ConnectionError` when a relation changed event tries to update Loki alert rules in the `LokiPushApi` provider workload container but pebble is not yet ready.

## Solution
This PR ensures that LokiPushApi consumers are updated and any new
alert rules set in the provider each time a pebble ready event is
triggered for the Provider. This is done because there is no guaranteed
ordering of pebble ready and relation changed events. Hence it can and
does happen that sometimes relation changed events are triggered for the
provider when it's workload container is not yet ready. To deal with
this situation a pebble ready event handler has been added that does
exactly the same thing as a relation changed event handler. Since the
workload container is only restarted when the layer configuration
changes but not when alert rule files change this change should not lead
to a infinite loop of container restarts. Also a guard has been added to
the relation changed handler to check if pebble is ready. If pebble is
not ready in the relation changed handler the Provider charm does
nothing, since a  subsequent pebble ready event is going to eventually
ensure that consumers and providers are in sync. This approach avoids
using the `event.defer()` method which can potentially lead to
unexpected behaviour.

## Context
The `LokiPushApiProvider` object updates alert rules its workload and ensures "promtail" is available for the consumer charm in response to a relation changed event with the consumer. Sometimes this process can fail if the relation changed event is triggered when the providers workload container is not ready.

## Testing Instructions
A unit test has been provided to check that alert rules are updated in response to pebble ready event on the Provider side. 

## Release Notes
LokiPushApiProvider does not require workload container to be ready when a relation changed event happens.

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1233,6 +1233,9 @@ class LokiPushApiProvider(RelationManagerBase):
             event: a `CharmEvent` in response to which the consumer
                 charm must update its relation data.
         """
+        if not self.container.can_connect():
+            return
+
         if isinstance(event, RelationEvent):
             self._process_logging_relation_changed(event.relation)
         else:
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1239,6 +1239,9 @@ class LokiPushApiProvider(RelationManagerBase):
                 charm must update its relation data.
         """
         if not self.container.can_connect():
+            # We do not defer the event because, if it is a relation created/changed, we will
+            # process the relation again on pebble_ready, and the nature of the event is not
+            # taken into account in the following logic.
             return
 
         if isinstance(event, RelationEvent):
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1223,6 +1223,11 @@ class LokiPushApiProvider(RelationManagerBase):
         self.framework.observe(events.relation_changed, self._on_logging_relation_changed)
         self.framework.observe(events.relation_departed, self._on_logging_relation_departed)
 
+        for container_name in self._charm.unit.containers:
+            self.framework.observe(
+                self._charm.on[container_name].pebble_ready, self._on_logging_relation_changed
+            )
+
     def _on_logging_relation_changed(self, event: HookEvent):
         """Handle changes in related consumers.
 
--- a/tests/integration/test_rerelate.py
+++ b/tests/integration/test_rerelate.py
@@ -87,7 +87,11 @@ async def test_rerelate(ops_test: OpsTest):
     await asyncio.gather(
         *[ops_test.model.add_relation(app_name, app.name) for app in related_apps],
     )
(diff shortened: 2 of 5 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1233,6 +1238,9 @@ def _on_logging_relation_changed(self, event: HookEvent):
             event: a `CharmEvent` in response to which the consumer
                 charm must update its relation data.
         """
+        if not self.container.can_connect():
+            return
```
mmanciop, line 1242: I think it is worth leaving here a comment as to why this event is not deferred.

mmanciop:
```suggestion
            # We do not defer the event because, if it is a relation created/changed, we will
            # process the relation again on pebble_ready, and the nature of the event is not
            # taken into account in the following logic.
            return
```

mmanciop (changes requested):
I think we need a little more inline comments in the code (see suggestion), the rest LGTM

mmanciop (approved):

Abuelodelanada (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1223,6 +1223,11 @@ def __init__(
         self.framework.observe(events.relation_changed, self._on_logging_relation_changed)
         self.framework.observe(events.relation_departed, self._on_logging_relation_departed)
 
+        for container_name in self._charm.unit.containers:
+            self.framework.observe(
+                self._charm.on[container_name].pebble_ready, self._on_logging_relation_changed
```
mmanciop, line 1228: I think we should track `relation_joined` instead, as that is fired up even on pod churn.

mmanciop (changes requested):

lib/charms/loki_k8s/v0/loki_push_api.py
balbirthomas: Your points are valid. Just this morning I raised [similar concerns here](https://github.com/canonical/loki-k8s-operator/issues/112). 

However I am surprised by the statement that "`relation_joined` is fired up even on pod churn" given the [official docstring](https://ops.readthedocs.io/en/latest/#ops.charm.RelationJoinedEvent).

mmanciop: > The event fires only when that remote unit is first observed by the unit.

When a pod churns, IIRC it counts as a new unit. Right @jameinel?

simskij: That might very well be the case, but it is not what was communicated when we created the Juju state machine diagram.

At that point, if I'm not completely mistaken, it was stated that a pod churn won't count as a new unit, as for all juju cares, it's the same (ie `charm/0` is still `charm/0`).

sed-i: Iirc, on upgrade we do not see any relation events.

balbirthomas:
Closing this PR because there is a new PR https://github.com/canonical/loki-k8s-operator/pull/152 created to fix issue #128 for which this PR was created but got blocked by failure of integration tests.

PR #114, opened 2022-03-18
Relation-departed shouldn't raise ModelError

sed-i:
This WIP is intended to resolve #102.

Crossref: OPENG-647

PR #134, opened 2022-04-06
Fix default value for json.loads

sed-i:
## Issue
#132 itest is [currently failing](https://github.com/canonical/loki-k8s-operator/runs/5843517601?check_suite_focus=true#step:4:527) for a JSONDecodeError becaus a fix in #114 didn't quite fix the problem.

## Solution
Instead of 
https://github.com/canonical/loki-k8s-operator/blob/c7e25d66ad8d5f158378910e3e54f33a71f1462b/lib/charms/loki_k8s/v0/loki_push_api.py#L2009
it should be `get("endpoints", "[]"))`, similar to elsewhere in the same file:
https://github.com/canonical/loki-k8s-operator/blob/c7e25d66ad8d5f158378910e3e54f33a71f1462b/lib/charms/loki_k8s/v0/loki_push_api.py#L1611

## Context
#114 was merged with a red itest after confirming the tests pass locally. Seems like they passed due to timing coincidence that avoided `raise_on_error`.

## Testing Instructions
`tox -e integration -- -k test_rerelate`

## Release Notes
Fix default value for json.loads in loki_push_api.

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -434,7 +434,7 @@ Loki Push API and alert rules.
 Units of consumer charm send their alert rules over app relation data using the `alert_rules`
 key.
 """
-
+import contextlib
 import json
 import logging
 import os
@@ -1221,7 +1221,13 @@ class LokiPushApiProvider(RelationManagerBase):
         events = self._charm.on[relation_name]
         self.framework.observe(self._charm.on.upgrade_charm, self._on_logging_relation_changed)
         self.framework.observe(events.relation_changed, self._on_logging_relation_changed)
+        self.framework.observe(events.relation_created, self._on_logging_relation_created)
         self.framework.observe(events.relation_departed, self._on_logging_relation_departed)
+
+    def _on_logging_relation_created(self, event: RelationCreatedEvent):
+        if self._charm.unit.is_leader():
+            event.relation.data[self._charm.app].update(self._promtail_binary_url)
+            logger.debug("Saved promtail binary url: %s", self._promtail_binary_url)
 
     def _on_logging_relation_changed(self, event: HookEvent):
         """Handle changes in related consumers.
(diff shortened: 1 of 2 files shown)


rbarry82 (changes requested):
This is a temporary fix, but the wrong one.

There is a workloadless tester charm which can drive Loki now. We should use _that_ and tear out the circular dependencies in [rerelate](https://github.com/canonical/loki-k8s-operator/blob/2b7cceae9f4d084123c88df56deb3b93acf0af46/tests/integration/test_rerelate.py). 

As a general pattern of the "testing pyramid" which has been discussed to death, integration tests for individual charms should not be dependent on other charms which use the same libraries for exactly this reason. A patch was merged with failing integration tests due to timing.

The fix to make tests pass _here_ should not be "copy the change from this library into another repo, modify the integration tests in this repo, and test". That workflow is upside-down. Changes in this repository should be able to be verified **in this repository**, and a failing test is an indication that we have broken the API. 

The first step towards correcting a broken API is to... fix the API. If it's discussed and absolutely necessary, then and only then does it merge.

In the meantime, "dependent" workflows (like "does `cassandra` work with this") can and should be done via something like:

```yaml
on: [push]
jobs:
  dispatch:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/github-script@v6
      with:
        github-token: ${{ secrets.PAT_TOKEN }}
        script: |
          await github.rest.actions.createWorkflowDispatch({
            owner: 'canonical',
            repo: 'cassandra-k8s-operator',
            workflow_id: 'test-libs.yml',
            ref: 'master'
          })
```

There's probably an action which uses the GH checks API to watch that. It would even be preferable as a middle ground to clone the entire repo for the other charms into this one and copy the libs (like the workloadless tester). CI for Loki depending on lib versions in a **published** charm is a terrible antipattern unless we agree that we will never break API, ever, under any circumstances, because "it's ok to merge when red this one time" is a slippery slope leading to exactly this PR.

simskij:
> As a general pattern of the "testing pyramid" which has been discussed to death, integration tests for individual charms should not be dependent on other charms which use the same libraries for exactly this reason.

This, very much this. We need to remove all dependencies to "real" charms from our charm-local integration tests asap.

That goes for Cassandra, Avalanche, and any other I've forgotten alike. Tester charms used as part of integration tests should live near the charm, preferably in the same repository.

sed-i:
Currently, `loki-tester` is only `loki_push_api` tester.
Is the intention to have one tester for all relations or one tester charm per relation?
@rbarry82 @simskij

sed-i:
Superseded by #139

PR #135, opened 2022-04-06
Create provider only when ready

balbirthomas:
## Issue

This PR addresses issue #112 and some of the additional concerns raised in that ticket.

At present the `LokiPushApiProvider` object is always instantiated in the Loki charm's constructor. However instantiating this provider object when the Loki server is unable to service HTTP requests may lead to errors, at least two of which can be anticipated
1. If `LokiPushApiProvider` gets a `RelationJoinedEvent` it will provide the consumer with its Push API endpoint URL. However this endpoint may not be operational and hence lead to errors in the consumer charm.
2. If the consumer charm forwards alert rules to `LokiPushApiProvider`, the provider charm will store these alert rules into Loki's alert rule folder and reload the rules. The provider charm will then try to check if the alert rules are valid by connecting to the `LokiPushApiProvider` HTTP API endpoint. Failure to make this connection will result in a false negative for the validity of these alert rules.

## Solution
In view of the above observations it is proposed that

- The _provide_loki() method be renamed to something like _is_loki_active(). Instead of raising exceptions this method may return a boolean that asserts if the Loki server HTTP API is responsive or not.
- Instantiation of the LokiPushApiProvider be guarded by check for _is_loki_active().

## Context
Issue #112 provides a details on context.

## Testing Instructions
This PR is work in progress and testing has not yet been considered

## Release Notes
LokiPushApiProvider now guarantees that when its sets a endpoint URL for the consumer, this endpoint will be reachable and active.

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -452,6 +452,7 @@ import yaml
 from ops.charm import (
     CharmBase,
     HookEvent,
+    RelationBrokenEvent,
     RelationCreatedEvent,
     RelationDepartedEvent,
     RelationEvent,
@@ -470,7 +471,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 10
+LIBPATCH = 11
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1553,9 +1565,14 @@ def __init__(
...
         super().__init__(charm, relation_name, alert_rules_path, recursive)
         events = self._charm.on[relation_name]
+
         self.framework.observe(self._charm.on.upgrade_charm, self._on_logging_relation_changed)
         self.framework.observe(events.relation_changed, self._on_logging_relation_changed)
-        self.framework.observe(events.relation_departed, self._on_logging_relation_departed)
+
+        # This will keep track of broken relations so that we can ignore the endpoints
+        # we find in their application data bags
+        self._ignored_relations = []  # type: list
```
sed-i, line 1574: Afaik, relation-broken is only seen by the departing unit - the remaining unit does not see this event.
Is this correct?

balbirthomas:
Closing this PR since there is now a new PR https://github.com/canonical/loki-k8s-operator/pull/151 to address issue #112 for which this PR was created but got blocked by failure of integration tests.

Issue #102, opened 2022-02-24
Loki push api: ModelError on relation-departed


PietroPasotti:
This happens from time to time on `juju remove-application loki-k8s`; might be a timing issue, if the relation-departed is queued after the databag gets deleted?

```
in _on_relation_departed                                                                                  
    new_config = self._promtail_config                                                                    
  File "/var/lib/juju/agents/unit-loki-tester-0/charm/lib/charms/loki_k8s/v0/loki_push_api.py", line 1890,
in _promtail_config                                                                                       
    config = {"clients": self._clients_list()}                                                            
  File "/var/lib/juju/agents/unit-loki-tester-0/charm/lib/charms/loki_k8s/v0/loki_push_api.py", line 1904,
in _clients_list                                                                                          
    clients = clients + json.loads(relation.data[relation.app]["endpoints"])                              
  File "/var/lib/juju/agents/unit-loki-tester-0/charm/venv/ops/model.py", line 430, in __getitem__        
    return self._data[key]                                                                                
  File "/var/lib/juju/agents/unit-loki-tester-0/charm/venv/ops/model.py", line 414, in _data              
    data = self._lazy_data = self._load()                                                                 
  File "/var/lib/juju/agents/unit-loki-tester-0/charm/venv/ops/model.py", line 779, in _load              
    return self._backend.relation_get(self.relation.id, self._entity.name, self._is_app)                  
  File "/var/lib/juju/agents/unit-loki-tester-0/charm/venv/ops/model.py", line 1588, in relation_get      
    return self._run(*args, return_output=True, use_json=True)                                            
  File "/var/lib/juju/agents/unit-loki-tester-0/charm/venv/ops/model.py", line 1523, in _run              
    raise ModelError(e.stderr)                                                                            
ops.model.ModelError: b'ERROR permission denied\n'                                                        
```

sed-i:
Something similar happens on `juju relate`/`relation-changed`:

```
cass/0*  error     idle   10.1.73.149         hook failed: "log-proxy-relation-changed" for loki:logging
```
```
unit-cass-0: 07:21:03.144 ERROR unit.cass/0.juju-log log-proxy:2: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 308, in <module>
    main(CassandraOperatorCharm, use_juju_for_storage=True)
  File "/var/lib/juju/agents/unit-cass-0/charm/venv/ops/main.py", line 426, in main
[...]

Issue #103, opened 2022-02-25
Alert rules integration tests improvements


Abuelodelanada:
- Consider renaming file e.g. `test_multiple_rule_providing_apps`: - _Originally posted by @sed-i in https://github.com/canonical/loki-k8s-operator/pull/99#discussion_r814292038_
- `async def deploy_rule_providing_app(app_name: str)`? - _Originally posted by @sed-i in https://github.com/canonical/loki-k8s-operator/pull/99#discussion_r814294535_
- If instead you remove `rules_app2` then you could easily have a stronger assertion: `assert rules_after_delete_relation2 == rules_after_relation` - _Originally posted by @sed-i in https://github.com/canonical/loki-k8s-operator/pull/99#discussion_r814299141_
- Add upgrade_charm test, see: https://github.com/canonical/loki-k8s-operator/pull/97

PR #139, opened 2022-04-08
Stabilize integration tests, no more catch-all methods

rbarry82:
Closes https://github.com/canonical/loki-k8s-operator/issues/102
Closes https://github.com/canonical/loki-k8s-operator/issues/128
Closes https://github.com/canonical/loki-k8s-operator/pull/110

## Issue
Integration tests are repeatedly and consistently failing, with no clear way to get through them as there is a mix of "check out files from the repo you are testing in an unmodifiable loop" and usage of another charm which carries a library from this charm as part of its tests.

Specifically:
* The usage of `cos-config` to check out alert rules in testing from `main` on this repository (or `error_alerts` -- previously deleted and now restored) combined with a disconnect in the testing for invalid alerts in #105 and the integration test for it in #103 means that even after fixing libraries, integration tests will immediately fail on `main` unless this is corrected even if this PR (or another one) passes tests. In order to break this loop, the alert rules are now copied into the tester in a manner similar to the way libraries are copied
* The usage of `cassandra-k8s` has been dropped. While we should not break API, and the tests there are valuable, it also means that fixing any bug in the library itself (such as #134) necessarily involves one side of the other with failing integration tests until it is updated and published.

## Solution
`loki-tester` has been updated to carry alert rules. It is also smarter about setting the logger, since Python's `logging` object carries global references, and `logger.addHandler()` adds to those after `logging.getLogger(...)`.   `removeHandler(...)` must be used to clear them, not simply setting to `None `.

A drive-by fix to endpoints and a guard around pushing files to the container to avoid extraneous log messages has been added.

`loki-tester` handles the logger better in general, and will not call a method which inadvertently tries to get/set app data as part of the constructor.

As part of this patch, remove the "common exit hook" pattern for relation events in `loki_push_api`, stop creating a dedented string which is immediately converted to a dict then back to YAML, and do as much cleanup as necessary to get the tests to run (proper return from `endpoint_list`, for example)

## Context
There is a very likely Juju bug causing workload-only charms to hang after scaling to 0, and never be fully removed, despite all k8s resources being gone. This _may_ be on the Juju roadmap for the next release, but will confirm/deny the circumstances early next week, or file a bug. Adding a useless container+resource to `loki-tester` in the same manner as other `-tester` charms appears to resolve this. Since the other of the existing "bug" is around Pebble, this is likely to be new. I have suspicions about the root cause, but we'll wait to find out.

In addition, if we are going to use `cos-config` this way in tests, we shoudl discuss the creation of a lookaside repo which can hold both examples of "working" alert rules, scrape configs, recording rules, dashboards, etc, as well as be used by integration testing _without_ requiring a PR which creates a secondary branch since it is not possible for the tests to pass with the file structure on `main`.

## Testing Instructions
Check out this PR.  `tox -e integration`

## Release Notes
Stabilize integration tests and drop external dependencies.


sed-i:
I am getting 
`loki-tester/0*   error     idle   10.1.157.93          hook failed: "install"` when I dropped that file into my PR.

rbarry82:
> I am getting `loki-tester/0* error idle 10.1.157.93 hook failed: "install"` `unit.loki-tester/0.install /var/lib/juju/agents/unit-loki-tester-0/charm/dispatch: 3: exec: ./src/charm.py: Permission denied` when I dropped that file into my PR.

Do you have an actual trace for this? `dispatch` invoking `./src/charm.py` is base OF, and completely unrelated to this change in any way. Built as the wrong user? Permissions wrong in the pod?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1251,18 +1251,25 @@ def _process_logging_relation_changed(self, relation: Relation):
 
         Args:
             relation: the `Relation` instance to update.
+            event: an optional event, the type of which is checked. If it is a
+                RelationDepartedEvent no updates are sent to the relation data bag
+
```
sed-i, line 1256:
```suggestion
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1251,18 +1251,25 @@ def _process_logging_relation_changed(self, relation: Relation):
...
         """
         if self._charm.unit.is_leader():
+            # We don't need to set a bunch of data if a relation is going away
             relation.data[self._charm.app].update(self._promtail_binary_url)
             logger.debug("Saved promtail binary url: %s", self._promtail_binary_url)
```
sed-i, line 1261: `_promtail_binary_url` should probably go into relation created (see  #134).

tests/integration/conftest.py
```diff
@@ -1,16 +1,52 @@
...
+
+
+store = Store()
+
+
+def timed_memoizer(func):
+    @functools.wraps(func)
+    async def wrapper(*args, **kwargs):
+        fname = func.__qualname__
+        logger.info("Started: %s" % fname)
```
sed-i, line 38:
```suggestion
        logger.info("Started: %s", fname)
```

tests/integration/conftest.py
```diff
@@ -1,16 +1,52 @@
...
+
+    def __setattr__(self, key, value):
+        """Override __setattr__ so dot syntax works on keys."""
+        self[key] = value
+
+
+store = Store()
+
+
+def timed_memoizer(func):
```
sed-i, line 34: This memoizer is not timed though, is it?

tests/integration/conftest.py
```diff
@@ -21,7 +57,24 @@ def copy_loki_library_into_tester_charm(ops_test):
     shutil.copyfile(library_path, install_path)
 
 
+@pytest.fixture(scope="module", autouse=True)
+def copy_sample_alerts_into_tester_charm(ops_test):
```
sed-i, line 61: How about a helper function to update the contents of a zip file as an alternative to this?

tests/integration/loki-tester/src/charm.py
```diff
@@ -23,6 +23,8 @@ def __init__(self, *args):
 
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.update_status, self._on_update_status)
+        self.framework.observe(self.on.stop, self._on_stop)
+        self.framework.observe(self.on.log_error_action, self._on_log_error_action)
```
sed-i, line 27: Is this line duplicated on purpose?

tests/integration/loki-tester/src/charm.py
```diff
@@ -35,30 +37,72 @@ def __init__(self, *args):
...
 
-        self.log("debug", "Constructed loki tester")
+    def _setup_logging(self, handler: dict = None) -> None:
+        """Ensure logging is configured correctly."""
+        handler = handler or {}
+        logger = logging.getLogger("Loki-Tester")
+        logger.setLevel(logging.INFO)
+
+        handlers = {"console": logging.StreamHandler()}
+        handlers.update(handler)
```
sed-i, line 49:
```suggestion
        handlers = {"console": logging.StreamHandler(), **handler}
```

tests/integration/test_alert_rules_fire.py
```diff
@@ -70,6 +75,8 @@ async def test_alert_rules_do_forward_to_alertmanager(ops_test, loki_charm, loki
     alertmanager_app_name = "alertmanager"
     app_names = [loki_app_name, tester_app_name, alertmanager_app_name]
 
+    await ops_test.model.set_config({"logging-config": "<root>=WARNING; unit=DEBUG"})
```
sed-i, line 78: Shouldn't this be in the first test function in the file?

sed-i (commented):
At the time of writing this, itests been running for >2 hrs :|

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: Completely unrelated to this patch.

tests/integration/conftest.py
rbarry82: [Yes, it is](https://github.com/canonical/loki-k8s-operator/pull/139/files#diff-c87c5173e88bb0f60b00107f4586304f4869a976a4f998425efe3481c0db9f56R46). It reports how long it took to either build (sometimes long) or return the memoized version (very fast). It's not `memoizer_with_timeout`. It's literally timing execution and logging it.

tests/integration/conftest.py
rbarry82: I very much do not think that we should build test infrastructure around assumptions about the internal structure of charms (or any other archive we don't control).

This pattern is already used for libs in multiple charms, and if checksumming is ever added anywhere, it will work with that, too, because it's pre-build. What is the advantage of trying to update a zip?

rbarry82:
> At the time of writing this, itests been running for >2 hrs :|

As a comment, "please check out the PR and run `tox -e integration`" is the instruction for testing.

We know that the GH workers can be slow and unreliable (hence `timeout=1000`). Due to the Juju bug that I'm collecting trace info for, it does _occassionally_ fail locally. Checking whether these tests resolve the bugs in this charm and not whether or not there is a bug in Juju due to be resolved is the goal, after all, but checking it out locally can demonstrate that.

mmanciop (commented):
There is a distressing amount of changes here that seem not related with "breaking circular dependencies in integration tests". It makes this PR IMO way too hard to review.

tests/integration/test_multiple_rule_providing_apps.py
```diff
@@ -9,64 +9,52 @@
...
 
 logger = logging.getLogger(__name__)
 
 METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
 app_name = METADATA["name"]
 resources = {"loki-image": METADATA["resources"]["loki-image"]["upstream-source"]}
-rules_app = "cos-config"
-rules_app2 = "cos-config2"
-rules_app3 = "cos-config3"
+tester_resources = {
```
balbirthomas, line 19: TODO :-)

balbirthomas (dismissed):

tests/integration/conftest.py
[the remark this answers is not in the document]
rbarry82: It's not required to address this specifically, it's just cutting down the turnaround time for running the test suite by only building the charms once rather than every time. This is, incidentally, part of the on-hold "testing harness", but is useful for testing here rather than adding an additional 1-4m per test file.

tests/integration/conftest.py
```diff
@@ -1,16 +1,52 @@
...
+    @functools.wraps(func)
+    async def wrapper(*args, **kwargs):
+        fname = func.__qualname__
+        logger.info("Started: %s" % fname)
+        start_time = datetime.now()
+        if fname in store.keys():
+            ret = store[fname]
+        else:
+            logger.info("Return for {} not cached".format(fname))
+            ret = await func(*args, **kwargs)
```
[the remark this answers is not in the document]
rbarry82: Fortunately, it does not.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1234,35 +1235,57 @@ def _on_logging_relation_changed(self, event: HookEvent):
...
-            relation.data[self._charm.app]["endpoints"] = json.dumps(self._endpoints())
             logger.debug("Saved endpoints in relation data")
 
-        if relation.data.get(relation.app).get("alert_rules"):
+        if relation.data.get(relation.app).get("alert_rules", ""):
             logger.debug("Saved alerts rules to disk")
-            self._remove_alert_rules_files(self.container)
-            self._generate_alert_rules_files(self.container)
-            self._check_alert_rules()
+            if self.container.can_connect():
```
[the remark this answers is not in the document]
rbarry82: This actually seems to be reasonably reliable (or at least, I am not seeing messages churn in the logs anymore), but I agree that a guard per method would be preferable.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1234,35 +1235,57 @@ def _on_logging_relation_changed(self, event: HookEvent):
...
+
+            current_data.update(updated_data)
+            relation.data[self._charm.app].update(updated_data)
+
             logger.debug("Saved promtail binary url: %s", self._promtail_binary_url)
-            relation.data[self._charm.app]["endpoints"] = json.dumps(self._endpoints())
             logger.debug("Saved endpoints in relation data")
 
-        if relation.data.get(relation.app).get("alert_rules"):
+        if relation.data.get(relation.app).get("alert_rules", ""):
```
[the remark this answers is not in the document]
rbarry82: Since without this, every `RelationChanged` event will mindlessly set `promtail_binary_url` regardless of whether or not it is already present in the data bag, plus setting endpoints. The changes here are primarily an attempt to "quiet" the relation between `LokiPushApiConsumer` and `LokiPushApiProvider` charms to see if the Juju bug can be isolated

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1234,35 +1235,57 @@ def _on_logging_relation_changed(self, event: HookEvent):
...
+                RelationDepartedEvent no updates are sent to the relation data bag
+
         """
-        if self._charm.unit.is_leader():
-            relation.data[self._charm.app].update(self._promtail_binary_url)
+        if self._charm.unit.is_leader() and event not in [
+            RelationBrokenEvent,
+            RelationDepartedEvent,
+        ]:
+            # Condense the updates so we aren't setting keys to the same values
```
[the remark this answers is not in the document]
rbarry82: Since without this, every `RelationChanged` event will mindlessly set `promtail_binary_url` regardless of whether or not it is already present in the data bag, plus setting endpoints. The changes here are primarily an attempt to "quiet" the relation between `LokiPushApiConsumer` and `LokiPushApiProvider` charms to see if the Juju bug can be isolated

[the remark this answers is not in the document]
balbirthomas: I know. But what I was saying for if the relation data contained `foo = bar` and a second invocation of this method again set `foo = bar` into relation data, doesn't Juju guarantee that no `RelationChanged` event will be triggered on the other side of the relation ? My understanding was Juju does make this guarantee. If you have reason to believe that it does not then perhaps we should raise a ticket.

lib/charms/loki_k8s/v0/loki_push_api.py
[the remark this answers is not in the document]
balbirthomas: I am not fussed about this. If you prefer you can add this refactor as another TODO in issue https://github.com/canonical/loki-k8s-operator/issues/142

tests/integration/conftest.py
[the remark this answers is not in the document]
balbirthomas: Thanks for clarifying. I am still inclined to request this be moved into a separate PR if for no reason other than to clarify through the means of this PR precisely those issues that have adversarially impacted the reliability of integration tests in this Loki charm. But if you feel strongly about this I am happy to leave it as is.

balbirthomas (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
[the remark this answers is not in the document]
rbarry82: Yes, it does. On the hand hand, constantly setting the same data and forcing Juju to respond (plus the invocation time for setting itself [since OF doesn't check the equality either](https://github.com/canonical/operator/blob/6013988d00d68f15ac61a3419bf2a4b2c7f49268/ops/model.py#L1668)) is something we could live without.

But it's the second part of this clause which the changes was made for. If `_endpoints()` **has** changed, we will send an update which is useless to the charm on the other side, and which may have caused some reference in Juju which it was "holding on to". 

It's probably a reasonable change anyway, but it was partly exploratory for a workaround to the Juju issue.

balbirthomas (approved):

balbirthomas (dismissed):

tests/integration/helpers.py
```diff
@@ -31,6 +32,7 @@ async def is_loki_up(ops_test, app_name, unit_num=0) -> bool:
     logger.debug("Loki address: %s", address)
 
     url = f"http://{address}:3100/loki/api/v1/status/buildinfo"
+    logger.info("Connecting to URL %s : ", url)
```
mmanciop, line 35: We should leave `#FIXME` to get rid of them later

rbarry82:
Closes #102 
Closes #128
Closes #110

rbarry82:
Closes #102 
Closes #128
Closes #110

tests/integration/test_rerelate.py
```diff
@@ -40,82 +45,111 @@ async def deploy(self, ops_test: OpsTest):
...
+            channel="edge",
+        ),
     )
+
     await asyncio.gather(
-        *[ops_test.model.add_relation(app_name, app.name) for app in related_apps],
+        ops_test.model.add_relation(app_name, "loki-tester-new"),
+        ops_test.model.add_relation(app_name, "alertmanager-new"),
     )
+    await ops_test.model.wait_for_idle(status="active", timeout=600)
```
sed-i, line 153: Without the `apps` arg, `wait_for_idle` will complain (and eventually timeout) that `Waiting for model: ... (not enough units yet - 0/1)`, even though the unit that is `0/1` is the unit that was removed in the previous test function.

```suggestion
    await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=600)
```

This change would hide the fact that some other app wasn't entirely removed.

Haven't checked yet what happens without the `-new`.

rbarry82: I don't understand what you're suggesting here.

Why would we _want_ to add an argument which could possibly mask undesirable behavior?

sed-i: There is a double overlook here, which unmasked the error:
1. Calling wait_for_idle without appropriate args would hang as long as there is any app with zero units - technically, even if to speed things up, we typically specify the apps of interest.
2. Calling app.remove doesn't guarantee removal. That prompts for another assertion with block_until after remove, i.e. the fact that a previous test func left dirty state is not a concern for the next test function (this one).

sed-i: - There's only one thing to fail on here: when after a `remove`, our `block_until "app" not in applications" times out. https://github.com/juju/python-libjuju/issues/656
- But since we did not have that `block_until`, the next test function starts with leftovers.
- And because we didn't specify `apps=[...]` in `wait_for_idle`, you'd keep seeing `not enough units yet - 0/1` for the leftover unit -- it is a feature of `wait_for_idle` to keep waiting if the number of units is still zero:
  - https://github.com/juju/python-libjuju/issues/574
  - https://github.com/juju/python-libjuju/issues/608

rbarry82: I really don't understand what the context is here anymore.

From my POV, there is only one thing to fail on:

* The state of the model does not match what our tests say it should. That's it.

This is basic test idempotency. If the code says "Install `foo`, `bar`, and `baz`, then remove `foo` and `baz` and wait for `bar` to be ready", I don't want to know _only_ if `bar` is ready, I want to know that the state of the environment matches what I expect, which _also_ includes the things I told to go away actually going away.

Seeing `not enough units yet - 0/1` for a "leftover" unit which we told to go away and did not and having the test fail is **exactly** the point of integration testing from my perspective. We talk a lot about testing the "happy path" and not having tests for unknown bugs. Well, we do, accidentally, and we've been working at it for most of the last week. This is not a bad or undesirable thing.

sed-i: Sorry, my `apps=[app_name]` must have caused the confusion: this was just an example - of course you list there all the apps you expect to see: `"loki-tester-new"` and `"alertmanager-new"`.
However, the `not enough units yet - 0/1` is for `"loki-tester"` or `"alertmanager"`. What I'm saying is that "The state of the model does not match what our tests say it should" for `"loki-tester"` or `"alertmanager"` should be covered by the previous test - `test_remove_related_app`, not this one.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -2008,7 +2056,7 @@ def _clients_list(self) -> list:
         """
         clients = []  # type: list
         for relation in self._charm.model.relations.get(self._relation_name, []):
-            endpoints = json.loads(relation.data[relation.app].get("endpoints", ""))
+            endpoints = json.loads(relation.data[relation.app].get("endpoints", []))
```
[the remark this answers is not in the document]
rbarry82: I can only blame git, because they're present in my local copy and `git status` is clean.

PR #144, opened 2022-04-11
Move endpoints to unit data to correctly handle rescale

### LINKED ISSUE #137: loki_push_api library misbehaves on relation_broken events
### Bug Description

The `loki_push_api` relation providers store data in the application data bag. That data survives all units being deleted (i.e., `juju scape-application loki 0`), and the `loki_push_api` library will provide wrong data with the `endpoints` API over a `relation_broken` hook.

### To Reproduce

Write a unit test that checks what happens when you access the `LokiPushApiConsumer.endpoints` property over a `relation_broken` hook :-)

### Environment

EVERYWHERE

### Relevant log output

```shell
(˚Õ˚)ر ~~~~╚╩╩╝
```

### Additional context

_No response_

### LINKED ISSUE #145: LogPushApi library broken with respect to scale up and scale down of provider units
### Bug Description

The `LokiPushApiProvider` object fails to update the URLs exposed in its application data bag when follower provider units join or depart. This problem is sometimes obscured by the fact that, on charm upgrade or on pebble_ready by the leader unit, relation data is synchronized again.

### To Reproduce

1. juju deploy loki-k8s --channel=edge --trust
3. juju deploy grafana-agent-k8s --channel=edge --trust
4. juju relate loki-k8s:logging grafana-agent-k8s
5. wait until loki-k8s's leader unit is idle
6. juju scale-application loki-k8s 3
7. wait until all units are idle
8. juju show-unit grafana-agent-k8s/0

### Environment

Every. Where.

### Relevant log output

```shell
U+1F4A9
```

### Additional context

_No response_

sed-i:
## Issue
Closes #145.
Closes #137.

## Solution
- Add tests.
- Move endpoints from app to unit data.

## Testing Instructions
- `tox -e integration -- --keep-models -k test_rescale_charm`
- `tox -e unit`

## Release Notes
- Move endpoints from app to unit relation data
- Add integration tests for scaling the application.

Crossref: [OPENG-762](https://warthogs.atlassian.net/browse/OPENG-762)


rbarry82:
Ok in general, but it feels like this test should actually check for some kind of data/result from Loki or its relations

sed-i:
> Ok in general, but it feels like this test should actually check for some kind of data/result from Loki or its relations

Thanks! Yes of course, so far it's just the scaffolding.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1365,8 +1371,9 @@ def unit_ip(self) -> str:
         if bind_address:
             logger.warning("Returning IP %s", str(bind_address))
             return str(bind_address)
-        logger.warning("No address found")
-        return ""
+        else:
+            logger.warning("No address found")
```
mmanciop, line 1375: IMO this log message is not informative enough to the end use, and should definitely not be a warning if we recover from it later. And if we do not recover, it's ERROR material ;-)

metadata.yaml
```diff
@@ -44,6 +44,10 @@ requires:
   alertmanager:
     interface: alertmanager_dispatch
 
+peers:
+  replicas:
```
mmanciop, line 48: I don't think we use this anywhere. And definitely we said we should not add a peer relation to handle this use-case.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1273,7 +1280,6 @@ def _process_logging_relation_changed(self, relation: Relation):
 
     def _update_relation_data(self, relation):
         if self._charm.unit.is_leader():
-            relation.data[self._charm.app].update(self._promtail_binary_url)
             logger.debug("Saved promtail binary url: %s", self._promtail_binary_url)
```
mmanciop, line 1283: This log message needs to go together with its line above

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1273,7 +1280,6 @@ def _process_logging_relation_changed(self, relation: Relation):
...
             relation.data[self._charm.app]["endpoints"] = json.dumps(self._endpoints())
             logger.debug("Saved endpoints in relation data")
```
mmanciop, line 1285: I would remove this log entirely, it is very little use in `juju debug-log` for Juju admins.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1273,7 +1280,6 @@ def _process_logging_relation_changed(self, relation: Relation):
 
     def _update_relation_data(self, relation):
         if self._charm.unit.is_leader():
-            relation.data[self._charm.app].update(self._promtail_binary_url)
             logger.debug("Saved promtail binary url: %s", self._promtail_binary_url)
             relation.data[self._charm.app]["endpoints"] = json.dumps(self._endpoints())
```
mmanciop, line 1284: Shouldn't this data be in UNIT data bags?

mmanciop (changes requested):

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i: @rbarry82 is taking care of logs in another PR.

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i: @rbarry82 is taking care of logs in another PR.

metadata.yaml
sed-i: Not intended to.
Are there any undesired implications for having it declared?

mmanciop: The rescale does _not_ need a peer relation, because the relation unit databags will be cleaned up by Juju automatically and that is the issue that is making us redesign the relation, which is preferable than having the `loki_push_api` depend on a `peer` relation for reasons of encapsulation.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1314,20 +1318,58 @@ def _regenerate_alert_rules(self):
...
+
+        Args:
+            url: An optional url value to update relation data.
+            relation: An optional instance of `class:ops.model.Relation` to update.
+        """
+        if not relation:
+            if not self._charm.model.get_relation(self._relation_name):
+                return
+
+            relation = self._charm.model.get_relation(self._relation_name)
```
sed-i, line 1338:
```suggestion
        relation = relation or self._charm.model.get_relation(self._relation_name)
        if not relation:
            return
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1314,20 +1318,58 @@ def _regenerate_alert_rules(self):
...
+
+        endpoint = self._endpoint(self._url)
+
+        if url:
+            endpoint = self._endpoint(url)
```
sed-i, line 1343:
```suggestion
        endpoint = self._endpoint(url or self._url)
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1314,20 +1318,58 @@ def _regenerate_alert_rules(self):
...
+        Returns: str
+        """
+        return "http://{}-{}.{}-endpoints.{}.svc.cluster.local:{}".format(
             self._charm.app.name,
-            unit_number,
+            self._charm.unit.name.split("/")[-1],
             self._charm.app.name,
             self._charm.model.name,
             self.port,
         )
```
sed-i, line 1360: Seems like we could do
```python
return "http://{}:{}".format(socket.getfqdn(), self.port)
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1314,20 +1318,58 @@ def _regenerate_alert_rules(self):
...
+    def _endpoint(self, url) -> dict:
+        """Get Loki push API endpoint for a given url.
+
+        Args:
+            url: A loki unit URL.
+
+        Returns: str
+        """
+        endpoint = "/loki/api/v1/push"
+        return {"url": "{}{}".format(url.rstrip("/"), endpoint)}
```
sed-i, line 1371:
```suggestion
        from urllib.parse import urljoin  # at the top of the file though :)
        return urljoin(url, endpoint)
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1680,8 +1723,18 @@ def loki_endpoints(self) -> List[dict]:
...
-            endpoints = endpoints + json.loads(relation.data[relation.app].get("endpoints", "[]"))
+            for unit in relation.units:
+                if unit.app is self._charm.app:
+                    # This is a peer unit
+                    continue
+
+                endpoint = relation.data[unit].get("endpoint")
+                if endpoint:
+                    deserialized_endpoint = json.loads(endpoint)
+                    endpoints.append(deserialized_endpoint)
```
sed-i, line 1736: Would this produce the desired behavior in the case of several related loki _apps_?

metadata.yaml
sed-i: @Abuelodelanada per comments above I think we can drop this for now.

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: If we do this we are going to override the relation that can be passed to the function

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: Nice catch!

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: Yes.

An `agent` (grafana agent) related to 4 different loki apps:

```
Model  Controller  Cloud/Region        Version  SLA          Timestamp
cos    charm-dev   microk8s/localhost  2.9.25   unsupported  17:54:14-03:00

App     Version  Status  Scale  Charm              Channel  Rev  Address         Exposed  Message
agent            active      1  grafana-agent-k8s             2  10.152.183.51   no
jose             active      1  loki-k8s                     14  10.152.183.135  no
kilolo           active      2  loki-k8s                     12  10.152.183.58   no
leon             active      1  loki-k8s                     13  10.152.183.117  no
loki             active      2  loki-k8s                     11  10.152.183.164  no

Unit       Workload  Agent  Address       Ports  Message
agent/0*   active    idle   10.1.157.76
jose/0*    active    idle   10.1.157.100
kilolo/0   active    idle   10.1.157.101
kilolo/1*  active    idle   10.1.157.85
leon/0*    active    idle   10.1.157.102
loki/0*    active    idle   10.1.157.119
loki/1     active    idle   10.1.157.65

Relation provider  Requirer                Interface      Type     Message
jose:logging       agent:logging-consumer  loki_push_api  regular
kilolo:logging     agent:logging-consumer  loki_push_api  regular
leon:logging       agent:logging-consumer  loki_push_api  regular
loki:logging       agent:logging-consumer  loki_push_api  regular
```

The config produced in in agent config file:

```yaml
loki:
  configs:
  - clients:
    - url: http://loki-0.loki-endpoints.cos.svc.cluster.local:3100/loki/api/v1/push
    - url: http://loki-1.loki-endpoints.cos.svc.cluster.local:3100/loki/api/v1/push
    - url: http://kilolo-1.kilolo-endpoints.cos.svc.cluster.local:3100/loki/api/v1/push
    - url: http://kilolo-0.kilolo-endpoints.cos.svc.cluster.local:3100/loki/api/v1/push
    - url: http://leon-0.leon-endpoints.cos.svc.cluster.local:3100/loki/api/v1/push
    - url: http://jose-0.jose-endpoints.cos.svc.cluster.local:3100/loki/api/v1/push
    name: promtail
    positions:
      filename: /tmp/positions.yaml
    scrape_configs:
    - job_name: loki
      loki_push_api:
        server:
          grpc_listen_port: 3600
          http_listen_port: 3500
```

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: Much better!

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1314,19 +1320,48 @@ def _regenerate_alert_rules(self):
...
+
+        relation.data[self._charm.unit].update({"endpoint": json.dumps(endpoint)})
+        logger.debug("Saved endpoint in unit relation data")
+
+    @property
+    def _url(self) -> str:
+        """Get local Loki Push API url.
+
+        Returns: str
+        """
```
sed-i, line 1352:
```suggestion
        """Return url to loki, including port number, but without the endpoint subpath."""
```

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i: Nice!
Does it make sense for `loki_endpoints` to flatten all endpoints? I.e. is there a reason to aggregate them by relation, and have the charm code flatten it?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1680,8 +1716,18 @@ def loki_endpoints(self) -> List[dict]:
             A list with Loki Push API endpoints.
```
sed-i, line 1716: List of strings or list of dicts? I.e. does the return type annotation needs an update?
(And why did mypy not complain..?)

sed-i (commented):
GitHub won't let me approve the PR because I opened it... Anyway, :+1:

balbirthomas (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: I think I'm not getting your point... what do you mean with: "flatten all endpoints" ??

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada: List of dicts:

```python
[{"url": "http://endpoint1"}, "url": "http://endpoint2"}]
```

tests/integration/test_rerelate.py
```diff
@@ -27,19 +27,6 @@
...
-        self.src = src
-        self.relname = relname
-        self.config = config
-
-    async def deploy(self, ops_test: OpsTest):
-        await ops_test.model.deploy(
-            self.src, application_name=self.name, channel="edge", config=self.config
-        )
-
-
```
rbarry82, line 42: :clinking_glasses:

tox.ini
```diff
@@ -83,6 +83,7 @@ deps =
     coverage[toml]
     responses
     fs
+    hypothesis
```
rbarry82, line 86: Are we actually using this? IMO it would be better to remove everywhere it is used.

tests/integration/test_rescale_loki_push_api.py
```diff
@@ -0,0 +1,121 @@
...
+
+Feature: When the Loki charm is rescaled, all the related charms over loki_push_api need to have
+up-to-date information of current (remaining) Loki units.
+
+Scenario:
+1. Deploy the charm with one unit.
+2. Scale up from one to three units.
+3. Scale down to zero units.
+"""
+
```
rbarry82, line 15: I have mixed feelings about this integration test. The `show_unit` helper is *great*. Relying on it for workflow doesn't give any assurances that the data written "makes it" to anything useful, though. @balbirthomas did a great job with `test_alert_rules_fire` which could probably mostly be re-used here (or just add/change that test so we don't have more instances of the "test this with a single unit, then have a whole different build/model/etc for multiple" pattern).

When we scale up, fire the action and make sure alerts flag. That's a reasonable check for "logs made it to the endpoint and were processed." Otherwise, we've all seen enough instances of "this made it into juju relation data but the workload lost it somewhere in the mix"

rbarry82: I just deleted the entire file (the test suite is long enough) and added verification in `test_alert_rules_fire` with an additional testcase which actually scales the units up and makes sure they fire.

Abuelodelanada (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1314,19 +1317,48 @@ def _regenerate_alert_rules(self):
...
+    def _endpoint(self, url) -> dict:
+        """Get Loki push API endpoint for a given url.
+
+        Args:
+            url: A loki unit URL.
+
+        Returns: str
+        """
+        endpoint = "/loki/api/v1/push"
+        return {"url": urljoin(url, endpoint)}
```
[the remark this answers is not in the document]
Abuelodelanada: Not much, We made the following change to make the code easier to read/maintain

```diff
-        return {"url": "{}{}".format(url.rstrip("/"), endpoint)}
+        return {"url": urljoin(url, endpoint)}
```

tox.ini
Abuelodelanada: @sed-i since you add it, it's possible to remove this?

tests/integration/test_rescale_loki_push_api.py
Abuelodelanada: We got rid of `test_rescale_charm.py` since it was not a very focused test and `test_rescale_loki_push_api.py` was added to replace it.... 

As far as I understand, `test_alert_rules_fire` tests are not checking for Loki endpoints in relation data..... or am I missing something?

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i: I am not too familiar with the agent config but I was wondering if it makes sense for the lib to return
```python
[{"url": "http://..."}, ...]
```
per relation. For example:
```python
[
  [{"url": "url: http://loki-0"}, {"url": "url: http://loki-0"}],
  [{"url": "url: http://kilolo-0"}, {"url": "url: http://kilolo-0"}],
]
```
Not sure if it is needed though or if backwards compat is a consideration here.

Abuelodelanada: I think it is not necessary, since the `loki_endpoints` method in `LokiPushApiConsumer` returns endpoints for the current relation

Abuelodelanada (dismissed):

tests/unit/test_charm.py
```diff
@@ -129,6 +273,49 @@ def test_instance_address_is_set_to_this_unit_ip(self):
...
 
+class TestPebblePlan(unittest.TestCase):
+    """Feature: Multi-unit loki deployments.
+
+    Background: TODO
+    """
+
+    @patch("charm.KubernetesServicePatch", lambda x, y: None)
+    @patch_network_get(private_address="1.1.1.1")
+    @given(st.booleans(), st.integers(1, 3), st.integers(0, 3))
```
sed-i, line 284: @Abuelodelanada `hypothesis` is used here.
I added it to save time. Feel free to reduce this to a single test or remove completely.

rbarry82 (approved):

tox.ini
rbarry82: Seems to only be used by:

```
class TestPebblePlan(unittest.TestCase):
    """Feature: Multi-unit loki deployments.
    Background: TODO
    """
```
What does this test do? Who knows? It appears to be nothing and 100% overlap with other tests.

Multi-unit testing can be (and should be, and is) done in integration tests anyway. Just delete this entire testcase, and no more hypothesis anyway.

rbarry82 (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1314,19 +1317,48 @@ def _regenerate_alert_rules(self):
...
+        if not relation:
+            if not self._charm.model.get_relation(self._relation_name):
+                return
+
+            relation = self._charm.model.get_relation(self._relation_name)
+
+        endpoint = self._endpoint(url or self._url)
+
+        relation.data[self._charm.unit].update({"endpoint": json.dumps(endpoint)})
+        logger.debug("Saved endpoint in unit relation data")
```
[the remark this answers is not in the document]
Abuelodelanada: The underlying idea with `update_endpoint` method is not also to provide a way to update endpoints data when a `logging_relation` is established (or changed) between Loki and a consumer charm but also to provide a way to update endpoints (from outside the lib) when an `ingress_relation` is established (or changed).
This way we are going to be able to forward ingress generated URLs to Loki's consumer.

The idea was:

1. An Ingress charm (let say Traefik), Loki and Loki's Consumer charm (Grafana agent) are deployed
2. A relation between Loki and Consumer charm is created
  a. Provider side (Loki) will update units relation data with Loki endpoints using `update_endpoint` method.
  b. Consumer side will get those endpoints using [`loki_endpoints`](https://github.com/canonical/loki-k8s-operator/blob/f9e4b216a062bcd8c72c4a036179ada9380e6f4f/lib/charms/loki_k8s/v0/loki_push_api.py#L1696) method and will update its config file with this data.
3. A relation between Loki and Traefik is created (`ingress-per-unit`).
  a. Traefik will create a new URL for each Loki unit. These URLs can be obtained using the Traefik's [`url` property](https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_k8s/v0/ingress_per_unit.py#L739)
  b. We will get this new URL and call `update_endpoint` method from within the charm. This way we will forward this new endpoints to Loki's consumers since `update_endpoint` will change unit relation data.
  
We have a [work-in-progess PR here](https://github.com/canonical/loki-k8s-operator/pull/120) that needs to be updated once all our stabilisation PRs are merged.

Does this approach still make sense to you or you would prefer to modify it?

Abuelodelanada (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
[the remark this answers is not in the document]
rbarry82: Ah, I got it. Makes perfect sense to me now.

rbarry82 (approved):

Abuelodelanada (approved):

Issue #146, opened 2022-04-13
Loki Charm library should not handle updating alert rules in workload container


balbirthomas:
### Bug Description

At present the `LokiPushApiProvider` library has methods such as `_remove_alert_rules()`, `_generate_alert_rules()` . These methods modify alert rule files in the workload container of the Loki Charm. This violates current practice of maintaining a separation of concern between charm libraries and charms whereby charms manage their workloads and charm libraries manage relation data. 

### To Reproduce

NA

### Environment

NA

### Relevant log output

```shell
NA
```

### Additional context

NA

balbirthomas:
If this proposed refactor is accepted then refactoring `_check_alert_rules()` concomitantly may be necessary. This method should differentiate between the following cases in the information provided to the Provider charm
- Loki server was successfully contacted and there are no errors in alert rules
- Loki server was successfully contacted and errors were found in the alert rules
- Loki server was successfully contacted and no alert rules were found
- There was a failure to contact the Loki server

Abuelodelanada:
I fully agree with the principle: _"maintaining a separation of concern between charm libraries and charms whereby charms manage their workloads and charm libraries manage relation data"_

That been said, from my point of view if we do not place `_remove_alert_rules()`, `_generate_alert_rules()`, `_check_alert_rules()` methods in the library, we have to place these functionalities in the charm side. If we place them in the charm side, charm authors that want to use this lib, will have to implement these methods by themself.

If this is correct we will violate our other principle for libs: _"Write libs that are easy to use. Just fetch the lib, import it, instantiate it and voilà!"_

Is that correct or am I missing something here?

balbirthomas:
Yes indeed charms will have to write alert rules to their workloads themselves. This is how the Prometheus charm does it. The charm library should provide a public method to retrieve the alert rules. The charm library *may also* provide a public method that validates the alert rules using address of Loki server as an input. 

This ticket and the above observations are only suggested considerations for future refactors. The Loki charm developers may have a deeper understanding if these suggestions are appropriate for their charm and hence I trust their better judgement on what is appropriate.

Issue #140, opened 2022-04-11
Refactor to avoid invoking same method in all logging relation changes


balbirthomas:
### Bug Description

This is a proposed refactor of Loki Push API library. This provider object in this library uses a single method `_process_logging_relation_changed()`. Hence this method is forced to check the type of the invoking event and then conditionally act upon them. The complicates the code flow of the method. This method is essentially doing two things
- Inject promtail into relation data for consumer
- Save alert rules provided by consumers into Loki alert rule folder
Hence it is proposed to split this into two methods for each of these functions, 
- `set_promtail()`
- `save_alert_rules()`
With this refactor there can be separate event handlers that invoke these methods as required without the need for duplicating code.

### To Reproduce

NA

### Environment

NA

### Relevant log output

```shell
NA
```

### Additional context

NA

Issue #133, opened 2022-04-05
Loki push API consumer does not handle pebble ready event


balbirthomas:
### Bug Description

The Loki push API consumer object does not handle pebble ready event. This could lead to a potential bug if the consumer receives a relation changed event when pebble is not ready. Such a bug was reported for the provider in issue #128. Shouldn't the consumer also handle the pebble ready event ?

### To Reproduce

NA

### Environment

NA

### Relevant log output

```shell
NA
```

rbarry82:
The consumer does not directly manipulate containers, and only needs to pull `loki_endpoints`, which will presumably be done as part of the configuration for any given consumer charm once Pebble _is_ ready. Did I miss something?

dstathis:
I agree with @rbarry82 here. As the library does not access the container, this should not be an issue.

PR #147, opened 2022-04-14
Avoid setting data in relation departed event handler

balbirthomas:
## Issue
This PR is an trying to extract a minimal subset of work recently done by @rbarry82 @sed-i @balbirthomas and @Abuelodelanada in trying to make the integration tests for the Loki operator more stable. There were multiple issues as detailed below.

1. The integration tests use the cos-config operator. The cos-config operator loaded alert rules from the main github branch of this Loki charm. In order to test behaviour of faulty alert rules an incorrect alert rule was introduced into the set of test rules in this Loki charm. The cos-config operator recursively loaded all alert rules, in all the integration tests (including the fault alert rule). This was one cause of integration test failures. It was not possible to fix this failure without removing the cos-config operator because it was not possible to test the fix locally on a developer machine or on a PR branch. This was because the cos-config operator loaded the alert rules from the Loki operators main branch on github. 
2. Integration tests were also failing due to exceptions raised when the LokiPushApi provider tried to write to relation data when handling a relation departed event. This commit minimally refactors the Loki Operator to avoid this behaviour. Further refactor which may be useful have been documented in various github issues and are hence not implemented as part of this PR.
3. After replacing cos-config with loki-tester it was observed that charmcraft cached the state of the loki-tester. For example if an integration test inserted a faulty alert rule into loki-tester rebuilt and redeployed it, the deployed loki-tester was still missing the newly inserted alert rule. The same problem was observed if an alert rule was removed, loki-tester rebuilt and redeployed.
4. For quite some time (months) it was observed that errors used to be raised by python-libjuju and pytest-operator when the integration tests were tearing down models. However these errors did not lead to failing tests. However recent changes to these tools surfaced these failures as failures of the integration tests even though these errors occurred after the tests passed.
5. The loki operator sets its status to active even before its loki server workload is actually active and ready to service HTTP requests. As a result of this a prior integration test was using tenacity to query the loki server's status repeatedly till it passed. 
6. Juju sporadically results in a hung state when trying to install charms.
7. Juju sporadically results in a hung state when trying to remove applications and/or models.

## Solution
1. Replaced cos-proxy with a Loki Tester charm in all affected integration tests.
2. Minimally refactored Loki Operator to avoid writing application data in relation departed event. This refactor starts the process of moving away from using a single event handler of LokiPushApiProvider to handle multiple relation events with the consumer. Further refactors may be effected in subsequent commits such as those mentioned in issues https://github.com/canonical/loki-k8s-operator/issues/140, https://github.com/canonical/loki-k8s-operator/issues/146, https://github.com/canonical/loki-k8s-operator/issues/133, https://github.com/canonical/loki-k8s-operator/issues/149 and others

Specifically this change ensures the following

- There are three separate event handler for
   + upgrade event
   + relation departed event
   + relation changed event

- The upgrade event handler iterates over all relations with
   consumers and does the following
   + updates relation data for consumers
   + checks and updates all alert rules obtained from consumers

- On relation departed the LokiPushApi provider only updates
   its own alert rules. Prior to this commit the provider
   also set relation data for the consumer and checked alert
   rules. Neither of this are necessary. For example no new
   alert rule is obtained in a relation departed event hence
   there is no need to recheck the alert rules.

- This change also puts simple pebble can connect guards for
   two existing code blocks. Future refactors may make these
   guards more intelligent.

3. The integration test helper methods used to build loki-tester have been updated to ensure that `charmcraft clean` is invoked after inserting or removing additional alert rule files.
4. Both python-libjuju and pytest-operator were reverted to older versions known not to cause false positive integration test failures.
5. The use of tenacity was extended to another integration test that was checking if loki server is alive before it actually had the time to spin up. However an github issue ticket was created to address this issue properly in a future PR : https://github.com/canonical/loki-k8s-operator/issues/149. 
6. It is not possible to address the sporadic "hang on installing agent" through changes to a charm's integration testing code. This problem can only be resolved in Juju. It may be possible to detect this and restart the test but this is perhaps best left to human intervention to do so.
7. Exceptions raised by python-libjuju and pytest-operator on application removal are trapped in the tests where they are frequently observed. This allows the test to continue execution and check the test assertion to see if the tests still pass or not.

## Context
As described above

## Testing Instructions
All integration tests need to be run repeatedly to check they consistently pass.

## Release Notes
- Improved stability of integration tests
- Partially refactored Loki charm to have separate handler for each event.


Abuelodelanada (dismissed):

tests/integration/test_rerelate.py
```diff
@@ -94,28 +94,59 @@ async def test_rerelate(ops_test: OpsTest):
...
+            app,
+        ]
+        logger.info("Forcibly removing {}".format(app))
+        await ops_test.run(*cmd)
+
+    logger.info("Waiting for idle")
+    try:
+        await ops_test.model.wait_for_idle(status="active", timeout=300)
+    except asyncio.exceptions.TimeoutError:
+        logger.warning("Timeout waiting for idle, ignoring it.")
```
simskij, line 127: Just swallowing these errors in the charm code does not seem like the right solution to me.

tests/integration/test_alert_rules_fire.py
```diff
@@ -70,6 +69,8 @@ async def test_alert_rules_do_forward_to_alertmanager(ops_test, loki_charm, loki
     alertmanager_app_name = "alertmanager"
     app_names = [loki_app_name, tester_app_name, alertmanager_app_name]
 
+    await ops_test.model.set_config({"logging-config": "<root>=WARNING; unit=DEBUG"})
+
```
simskij, line 73: Looks like a debug statement that should be removed prior to merging?

simskij (changes requested):
In general I think this PR contains a lot of good things, like untangling the common exit hook, adding `can_connect` guards and such. 

What I do not agree with is adding manual clean up test cases, or swallowing errors to make the test pass.

tests/integration/test_rerelate.py
balbirthomas: True. However these errors are being generated by Juju/libjuju. And the test case assert is to check if loki is alive after all the changes made by the test case. That assert has not been removed or compromised and that assert still passes. Short of fixing the issue with Juju,  I see very limited options here. Besides all test cases run to completion rather than hanging for hours. There are two Juju issues
- Hang on installing agent
- Hang on removing model
Both of these issues are seen in other charms and I have highlighted this by reviewing PRs in other repositories today.

tests/integration/test_alert_rules_fire.py
balbirthomas: I am happy to remove. Pinging @sed-i to get a second opinion if this is still required and why it was added in the first place.

tests/integration/test_rerelate.py
balbirthomas: Also the forced removal of applications has existed long before this PR. @sed-i Does this imply that the "Hang on removing model" has also existed for quite some time across multiple Juju releases.

simskij: My bad, I went through this too quickly

simskij (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1252,17 +1261,22 @@ def _process_logging_relation_changed(self, relation: Relation):
         Args:
             relation: the `Relation` instance to update.
         """
+        self._update_relation_data(relation)
+        self._update_alert_rules(relation)
+        self._check_alert_rules()
```
rbarry82, line 1266: It definitely seems that `_check_alert_rules()` should simply be called from `_update_alert_rules()`. If it is not, there is a possibility that the rules are _not_ rechecked after generation during `relation-departed`

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1252,17 +1261,22 @@ def _process_logging_relation_changed(self, relation: Relation):
...
+
+    def _update_relation_data(self, relation):
         if self._charm.unit.is_leader():
             relation.data[self._charm.app].update(self._promtail_binary_url)
             logger.debug("Saved promtail binary url: %s", self._promtail_binary_url)
             relation.data[self._charm.app]["endpoints"] = json.dumps(self._endpoints())
             logger.debug("Saved endpoints in relation data")
```
rbarry82, line 1273: I also changed this is my patch to verify things first, but it should be done here also. It's amazingly chatty for no reason whatsoever, and ends up with [weird guards like this one](https://github.com/canonical/loki-k8s-operator/blob/cc07f57faa1a15c0ad62d87ae3edb4df109aecf0/lib/charms/loki_k8s/v0/loki_push_api.py#L1505).

Imagine a consumer joins.

A  `relation-changed` event is fired from that side, and the provider responds with more updates, which triggers another change on the other side, and so on. Endlessly. We are relying on Juju inferring that the data is the same and it should _not_ fire an event to break this loop, and it's unnecessary.

Neither the Provider nor the Consumer listens to `_joined` at all, where "initial data" (such as juju topology/"metadata" from the consumer and the promtail URL on the provider) from the consumer side should be set, since we are reasonably sure that they will never change.

This is questionable at best and dangerous at worst, since the only reason this is working at is because `RelationChangedEvent` always fires on joined, and because Juju is checking for diffs in the state data to prevent a loop.

The pattern violates the basic assumptions relations [operate under](https://juju.is/docs/sdk/relations). In particular:

> Callback methods bound to this event should be the only ones that rely on remote relation settings. They should not error if the settings are incomplete, since it can be guaranteed that when the remote unit or application changes its settings, the event will fire again.

Nothing in this handler on either side needs this. For lifecycle events, sure. For `RelationChangedEvent`, this should just be in Joined.

tests/integration/conftest.py
```diff
@@ -21,9 +18,43 @@ def copy_loki_library_into_tester_charm(ops_test):
...
+    try:
+        os.remove(bad_rule_path)
+    except FileNotFoundError:
+        pass
     charm = await ops_test.build_charm(charm_path)
     return charm
+
+
+@pytest.fixture(scope="module")
+async def faulty_loki_tester_charm(ops_test):
```
rbarry82, line 44: I understand that this was added for resiliency, but this adds an enormous amount of runtime to the integration tests simply to avoid `cos-config`.

As long as `cos-config` "points" at a non-`main` branch here, we can avoid the massive runtime hit from completely cleaning and rebuilding both charms on every test.

tests/integration/test_alert_rules_fire.py
rbarry82: It's not needed. It was added so one of the models hanging (either at `0/1` units or `Installing agent`) would have more rigorous logs "out of the box".

tests/integration/test_alert_rules_load.py
```diff
@@ -16,56 +16,38 @@
 METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
 app_name = METADATA["name"]
 resources = {"loki-image": METADATA["resources"]["loki-image"]["upstream-source"]}
-rules_app = "cos-config"
-rules_app2 = "cos-config2"
-rules_app3 = "cos-config3"
+rules_app = "loki-tester"
```
rbarry82, line 19: As a general comment, we should just `git rm` this entire file. It's obvious from the changes below and from running it that there are no meaningful differences between this and `test_multiple_rule_providing_apps`

Abuelodelanada (dismissed):

tests/integration/test_rerelate.py
rbarry82: The forced removal of applications is new to this PR because of the hanging problem. You will not find it in current `main`.

If the hang on removing an application occurs in any other charm, please indicate it. I didn't see any GH comments about it, and testing indicated that it was restricted only to charms using `loki_push_api` (so this and grafana-agent, basically, plus cassandra).

The point of integration tests is not simply "is the limited scope of my testcase ok", but "is this ok in actual operation". Failing to remove is an excellent example of "things are not ok", and we should not intentionally mask it. Similar to the large amount of mocking, "my tests are green" isn't the only meaningful thing, otherwise they'd all just `return True`. 

This could very easily also cover up a failed `relation-departed` or `relation-broken` hook _not_ related to Juju. Since Juju 2.9.29 resolves, pinning to that should be used, and this exception handling removed. Not every problem is a Juju problem.

rbarry82 (changes requested):

tests/integration/conftest.py
balbirthomas: This is an issue with charmcraft. A `charmcraft clean` will not be necessary once charmcraft caching is more intelligent.

lib/charms/loki_k8s/v0/loki_push_api.py
balbirthomas: When a LokiPushAPI consumer departs al that is required is to remove the alert rules corresponding to that consumer. There is no need to recheck all the existing rules which were already checked when they were inserted in the first place.

lib/charms/loki_k8s/v0/loki_push_api.py
balbirthomas: Can we take such issues in future refactors. The purpose of this PR is to implement the *minimal* set of changes that brings a measure of stability to the integration tests. The whole point of doing this is so that we can progress the development of the Loki charm. As mentioned in the PR description various github issue tickets have been opened to consider further improvements to Loki operator.

tests/integration/test_alert_rules_load.py
balbirthomas: I agree. But I thought it would be cheating to remove tests as a way to demonstrate improvements in stability of integration tests in a PR trying to do so.

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: `_check_alert_rules()` verifies "is Loki accessible?", and that is still absolutely necessary.
 
 If, for some reason, removing the alert rules put Loki into an error state, we would end up with a charm in `ActiveStatus` with a bad workload. This should not happen.
 
 Conversely, if the charm _was_ in `BlockedStatus` due to bad alert rules, a successful check will still send an event back to the charm which clears that status and puts it into `ActiveStatus` if the check is ok.
 
 Skipping this step is not optional if we want `juju status` to have any meaning.

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: Not to put too fine a point on it, but we **do not know the "minimal" set of changes**. Catching exceptions in the integration tests so they turn green does not address what the "minimal" set of changes may be, and we cannot stand on that as an example unless the exception handling is _removed_ and they still pass.

tests/integration/conftest.py
rbarry82: Frankly, "this is an issue with something else" is both true and meaningless. 

"Unless the opposition to using rules in `cos-config` for the single error check which would avoid this entire clean/repack thing is something other than philosophical, we are cutting of our nose to spite our face.

A `charmcraft clean` would also not be necessary if the error rules check pulled from a different branch (which I recreated so it passed in my PR). Using `loki-tester` is great. It does not have to be this prescriptive.

lib/charms/loki_k8s/v0/loki_push_api.py
balbirthomas: - Issue #149 was created to address the problem of checking "is Loki Accessible".
- Please do create additional issues for suggested improvements. This PR is not meant to be that last one there ever will be with Loki improvements.

lib/charms/loki_k8s/v0/loki_push_api.py
balbirthomas: I have explained in PR description why adding the exception handlers is not so that the tests "turn green". It is to allow the test and all subsequent tests in the same module to run to completion and check all their asserts in the process. It is those asserts that decide if the tests should be green or not.

Once again please do create additional issues for any other improvements you feel necessary. This purpose of this PR is not to fix all the problems of Juju, python-libjuju and pytest-operator.

tests/integration/conftest.py
balbirthomas: One lesson I would draw from item 1 in the PR description is when ever possible if we can use test doubles and charm local tests we should.

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: I think that PR is an improvement is great. But this PR also potentially breaks the Loki workflow.

Picture the following:

* Deploy Loki
* Deploy another charm with "good" alert rules, and a charm with "bad" alert rules
* Relate the charm with "good" alert rules. `_check_alert_rules()` happens, a ` LokiPushApiAlertRulesChanged` event is emitted, the charm gets it, and the `error` state is null, so it stays active.
* Relate the charm with the "bad" alert charm. `_check_alert_rules()` fails, a ` LokiPushApiAlertRulesChanged` event is emitted, the charm gets it, the `error` state is True, so it puts the charm in `BlockedStatus` with the message.
* Remove the relation to the "bad" rules charm. With this PR, `_check_alert_rules()` is never fired, no event is emitted, Loki is not checked, and the charm stays in `BlockedStatus()` with a message about bad alert rules from a relation that doesn't exist anymore.
* `juju status` and the state of the charm do not reflect the reality.

It's not about whether or not we should need tenacity or the "check the endpoint constantly in `charm.py`" part, this is just basic workflow, and this PR breaks it unless `_check_alert_rules()` is also checked on departed/broken.

Abuelodelanada: You made a solid point about `juju status` @rbarry82.!

`_update_alert_rules()`  is only called [here](https://github.com/canonical/loki-k8s-operator/blob/cc07f57faa1a15c0ad62d87ae3edb4df109aecf0/lib/charms/loki_k8s/v0/loki_push_api.py#L1250), and [here](https://github.com/canonical/loki-k8s-operator/blob/cc07f57faa1a15c0ad62d87ae3edb4df109aecf0/lib/charms/loki_k8s/v0/loki_push_api.py#L1265)

What about always calling `_check_alert_rules()` **after** `_update_alert_rules()`? This way we will not break the single responsibility principle.... my two cents

balbirthomas: This problem can not be solved without addressing issues such as #149 , [PR 135](https://github.com/canonical/loki-k8s-operator/pull/135) and a couple of others. My point is *not* that this PR fixes all issues. It is that we need to put our integration tests in a sane state so we can start addressing all the issues including this one. It has been half a month since we started trying to unblock Loki PRs.

balbirthomas: We got to this point because we were happy to merge PRs when integration tests were Red but now we insist on the most perfect PR that solves all the problems even if integration tests are passing.

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: This is middle-ground position which does not make sense.

To "fix" the integration tests:

* Replace `cassandra` with `loki-tester`
* Fix `_endpoints`
* Get `loki-tester` built without the error-ed alert rules nested in the path
* Either `faulty-loki-tester` or `cos-config` (with the re-created `error_alerts`) branch.
* Use an older version of `python-libjuju` and `pytest-operator` which doesn't throw `asyncio` exceptions all the way up the stack.

That's pretty much it. The rest of the changes around common event handling/etc are nice to have, but they are not a baseline requirement if that is our goal.

As repeatedly stated, if the "scale to `0/1` problem" happens _only_ with charms using `loki_push_api`, is basic differential diagnosis that this is not necessarily a problem with Juju. That is the "Installing agent..." bug we see all over the place.

Trapping exceptions around failing to tear down the model is like waiting for your kettle to boil while the house burns down. Allowing the tests in the same module to run to completion isn't meaningful considering that one of the intended test results was "destroy these applications". Otherwise, we could just remove the relations, deploy brand-new charms, and move on with the "leftover" charms related to nothing and alive in the model.

The fact that we very explicitly `remove()` and then wait is an assertion in and of itself, in that it will elevate problems in `relation-departed|broken` when those applications _fail_ to be removed and the state of the model never matches what we expect.

It is, very purely,"turning the tests green", especially if Juju 2.9.29 resolves it. Then we should just use that and cut the exception handling.

balbirthomas: - Have you tried all the above suggested changes ? 

- Do they produce integration tests that run to completion and at least a shade more robust than we currently have ?

- Have you identified what about the Loki operator causes the "scale to `0/1`" issue as you call it ? I see it is juju instability issues mentioned in the PR description. I see no evidence to support that this is specific to Loki. 

- Without such concrete results aren't we chasing ghosts ?

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: What? No.

We are not talking about the same problem.

We cannot merge a patch which fundamentally breaks the workflow of Loki while we wait for other issues and PRs for the integration tests.

The ask here is easy: call `_check_alert_rules()` in departed so a `LokiPushApiAlertRulesChanged` is emitted to the charm so a charm which _was_ in `BlockedStatus` can (correctly) go back to `ActiveStatus` if Loki successfully reloads, and so Loki can restart to reload the alert rules (again, I'm not sure if Loki requires this for provisioning)

PR #154, opened 2022-04-27
No common exit hook in lib

### LINKED ISSUE #65: LokiPushApi events review
Shouldn’t `LogProxyConsumer` also observe `relation_broken` for safety? Also it needs `relation_joined` for new consumer units being spun up, e.g., after a pod churn, as those do not get the `relation_created`, and I am not sure `relation_changed` will fire.

`LokiPushApiProvider`: alert rules probably should be deleted on `relation_broken`: relation_departed is for single units going away, but alert rules are global to the consumer Juju app. Also, same comment about `relation_joined` applies.

_Originally posted by @mmanciop in https://github.com/canonical/loki-k8s-operator/issues/63#issuecomment-1007194880_

### LINKED ISSUE #140: Refactor to avoid invoking same method in all logging relation changes

rbarry82:
## Issue
Use of a common handler for virtually all relation events meant that there was a somewhat excessive amount of chatter between provider/consumer.

Move the relevant data/updates to the events they "belong to".

Closes #140 
Closes #65 

## Solution
Set relatively static data (alert rules, promtail_binary_url,
topology/metadata) in relation_joined. Fire events on
relation-changed so endpoints are still updated.

Streamline event handling.

Remove more mocking.

Make the state-dependent unit tests for reloading alert rules
explicitly ignore sorting. Unittests should:

* Be in a `unittest.TestSuite` if ordering matters
* Not be dependent on the order of previous tests, since `unittest` sorts (by default) using `dir()` and alphabetic sorting. Making functions "digestible" at the cost of predictability is not worth it

## Context
We all know the common exithook.

## Testing Instructions
Unit+integration tests

## Release Notes
Remove common exit hook from loki_push_api

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -1,7 +1,7 @@
 name: Pull Request
 
 on:
-  pull_request: 
+  pull_request:
     branches:
       - main
 
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
         with:
-          fetch-depth: 0  
+          fetch-depth: 0
       - name: Check libs
         uses: canonical/charming-actions/check-libraries@1.0.1-rc
         with:
@@ -36,7 +36,7 @@ jobs:
         run: python3 -m pip install tox
       - name: Run static analysis for /lib for 3.5
         run: tox -vve static-lib
-  
+
   static-charm:
     name: Static analysis of the charm and tests
     runs-on: ubuntu-latest
@@ -99,7 +99,7 @@ jobs:
       - name: Setup operator environment
         uses: charmed-kubernetes/actions-operator@main
         with:
-          juju-channel: latest/stable
+          juju-channel: 2.9/edge
           provider: microk8s
       - name: Run integration tests
         run: tox -vve integration
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -99,7 +99,7 @@ jobs:
       - name: Setup operator environment
         uses: charmed-kubernetes/actions-operator@main
         with:
-          juju-channel: 2.9/edge
+          juju-channel: 2.9/candidate
           provider: microk8s
       - name: Run integration tests
         run: tox -vve integration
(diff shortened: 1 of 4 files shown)


dstathis:
I'm not really seeing how this reduces chatter. Data will only be sent if you actually *change* the relation data. Rewriting the promtail url, for example, won't trigger anything.

sed-i (dismissed):

rbarry82:
> I'm not really seeing how this reduces chatter. Data will only be sent if you actually _change_ the relation data. Rewriting the promtail url, for example, won't trigger anything.

One of the biggest struggles in trying to untangle the bug which blocked us was in actually isolating the behavior from any given state to see _what_ was triggering it. We already have [at least one workaround](https://github.com/canonical/loki-k8s-operator/blob/e5e0cc40da4d454783a39919b99414ff93c9f151/lib/charms/loki_k8s/v0/loki_push_api.py#L1544) to prevent pointless `relation-changed` events.

If there is _ever_ a behavior change in Juju, relying on a single method only puts us into an endless loop of:
* -> alert rules changed
* <- endpoints changed
* -> alert rules
* <- endpoints
* -> alerts
* <- endpoints

Sure, this would be a bug, but we aren't regularly testing `beta|edge` versions. How long would an emergency fix take to make it into `stable`?

In addition, in terms of sheer load outside of a purely development workflow, the best way to reduce the number of RPC/API calls is to not send everything all the time. It's the same principle as GraphQL. The best way to reduce the load from state comparisons is to not do them at all. The "chatter" here is not just betwen charms, but also between charms and the controller itself.

Abuelodelanada (approved):

sed-i (approved):

PR #155, opened 2022-04-27
Handle missing promtail configuration during comparisons

### LINKED ISSUE #62: log_proxy_departed fails if the promtail config file does not exist
This is not very high priority since it should only happen when another bug causes the file to not exist. (At time of writing such a bug exists #60).

The bug here is that the `_build_config_file` requires the config file to exist already. This seems like it could possibly lead to more serious issues down and we should consider fixing it.

rbarry82:
## Issue
If the promtail configuration is missing for any reason, a Pebble pull will throw an exception back to `LogProxyConsumer` which was not handled

Closes #62 

## Solution
Add both a `can_connect` guard and an exception handler around `ProtocolError | PathError` from Pebble. Return an empty dict if either is true

## Testing Instructions
Relate promtail. Delete the configuration file and break the relation.

## Release Notes
Handle missing promtail configuration during comparisons

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1329,14 +1329,14 @@ class LokiPushApiProvider(RelationManagerBase):
             )
             return False
         except URLError as e:
-            logger.error("URLERROR: Checking alert rules: %s", e.reason)
+            logger.error("Checking alert rules: %s", e.reason)
             self.on.loki_push_api_alert_rules_changed.emit(
                 error=True,
                 message="Errors in alert rule groups. Check juju debug-log.",
             )
             return False
         except Exception as e:
-            logger.error("EXCEPTION: Checking alert rules: %s", e)
+            logger.error("Checking alert rules: %s", e)
             self.on.loki_push_api_alert_rules_changed.emit(
                 error=True,
                 message="Errors in alert rule groups. Check juju debug-log.",
@@ -1357,13 +1357,11 @@ class LokiPushApiProvider(RelationManagerBase):
     def unit_ip(self) -> str:
         """Returns unit's IP."""
         bind_address = ""
-        logger.warning("Trying to get IP")
         if self._charm.model.relations[self._relation_name]:
             relation = self._charm.model.relations[self._relation_name][0]
             bind_address = relation.data[self._charm.unit].get("public_address", "")
 
         if bind_address:
-            logger.warning("Returning IP %s", str(bind_address))
             return str(bind_address)
         logger.warning("No address found")
         return ""
@@ -1392,7 +1390,7 @@ class LokiPushApiProvider(RelationManagerBase):
         if not container.can_connect():
             return
 
-        logger.warning("Generating alert rules files")
+        logger.debug("Generating alert rules files")
         for identifier, alert_rules in self.alerts().items():
             filename = "{}_alert.rules".format(identifier)
             path = os.path.join(self._rules_dir, filename)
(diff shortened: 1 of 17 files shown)


dstathis (dismissed):
Looks good. Only thing I would change for next time is try to keep changes in scope. If you want to fix a bunch of unrelated logging, just open another PR.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -2021,9 +2010,19 @@ def _current_config(self) -> dict:
...
-        raw_current = self._container.pull(WORKLOAD_CONFIG_PATH).read()
-        current_config = yaml.safe_load(raw_current)
-        return current_config
+        if not self._container.can_connect():
+            logger.debug("Could not connect to promtail container!")
+            return {}
+        try:
+            raw_current = self._container.pull(WORKLOAD_CONFIG_PATH).read()
+            return yaml.safe_load(raw_current)
+        except (ProtocolError, PathError) as e:
```
balbirthomas, line 2019: Suppose the Loki provider has just received its first relation departed event i.e. for its last LogProxyConsumer. Now suppose that this `_current_config()` method returns an empty dictionary because it got a Protocol Error or Path Error.

In this case I presume the new config should also be an empty dictionary. Hence both current config and new config will be the same when processing the relation departed handler. Is this not a problem ?

tests/unit/test_log_proxy_consumer.py
```diff
@@ -274,6 +275,11 @@ def test__promtail_must_be_downloaded_in_workload_sha256_match(self):
                 ],
             )
 
+    @patch("ops.model.Container.pull")
```
balbirthomas, line 278: If my comment above regarding handling of relation departed event merits consideration, should there be a test case to check that the edge case mentioned there is tested for ?

sed-i: Can we avoid mocking pull here?

sed-i (dismissed):

rbarry82:
> Looks good. Only thing I would change for next time is try to keep changes in scope. If you want to fix a bunch of unrelated logging, just open another PR.

The unrelated logging actually snuck in with the patch which finally cleared CI, and shouldn't be there at all. Since I can't just revert on `main` :|

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: The new config (or the barebones template from `_promtail_config`) is set [here](https://github.com/canonical/loki-k8s-operator/blob/1152716c1198833aa7be40dbc7dba6d30d0ca371/lib/charms/loki_k8s/v0/loki_push_api.py#L2028), and will be at least:

```python
{
  "clients": [],
   "server": {
      "http_listen_port": HTTP_LISTEN_PORT,
      "grpc_listen_port": GRPC_LISTEN_PORT,
   },
   "positions": {
     "filename": WORKLOAD_POSITIONS_PATH
   },
   "scrape_configs": []
}
```

The comments always bear consideration, though.

tests/unit/test_log_proxy_consumer.py
rbarry82: We could, but since we explicitly want it to return an exception as a side effect (rather than mocking away the actual behavior of pulling a file in the harness), the alternative is just to simply use the harness `FakePebbleClient` to try to get a path which doesn't exist. I can do either, but it seems pretty innocuous here.

Abuelodelanada (approved):

balbirthomas (approved):

sed-i (approved):

Abuelodelanada (dismissed):

dstathis (approved):

Abuelodelanada (approved):

dstathis (approved):

PR #156, opened 2022-04-28
Rework AlertRules.group_name for safety+security

### LINKED ISSUE #47: Instantiating the library with alert rule paths as relative paths will break it
Commenting here so it's close, since I missed the last review, but as happy as I am to see a nested function as `_group_name`, I suppose it should be moved out since we decided not to use this as a pattern.

Sorry in advance, since this is long for a part of the library which isn't even *in* this PR.

`def _group_name(path) -> str:` should be top-level (and then typing it as `typing.Union[Path.PosixPath, Path.WindowsPath]`) wouldn't hurt. and using `path.relative_to`

Also:

```
return "{}_{}alerts".format(
    topology.identifier, "" if relpath == "." else relpath.replace(os.path.sep, "_") + "_"
)
```

This hurts to parse.

This is a nit, really, and I hate commenting on committed code, but it _is_ a nested function, and as long as it's being moved, it could look nicer. `os.path` is ok, but the format is definitely a place where it's not intuitive when `+ "_"` is actually applied, and re-reading it to grok it is slow. It looks like a smell, and the double `)` from `.format()` and `.replace()` makes me worry that later patches could subtly break it.

Arguably, the current formatting is a fragile, since relative paths with "../" or something, absolute paths will lead to `__`, this would not work on any Windows charms, as a library function in the proxy, etc. May as well code defensively as long as I'm looking at it.

It would be very possible, right now, to get a string like the following on Windows (or just keep the `..` part on Linux), and that may be an ugly bug.

`{topology.identifier)_C:_foo_.._bar_baz_alerts`

Pathlib ends up being somewhat more verbose, but much more flexible/resilient. Even if it isn't used, we should absolutely strip out `../` and `^([A-Z]+:)?/` with a regex and stick with the rest of `os.path`.

```
relpath = path.parent.relative_to(dir_path)

# We should account for both absolute paths and Windows paths. Convert it to a POSIX
# string, strip off any leading /, then join it

if relpath.samefile(Path(".")):
    path_str = ""
else:
    # Get rid of leaving / and optionally drive letters so they don't muck up the template later,
    # since Path.parts returns them. The 'if relpath.is_absolute ...' isn't even needed since re.sub
    # doesn't throw exceptions if it doesen't match, so it's optional, but it makes it clear what we're
    # doing.
    
    # Note that Path doesn't actually care whether the path is valid just to instantiate the object, so
    # we can happily strip that stuff out to make templating nicer
    relpath = Path(re.sub(r'^([A-Za-z]+:)?/', '', relpath.as_posix()) if relpath.is_absolute() else relpath
    
    # Get rid of relative path characters in the middle which both os.path and pathlib leave hanging
    # around. We could use path.resolve(), but that would lead to very long template strings when
    # rules come from pods and/or other deeply nested charm paths
    char_filter = lambda x: x not in ["..", "/"]
    path_str = "{}_".format("_".join(filter(char_filter, relpath.parts)))

return "{}_{}alerts.format(
    topology.identifier, path_str
)
```

_Originally posted by @rbarry82 in https://github.com/canonical/loki-operator/pull/45#discussion_r762172941_

rbarry82:
## Issue
Naive handling of paths may lead to side channel attacks if a clever author invokes the library with `../../../proc/pid/1` or similar. Handle edge cases.

Closes #47 

## Solution
Stop using `os.relpath` and use `pathlib.Path` instead to get
better/more modern resolution of paths and their separators.

Strip out possibly problematic characters (like `..` and `/`) from
paths.

Fix the filter since `filter(None, ...)` does not do what it looks
like on the box.

## Testing Instructions
Run unit tests and integration tests

## Release Notes
Rework AlertRules.group_name for safety+security

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -319,7 +319,7 @@ resources:
   promtail-bin:
       type: file
       description: Promtail binary for logging
-      filename: promtail-linux-amd64
+      filename: promtail-linux
 ```
 
 Which would then allow operators to deploy the charm this way:
@@ -440,15 +440,16 @@ import logging
 import os
 import re
 import typing
+import platform
 from collections import OrderedDict
 from copy import deepcopy
+from gzip import GzipFile
 from hashlib import sha256
 from io import BytesIO
 from pathlib import Path
 from typing import Dict, List, Optional, cast
 from urllib import request
 from urllib.error import HTTPError, URLError
-from zipfile import ZipFile
 
 import yaml
 from ops.charm import (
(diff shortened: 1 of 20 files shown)


sed-i (approved):

Abuelodelanada (approved):

sed-i (approved):

simskij (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -966,16 +973,41 @@ def _group_name(self, root_path: str, file_path: str, group_name: str) -> str:
...
+        file_path = Path(file_path) if not isinstance(file_path, Path) else file_path
+        root_path = Path(root_path) if not isinstance(root_path, Path) else root_path
+        rel_path = file_path.parent.relative_to(root_path.as_posix())
+
+        # We should account for both absolute paths and Windows paths. Convert it to a POSIX
+        # string, strip off any leading /, then join it
+
+        path_str = ""
+        if not rel_path == Path("."):
+            # Get rid of leaving / and optionally drive letters so they don't muck up
```
dstathis, line 985: "leading" not "leaving"

dstathis (approved):

Abuelodelanada (approved):

dstathis (dismissed):

Abuelodelanada (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -969,16 +976,41 @@ def _group_name(self, root_path: str, file_path: str, group_name: str) -> str:
         Returns:
             New group name, augmented by juju topology and relative path.
         """
-        rel_path = os.path.relpath(os.path.dirname(file_path), root_path)
-        rel_path = "" if rel_path == "." else rel_path.replace(os.path.sep, "_")
+        file_path = Path(file_path) if not isinstance(file_path, Path) else file_path
+        root_path = Path(root_path) if not isinstance(root_path, Path) else root_path
+        rel_path = file_path.parent.relative_to(root_path.as_posix())
```
balbirthomas, line 981: Do we want to trap `ValueError` here ?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -969,16 +976,41 @@ def _group_name(self, root_path: str, file_path: str, group_name: str) -> str:
...
+            # isn't even needed since re.sub doesn't throw exceptions if it doesn't match, so it's
+            # optional, but it makes it clear what we're doing.
+
+            # Note that Path doesn't actually care whether the path is valid just to instantiate
+            # the object, so we can happily strip that stuff out to make templating nicer
+            rel_path = Path(
+                re.sub(r"^([A-Za-z]+:)?/", "", rel_path.as_posix())
+                if rel_path.is_absolute()
+                else str(rel_path)
+            )
```
balbirthomas, line 999: This part was troubling to understand. It is not that regular expression which is simple enough. It is the fact that you are invoking `relative_to()` above to generate `rel_path` but yet checking `rel_path.is_absolute()` here. The otherwise useful comments did not help in clarifying this point either.

balbirthomas (approved):

PR #157, opened 2022-04-28
Add unit tests for startup events ordering

### LINKED ISSUE #75: Write unit test where `relation-departed` runs before `pebble-ready`
This happened. Let's make sure it doesn't happen again.

sed-i:
## Issue
Closes #75 

## Solution
Write unit tests to address scenario.

## Testing Instructions
`tox -e unit -- -k TestDelayedPebbleReady`

## Release Notes
Add unit tests for startup event ordering.

Crossref: [OPENG-402](https://warthogs.atlassian.net/browse/OPENG-402)

### DIFF
--- a/tests/unit/helpers.py
+++ b/tests/unit/helpers.py
@@ -8,6 +8,10 @@ from typing import Callable, Tuple
 from unittest.mock import patch
 
 
+def tautology(*args, **kwargs) -> bool:
+    return True
+
+
 def patch_network_get(private_address="10.1.157.116") -> Callable:
     def network_get(*args, **kwargs) -> dict:
         """Patch for the not-yet-implemented testing backend needed for `bind_address`.
--- a/tests/unit/helpers.py
+++ b/tests/unit/helpers.py
@@ -4,7 +4,30 @@
 
 import os
 import tempfile
-from typing import Tuple
+from typing import Callable, Tuple
+from unittest.mock import patch
+
+
+def patch_network_get(private_address="10.1.157.116") -> Callable:
+    def network_get(*args, **kwargs) -> dict:
+        """Patch for the not-yet-implemented testing backend needed for `bind_address`.
+
+        This patch decorator can be used for cases such as:
+        self.model.get_binding(event.relation).network.bind_address
+        """
+        return {
+            "bind-addresses": [
+                {
+                    "mac-address": "",
+                    "interface-name": "",
+                    "addresses": [{"hostname": "", "value": private_address, "cidr": ""}],
+                }
+            ],
+            "egress-subnets": ["10.152.183.65/32"],
+            "ingress-addresses": ["10.152.183.65"],
+        }
+
+    return patch("ops.testing._TestingModelBackend.network_get", network_get)
 
 
 class TempFolderSandbox:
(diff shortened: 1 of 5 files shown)


rbarry82 (approved):

tests/unit/test_charm.py
```diff
@@ -150,3 +155,57 @@ def test__loki_config(self):
...
+
+        # GIVEN this unit is the leader
+        self.harness.set_leader(True)
+
+        # AND a "logging" app joins with several units
+        self.log_rel_id = self.harness.add_relation("logging", "consumer-app")
+        self.harness.add_relation_unit(self.log_rel_id, "consumer-app/0")
+        self.harness.add_relation_unit(self.log_rel_id, "consumer-app/1")
+        self.harness.begin_with_initial_hooks()
+        # self.harness.charm.loki_provider._check_alert_rules = tautology
```
sed-i, line 179: @rbarry82 may I get a patching hint?
The decorator doesn't work for me here and I do not want to monkeypatch directly.

rbarry82: The gist is that `@patch` only applies to the method it patches. In the case of `setUp`, it patches _only_ `setUp`, not each subsequent method.

So you can either also add the decorator to the entire class (just above `class TestDelayedPebbleReady`) *and* `setUp` (due to the object instantiation and realtime AST patching/parsing, it will apply to all methods _except_ `setUp` otherwise), or do it directly in `setUp` and stop in `tearDown`.

The latter is the "right" way to do it, and the rest of the mocks seem to have just sorta gotten lucky so far.

rbarry82: I just pushed an update which demonstrates it.

tests/unit/test_charm.py
```diff
@@ -150,3 +155,62 @@ def test__loki_config(self):
...
+        """Scenario: a regular relation is removed quickly, before pebble-ready fires."""
+        # WHEN relation-broken fires before pebble-ready
+        self.harness.remove_relation(self.log_rel_id)
+
+        # THEN app status is "Waiting" before pebble-ready
+        self.assertIsInstance(self.harness.charm.unit.status, WaitingStatus)
+
+        # AND app status is "Active" after pebble-ready
+        self.harness.container_pebble_ready("loki")
+        self.assertIsInstance(self.harness.charm.unit.status, ActiveStatus)
```
Abuelodelanada, line 216: I like these tests!

Abuelodelanada (dismissed):

tests/unit/helpers.py
```diff
@@ -4,7 +4,34 @@
...
+from typing import Callable, Tuple
+from unittest.mock import patch
+
+
+def tautology(*args, **kwargs) -> bool:
+    return True
+
+
+def patch_network_get(private_address="10.1.157.116") -> Callable:
+    def network_get(*args, **kwargs) -> dict:
```
balbirthomas, line 16: Nice !. Sagacious use of nested functions.

balbirthomas (approved):

simskij (dismissed):

simskij (approved):

Abuelodelanada (approved):

PR #148, opened 2022-04-19
FIX #107 - Now we are using our statically compiled Promtail binary

Abuelodelanada:
## Issue

See #107 

## Solution

Now we are using our statically compilled promtail.

## Context

--

## Testing Instructions

1. Deploy this charm
2. Implement `LogProxyConsumer` in a consumer charm, for instance: https://github.com/jnsgruk/zinc-k8s-operator
3. Deploy consumer charm
4. Relate Loki with consumer charm
5. Verify that Promtail is running in consumer charm

PR #158, opened 2022-04-28
Support both AMD64 and ARM64 Promtail binary URLs

balbirthomas:
## Issue
`LogProxyConsumer` currently only supports AMD64 architectures. Hence it can not install and use Promtail on other architectures such as ARM64. The purpose of this PR is to support both AMD64 and ARM64 architectures and ensure additional architectures can be easily added in the future.

## Solution
The following changes were made as part of this refactor.
    
- The relation data structure with promtail information set by LokiPushApiProvider has been extended so that it includes
   + information for multiple architectures
   + URL and filename of Promtail for each architecture
   + sha256 sum of Promtail zip for each architecture
    + sha256 sum of Promtail binary for each architecture
- All hardcoding of filename and paths for Promtail binaries were removed. The filenames are extracted from relation data provided by the LokiPushApiProvider, in a architecture dependent way. The LogProxyConsumer selects the correct Promtail filename based on its architecture from the choices offered to by the LokiPushApiProvider.
- The _setup_promtail() method has been made idempotent and the check if _is_promtail_installed() has been moved into this method. Hence _setup_promtail() only installs promtail if it has not already been installed.

## Context
Explained in "Issue" above.

## Testing Instructions
Just deploy Loki and Log proxy consumer and check that logs from promtail are forwarded.

## Release Notes
`LokiPushApiProvider` and `LogProxyConsumer` objects now support AMD64 and ARM64 architectures for Promtail.

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -319,7 +319,7 @@ resources:
   promtail-bin:
       type: file
       description: Promtail binary for logging
-      filename: promtail-linux-amd64
+      filename: promtail-linux
 ```
 
 Which would then allow operators to deploy the charm this way:
@@ -438,6 +438,7 @@ key.
 import json
 import logging
 import os
+import platform
 from collections import OrderedDict
 from copy import deepcopy
 from gzip import GzipFile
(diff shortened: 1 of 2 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -481,24 +482,29 @@ def _alert_rules_error(self, event):
...
+        "filename": "promtail-linux-amd64",
+        "zipsha": "978391a174e71cfef444ab9dc012f95d5d7eae0d682eaf1da2ea18f793452031",
+        "binsha": "00ed6a4b899698abc97d471c483a6a7e7c95e761714f872eb8d6ffd45f3d32e6",
+    },
+    "arm64": {
+        "filename": "prometail-linux-arm64",
+        "zipsha": "89f66a5b0daf47a1d7cb4d7db1a82278976384c7f60138b68abb12c72e332ce2",
+        "binsha": "b1590454a6bf5a6d6188ad54060cd377b447e6dbfda6d1376561ab066731634d",
+    },
+}
```
Abuelodelanada, line 501: @balbirthomas I think we should close [this PR](https://github.com/canonical/loki-k8s-operator/pull/148) and implement changes made there here.

For instance, using our own statically compiled Promtail binaries: https://github.com/canonical/loki-k8s-operator/releases/tag/promtail-v2.5.0

Abuelodelanada (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
balbirthomas: Ok. Will do it. I think it should be straight forward - just changing the base URL, filenames and shasums. Will test and then push tomorrow.

Abuelodelanada (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -481,22 +482,24 @@ def _alert_rules_error(self, event):
...
+# update all sha256 sums in PROMTAIL_BINARIES. To support a new architecture
+# you only need to add a new key value pair for the architecture in PROMTAIL_BINARIES.
+PROMTAIL_VERSION = "v2.5.0"
+PROMTAIL_BINARIES = {
+    "amd64": {
+        "filename": "promtail-static-amd64",
+        "zipsha": "543e333b0184e14015a42c3c9e9e66d2464aaa66eca48b29e185a6a18f67ab6d",
+        "binsha": "17e2e271e65f793a9fbe81eab887b941e9d680abe82d5a0602888c50f5e0cac9",
+    },
+}
```
dstathis, line 496: I thought the point of this PR was to support arm64?

balbirthomas: yes but at present this is not arm64 v2.5.0 binary available in our repo. It is a simple matter to add one more key value pair to this dictionary when we upload those binaries.

balbirthomas: As I understood this refactor was to bring in "support" for multiple architectures. We can actually add additional architectures trivially subsequently.

dstathis (approved):

PR #161, opened 2022-05-04
Move container operations to charm code instead of LokiPushApiProvider

### LINKED ISSUE #146: Loki Charm library should not handle updating alert rules in workload container

rbarry82:
## Issue
Previously, `LokiPushApiProvider` was passing a `Container` object into the library itself, which did a bunch of manipulation of the running workload. While this makes sense for `LogProxyConsumer` as a more-or-less sidecar, it's otherwise an inversion of the workflow we expect for libraries -- they should handle relation data, and communicate to the "real" workload that some information has been updated, which the workload can fetch, which keeps the libraries reasonably agnostic.

Closes #146 

## Solution
Move the container logic into `charm.py`. In the process, rework parts of the integration tests to get rid of more mocking.

Change the signature of `LokiPushApiRulesChanged` events to instead serialize the same information as a `RelationEvent` (it is not used at all from the charm for now, anyway).

Deleting/rewriting/checking alert rules now happens from the workload itself, which allows it to directly manage its blocked/active status without type checking events.

Pull out `tautology` (it was literally more characters as a function definition and/or import than `lambda x: True`, and did the same thing).

Pull out Tenacity. A unit test with a "TODO" test case doesn't need to be there, and scaling testing happens in integration tests  now.

## Testing Instructions
Clone the PR. Make sure it doesn't explode.

## Release Notes
Move container operations to charm code instead of LokiPushApiProvider


Abuelodelanada (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1225,8 +1229,17 @@ def __init__(
...
     def _on_lifecycle_event(self, _):
         # Upgrade event or other charm-level event
+        should_update = False
         for relation in self._charm.model.relations[self._relation_name]:
-            self._process_logging_relation_changed(relation)
+            # Don't accidentally flip a True result back.
+            should_update = should_update or self._process_logging_relation_changed(relation)
+        if should_update:
+            # We don't have an event, so build it up by hand
+            first_rel = self._charm.model.relations[self._relation_name][0]
```
sed-i, line 1238: `first_rel` is just a dummy for building the event?
If it's not a dummy, why [0] is sufficient?
Add comment about it?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1271,33 +1295,51 @@ def _on_logging_relation_departed(self, event: RelationDepartedEvent):
...
-    def _process_logging_relation_changed(self, relation: Relation):
+    def _should_update_alert_rules(self, relation) -> bool:
+        """Determine whether alert rules should be regenerated.
+
+        If there are alert rules in the relation data bag, tell the charm
+        whether or not to regenerate them based on the boolean returned here.
+        """
+        if relation.data.get(relation.app).get("alert_rules"):
+            return True
+        return False
```
sed-i, line 1313:
```suggestion
        return "alert_rules" in relation.data.get(relation.app)
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1271,33 +1295,51 @@ def _on_logging_relation_departed(self, event: RelationDepartedEvent):
...
-        self._check_alert_rules()
+        self.on.loki_push_api_alert_rules_changed.emit(
+            relation=event.relation,
+            relation_id=event.relation.id,
+            app=self._charm.app,
+            unit=self._charm.unit,
+        )
 
-    def _process_logging_relation_changed(self, relation: Relation):
+    def _should_update_alert_rules(self, relation) -> bool:
```
sed-i, line 1305: The function name is a bit misleading: it will continue to return True even after alert rules have been regenerated.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1271,33 +1295,51 @@ def _on_logging_relation_departed(self, event: RelationDepartedEvent):
...
+        the relation, let the caller know.
         Besides Loki generates alert rules files based what
         consumer charms forwards,
 
         Args:
             relation: the `Relation` instance to update.
+
+        Returns:
+            A boolean indicating whether an event should be emitted so we
+            only emit one on lifecycle events
```
sed-i, line 1331: "event should be emitted ..." is a separate (external) concern. Move the comment there?

tests/unit/test_charm.py
```diff
@@ -322,20 +403,20 @@ class TestDelayedPebbleReady(unittest.TestCase):
...
         self.check_alert_rules_patcher.start()
-        self.harness = Harness(LokiOperatorCharm)
 
-        # GIVEN this unit is the leader
+        self.harness = Harness(LokiOperatorCharm)
+        self.addCleanup(self.harness.cleanup)
         self.harness.set_leader(True)
+        self.harness.begin_with_initial_hooks()
 
         # AND a "logging" app joins with several units
```
sed-i, line 416: The "AND" doesn't make sense here after removing the "GIVEN" above.

Change to..?
```diff
-# AND a "logging" app joins with several units
+# GIVEN a "logging" app joins with several units
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1271,33 +1295,51 @@ def _on_logging_relation_departed(self, event: RelationDepartedEvent):
...
+        """Determine whether alert rules should be regenerated.
+
+        If there are alert rules in the relation data bag, tell the charm
+        whether or not to regenerate them based on the boolean returned here.
+        """
+        if relation.data.get(relation.app).get("alert_rules"):
+            return True
+        return False
+
+    def _process_logging_relation_changed(self, relation: Relation) -> bool:
```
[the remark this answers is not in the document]
rbarry82: It is not, as above.

src/charm.py
```diff
@@ -253,6 +253,102 @@ def _loki_config(self) -> dict:
         )
         return yaml.safe_load(config)
 
+    def _loki_push_api_alert_rules_changed(self, _: LokiPushApiAlertRulesChanged) -> None:
+        """Perform all operations needed to keep alert rules in the right status."""
+        if self._ensure_alert_rules_path():
+            self._regenerate_alert_rules()
+
+            # Don't try to configure if checking the rules left us in BlockedStatus
+            if isinstance(self.unit.status, ActiveStatus):
```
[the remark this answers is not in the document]
rbarry82: The lifecycle event and/or relation events in the library will emit a `LokiPushApiAlertRulesChanged` event, which will be obseved here.

`_regenerate_alert_rules` will run first, and if there are no errors, it is [set back to `ActiveStatus`](https://github.com/canonical/loki-k8s-operator/pull/161/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR347). Then we hit this line, and everything looks ok with the rules, so we move on to `_configure`

src/charm.py
```diff
@@ -253,6 +253,102 @@ def _loki_config(self) -> dict:
...
+
+    def _remove_alert_rules_files(self) -> None:
+        """Remove alert rules files from workload container."""
+        if not self._container.can_connect():
+            logger.debug("Cannot connect to container to remove alert rule files!")
+            return
+
+        files = self._container.list_files(self._rules_dir)
+        for f in files:
+            self._container.remove_path(f.path)
```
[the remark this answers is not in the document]
rbarry82: They can be, but are not. As above, [this logic is unchanged, only moved out of the library](https://github.com/canonical/loki-k8s-operator/pull/161/files#diff-ad7051b5ca68befcdb77d638fcfa686fa902ae8089c8e72cea3c222df01a4597L1408).

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1125,21 +1116,52 @@ class LokiPushApiEndpointJoined(EventBase):
...
 
 class LokiPushApiAlertRulesChanged(EventBase):
-    """Event emitted if there is an error with Alert Rules."""
+    """Event emitted if there is a change in the alert rules."""
 
-    def __init__(self, handle, error: bool = False, message: str = ""):
+    def __init__(self, handle, relation, relation_id, app=None, unit=None):
+        """Pretend we are almost like a RelationEvent.
+
+        Fields to serialize:
```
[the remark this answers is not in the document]
rbarry82: This guidance to users is part of the global docstring [here](https://github.com/canonical/loki-k8s-operator/pull/161/files#diff-ad7051b5ca68befcdb77d638fcfa686fa902ae8089c8e72cea3c222df01a4597R136). Unless we expect them to be instantiating this event, they will not see this docstring anyway.

Abuelodelanada (dismissed):

dstathis (approved):

src/charm.py
```diff
@@ -255,6 +255,102 @@ def _loki_config(self) -> dict:
...
+
+        for identifier, alert_rules in self.loki_provider.alerts.items():
+            rules = yaml.dump({"groups": alert_rules["groups"]})
+            file_mappings["{}_alert.rules".format(identifier)] = rules
+
+        if self._container.can_connect():
+            for filename, content in file_mappings.items():
+                path = os.path.join(self._rules_dir, filename)
+                self._container.push(path, content, make_dirs=True)
+        logger.debug("Saved alert rules to disk")
```
sed-i, line 307: We could make this more useful:
```suggestion
        logger.debug("Saved %s alert rules to disk: %s", len(file_mappings), self._rules_dir)
```

src/charm.py
```diff
@@ -255,6 +255,102 @@ def _loki_config(self) -> dict:
...
+
+    def _remove_alert_rules_files(self) -> None:
+        """Remove alert rules files from workload container."""
+        if not self._container.can_connect():
+            logger.debug("Cannot connect to container to remove alert rule files!")
```
sed-i, line 312: Should this be a `warning`?

src/charm.py
```diff
@@ -255,6 +255,102 @@ def _loki_config(self) -> dict:
...
+            return
+
+        files = self._container.list_files(self._rules_dir)
+        for f in files:
+            self._container.remove_path(f.path)
+
+    def _check_alert_rules(self):
```
sed-i, line 319: Is there a sensible way of having most of this function inside [`class LokiServer`](https://github.com/canonical/loki-k8s-operator/blob/main/src/loki_server.py)?

sed-i (approved):

Issue #118, opened 2022-03-23
Explicitly document need to ensure logs are topology labeled


balbirthomas:
### Bug Description

When a relation is added between `LokiPushApiProvider` and `LokiPushApiConsumer` the provider sends the Loki push API endpoint URL to the consumer. Hence the consumer can forward logs directly to this endpoint. However if the consumer chooses to do so using a custom client rather than the provided `promtail` resource then Juju topology labels are not automatically injected into the logs and the consumer needs to add topology labels to each log explicitly. 

At present the Loki documentation does not make this explicit. This is a gotcha that needs to be made explicit and prominent in the docs. Perhaps it will be good to have an example on how to generate the Juju topology labels also.

### To Reproduce

NA

### Environment

NA

### Relevant log output

```shell
NA
```

rbarry82:
It is always possible for logs to be pushed directly to Loki, though, regardless of the relation. The k8s service can be looked up, among other means.

Considering the renaming of cross-model-relations and the intrinsic guards around permissions and privacy from outside any given model, it's unlikely that any sort of reverse mapping from "IP address -> topology" exists, not least of which because overlapping CIDRs in different clusters or substrates which are only externally visible through some DNAT/SNAT means that "10.10.10.1" may not be unique in any way, nor externally resolvable.

Do we really think we need to document "hey, our Loki charm is still Loki"? I mean this as a serious question. The documentation is intended for using as part of the charm ecosystem, and usage above and beyond that is pretty much "go read the upstream docs". We could certainly point there.

It's worth noting that `cos-configuration` similarly has no meaningful way to map dashboards/alerts back to automatically labeled metrics/logs/traces or any other telemetry. Writing alert rules/dashboards which can still target data with sufficient cardinality even in the absence of Juju Topology (since "contributed" alerts from `cos-config` cannot know what the topology of any given outside Cassandra will be, for example), and if we're documenting it there as a "how-to", we can/should point there instead.

This comes back to my earlier suggestion for using `cos-configuration` in the integration tests rather than keeping a second tester charm which only has "bad" rules. It is not useful in that sense, but using a lookaside repo for specific positive/negative integration testing could serve as "living documentation/reference" for "here is how you can use these without the safety net relations+topology gives you"

balbirthomas:
Just trying to clarify, are you suggesting that  

1. We must *not* advice Loki charm users that by pushing "logs directly to loki" they will loose the "continuity of logs" that Juju Topology provides ? 
[...]

PR #165, opened 2022-05-09
Add section to the docs about topology labels

dstathis:
## Issue
#118 

## Solution
Add documentation.

## Testing Instructions
None! Just give it a read.

## Release Notes
Added documentation.

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -208,6 +208,12 @@ self.framework.observe(
 
 The consumer charm can then choose to update its configuration in both situations.
 
+Note that when LokiPushApiConsumer is used, logs will not have any labels automatically added. In
+order to better integrate with the Canonical Observability Stack, you may want to configure your
+software to add Juju topology labels. The
+[observability-libs](https://charmhub.io/observability-libs) library can be used to get topology
+labels in charm code.
+
 ## LogProxyConsumer Library Usage
 
 Let's say that we have a workload charm that produces logs and we need to send those logs to a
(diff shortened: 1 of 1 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -218,6 +218,12 @@ def __init__(self, *args):
 
 The consumer charm can then choose to update its configuration in both situations.
 
+Note that when LokiPushApiConsumer is used, logs will not have any labels automatically added. In
+order to better integrate with the Canonical Observability Stack, you may want to configure your
```
sed-i, line 222:
```suggestion
Note that LokiPushApiConsumer does not add any labels automatically on its own. In
order to better integrate with the Canonical Observability Stack, you may want to configure your
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -218,6 +218,12 @@ def __init__(self, *args):
...
+software to add Juju topology labels. The
```
sed-i, line 223:
```suggestion
charm to add Juju topology labels. The
```

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -218,6 +218,12 @@ def __init__(self, *args):
...
+[observability-libs](https://charmhub.io/observability-libs) library can be used to get topology
+labels in charm code.
```
sed-i, line 225: I think a concrete example here could go a long way.

dstathis: As discussed in our meeting, an example could become out of date and there is a link to the docs here.

Abuelodelanada (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i:
```suggestion
[observability-libs](https://charmhub.io/observability-libs) library can be used to get topology
labels in charm code. See :func:`LogProxyConsumer._scrape_configs` for an example of how
to do this with promtail.
```

sed-i (dismissed):

rbarry82 (approved):

Abuelodelanada (approved):

PR #166, opened 2022-05-10
Build up log-proxy-tester, add continuity integration tests

### LINKED ISSUE #136: LokiPushApi related upgrade event integration tests is missing essential checks
### Bug Description

At present there are three integration tests in Loki in the file `tests/test_upgrade_charm.py` (one of which keeps failing due to a known issue). These tests are

1. Deploy Loki from charmhub then refresh it from a locally built charm and check Loki is active
2. Deploy Alertmanager and Grafana from charmhub and add a relation with the Loki Deployed above in step 1.  Refresh Loki again using the same locally built charm and check Loki is active.
3. Scale the deployed Loki (from step 1 and 2) to have more than one unit and refresh it again using the same locally built charm and check Loki is active.

Shouldn't the integration tests for Loki's upgrade event handler be testing use cases such as 

1. Loki continues to aggregate Logs from the same clients it was related to prior to upgrade.
2. Continuity of Logs (labels) is not broken after upgrade.
3. Upgrading Loki-Tester allows for the injection of new alert rules.
4. Upgrading Loki-Tester allows for the removal of old alert rules.

There certainly may be many more use cases. However the above observations do make the case that the current upgrade integration tests are not only redundant (because all of them only test if Loki is active after upgrade) but also they do not to test the key aspects of the upgrade event handling process that are vital. In view of the above it is proposed to replace these three tests with *at least* one new tests. The proposed new test is 

- Deploy Loki from charmhub, Loki-Tester (locally built) and relate them.
- Check Loki aggregates logs and alerts fire (obviously also check Loki is active before doing this).
- Upgrade Loki with locally built charm.
- Check Loki still aggregates logs and there is continuity of Loki-Tester logs (and of course check Loki is active before doing this).
- Check Alerts still fire.

In addition the test may be extended to check alert rules can be added and removed on upgrade as follows

- Deploy Loki (from anywhere), Loki-Tester (locally built) and relate them (this can be a continuation of the test described above).
- Ensure Loki Tester is instrumented with one alert rule and submits logs to make this alert rule fire
- Check Loki aggregates logs and alerts rules fire
- Instrument Loki Tester to add a new alert rule and remove the old one
- Upgrade Loki Tester
- Instrument Loki Tester to push logs that would have resulted in both alert rules firing.
- Check that only the new alert rule is firing and not the old one.

### To Reproduce

NA

### Environment

NA

### Relevant log output

```shell
NA
```

### Additional context

NA

rbarry82:
## Issue
Build up integration tests for `LogProxyConsumer` to also exercise syslog functionality (and `LogProxyConsumer` which sends both file _and_ syslog).

Add integration tests to ensure continuity of logs and that logs make it to all Lokis.

Fix multi-loki LogProxyConsumer logging on the way

Closes #136 

## Solution
Integration tests pretty much speak for themselves.

`promtail` may encounter a delay between `relation-*` events and an endpoint actually being set on the Loki side, so also check in `relation_changed`, and ensure that multiple Loki endpoints are covered

## Release Notes
Expand integration tests for `LogProxyConsumer` and continuity of logs across upgrades.


tests/integration/loki-tester/src/charm.py
```diff
@@ -47,22 +47,24 @@ def _setup_logging(self, handlers: dict = None) -> None:
...
             # Give each handler a "name" property which matches so we can find it
-            v.name = k
+            v["handler"].name = k
+            v["handler"].setLevel(v["level"])
+
+        handlers = {v["handler"].name: v["handler"] for v in handlers_init.values()}
 
         # Check against logger.Manager and exclude "useless" values like logging.Placeholder
-        existing_handlers: dict[str, logging.Handler] = {v.name: v for v in logger.manager.loggerDict.values() if type(v) == logging.Handler}  # type: ignore
+        existing_handlers: dict[str, logging.Handler] = {k: v for v in logger.manager.loggerDict.items() if not isinstance(v, logging.PlaceHolder) and "loki" not in k}  # type: ignore
```
sed-i, line 67: Is it supposed to be `for k, v in logger.manager...`?

tests/integration/test_applications_send_logs.py
```diff
@@ -0,0 +1,127 @@
...
+            log_proxy_tester_charm,
+            resources=tester_resources,
+            application_name="log-proxy-tester-syslog",
+            config={"syslog": "true", "file_forwarding": "false"},
+        ),
+    )
+    await ops_test.model.wait_for_idle(apps=app_names, status="active")
+
+    for t in tester_app_names:
+        await ops_test.model.add_relation(loki_app_name, t)
```
sed-i, line 63:
```python
gather(
    *[ops_test.model.add_relation(loki_app_name, t) for t in tester_app_names]
)
```

tests/integration/test_applications_send_logs.py
```diff
@@ -0,0 +1,127 @@
...
+    await ops_test.model.wait_for_idle(apps=app_names, status="active")
+
+    # This is silly, but we need to actually wait to get log data
+    time.sleep(30)
```
sed-i, line 67: Maybe we could `block_until` (with timeout) on a concrete log count ([http api](https://grafana.com/docs/loki/latest/api/#examples))?

tests/integration/test_applications_send_logs.py
```diff
@@ -0,0 +1,127 @@
...
+    for query in tester_apps.values():
+        logs = await loki_api_query(ops_test, loki_app_name, query)
+        assert len(logs["values"]) > 0
+
+
+@pytest.mark.abort_on_fail
+async def test_scale_up_also_gets_logs(ops_test):
+    await ops_test.model.applications[loki_app_name].scale(scale=3)
+    await ops_test.model.wait_for_idle(
+        apps=[loki_app_name], status="active", timeout=600, wait_for_exact_units=0
```
sed-i, line 78: `wait_for_exact_units=3`?

tests/integration/test_applications_send_logs.py
```diff
@@ -0,0 +1,127 @@
...
+    for t in tester_app_names:
+        await ops_test.model.add_relation(loki_app_name, t)
+    await ops_test.model.wait_for_idle(apps=app_names, status="active")
+
+    # This is silly, but we need to actually wait to get log data
+    time.sleep(30)
+
+    for query in tester_apps.values():
+        logs = await loki_api_query(ops_test, loki_app_name, query)
+        assert len(logs["values"]) > 0
```
sed-i, line 71: Use `all` as below?

tests/integration/test_applications_send_logs.py
```diff
@@ -0,0 +1,127 @@
...
+        "log-error", message="Error logged!"
+    )
+
+    for query in tester_apps.values():
+        logs_per_unit = await asyncio.gather(
+            loki_api_query(ops_test, loki_app_name, query, unit_num=0),
+            loki_api_query(ops_test, loki_app_name, query, unit_num=1),
+            loki_api_query(ops_test, loki_app_name, query, unit_num=2),
+        )
+        assert all(len(logs["values"]) > 0 for logs in logs_per_unit)
```
sed-i, line 127: Seems like the `time.sleep(30)` should be removed, otherwise this will always pass.
I.e. to check persistence across upgrade, we'd need to assert against the log content before any additions are made.

tests/integration/test_log_proxy.py
```diff
@@ -57,3 +54,103 @@ async def test_alert_rules_do_fire(ops_test, loki_charm, log_proxy_tester_charm)
         for key in ["juju_application", "juju_model", "juju_model_uuid"]
         for alert in alerts
     )
+    await ops_test.model.applications[tester_app_name].remove()
```
sed-i, line 57: Wasn't immediately clear to me why there's a `remove` here. Maybe add a comment before this line, e.g.
```python
# Cleanup before next test
```

tests/integration/test_log_proxy.py
```diff
@@ -57,3 +54,103 @@ async def test_alert_rules_do_fire(ops_test, loki_charm, log_proxy_tester_charm)
...
+        resources=tester_resources,
+        application_name=tester_app_name,
+    )
+    await ops_test.model.block_until(
+        lambda: (
+            len(ops_test.model.applications[loki_app_name].units) > 0
+            and len(ops_test.model.applications[tester_app_name].units) > 0
+        )
+    )
+    await ops_test.model.add_relation(loki_app_name, tester_app_name)
```
sed-i, line 80: The `block_until` and the `add_relation` could go into a `gather`.

tests/integration/test_log_proxy.py
```diff
@@ -57,3 +54,103 @@ async def test_alert_rules_do_fire(ops_test, loki_charm, log_proxy_tester_charm)
...
+    assert len(logs["values"]) > 0
+
+
+@pytest.mark.abort_on_fail
+async def test_logproxy_logs_to_file_and_syslog(ops_test, log_proxy_tester_charm):
+    """Make sure Loki endpoints propagate on scaling."""
+    loki_app_name = "loki"
+    tester_app_name = "log-proxy-tester-both"
+
+    await ops_test.model.deploy(
```
sed-i, line 129: Seems like a `remove` is missing before this line.

tests/integration/test_applications_send_logs.py
rbarry82: This patch literally uses the HTTP API.

No, this is not an option. The issue is that Loki endpoints are sometimes (frequently) sent _after_ `promtail` events fire, and I wanted to avoid modifying the library again so we could reconcile `LIBPATCH`. After more testing, emitting a `PromtailDigestError` in the case where a relation exists but there is no endpoint data and allowing the charm (or at least the tester charm) to go into BlockedStatus is unavailable.

tests/integration/test_log_proxy.py
rbarry82: It is not. Different application names are used every time for this reason.

tests/integration/test_applications_send_logs.py
rbarry82: The point of the test here is "after upgrade, do we still get logs which match the same labels?" The sleep has nothing to do with that, really, except for the `endpoints` timing above.

I'll add a count which can be compared before and after though.

tests/integration/log-proxy-tester/src/charm.py
```diff
@@ -23,13 +23,24 @@ class LogProxyTesterCharm(CharmBase):
...
+            log_files=self._log_files if self._enable_file_forwarding else [],
         )
         self.framework.observe(
             self._log_proxy.on.promtail_digest_error,
             self._promtail_error,
         )
+        self.framework.observe(
+            self._log_proxy.on.log_proxy_endpoint_joined,
+            self._promtail_success,
+        )
```
sed-i, line 43: Iiuc, `log_proxy_endpoint_joined` would be emitted in tandem with `relation-changed`.
If so,
1. Do we still need `log_proxy_endpoint_joined`?
2. If yes, then are we implicitly relying on `relation-changed` to fire before `log_proxy_endpoint_joined`?

tests/integration/test_applications_send_logs.py
```diff
@@ -0,0 +1,139 @@
...
+    await asyncio.gather(
+        *[ops_test.model.add_relation(loki_app_name, t) for t in tester_app_names]
+    )
+    await ops_test.model.wait_for_idle(apps=app_names, status="active")
+
+    for query in tester_apps.values():
+        logs_per_unit = await asyncio.gather(
+            loki_api_query(ops_test, loki_app_name, query, unit_num=0),
+            loki_api_query(ops_test, loki_app_name, query, unit_num=1),
+            loki_api_query(ops_test, loki_app_name, query, unit_num=2),
```
sed-i, line 70: Seems like this deserves a dedicated helper function.

tests/integration/log-proxy-tester/src/charm.py
rbarry82: No. `log_proxy_endpoint_joined` is emitted _once promtail has a valid configuration_ and is running.

`relation-changed` can easily be emitted when `endpoints` change or `promtail_binary_url` is set, neither of which is a sufficient condition on its own to actually start/configure promtail.

This is not necessarily comitant with any given relation event (promtail may still be being fetched, `endpoints` may not be available for some reason, or other), and this patch overtly removes any implicit assumptions about what will or will not happen first, since relying on any `relation-` event to signal that the charm is "ready" does not necessarily indicate that `promtail` has a valid configuration, is fetched, or is running yet.

`log_proxy_endpoint_joined` is still needed, since it is essentially "`promtail` has been started successfully". It is the only signal to the consumer charm that it is _actually_ sending logs.

tests/integration/test_applications_send_logs.py
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
```
Abuelodelanada, line 2: 2022 ;-)

tests/integration/test_applications_send_logs.py
```diff
@@ -0,0 +1,178 @@
...
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+resources = {"loki-image": METADATA["resources"]["loki-image"]["upstream-source"]}
+tester_resources = {
+    "workload-image": oci_image(
+        "./tests/integration/log-proxy-tester/metadata.yaml", "workload-image"
+    )
+}
+
+
+# WARNING: the Loki query API is very finicky about qhoting. It MUST be surrounded
```
Abuelodelanada, line 24: qhoting -> quoting

Abuelodelanada (dismissed):

tests/integration/test_applications_send_logs.py
```diff
@@ -0,0 +1,178 @@
...
+    await ops_test.model.wait_for_idle(apps=app_names, status="active")
+
+    await asyncio.gather(
+        *[ops_test.model.add_relation(loki_app_name, t) for t in tester_app_names]
+    )
+    await ops_test.model.wait_for_idle(apps=app_names, status="active")
+
+    for query in tester_apps.values():
+        logs_per_unit = await asyncio.gather(
+            loki_api_query(ops_test, loki_app_name, query, unit_num=0),
```
sed-i, line 73: Were `unit_num=1` and `unit_num=2` intentionally removed?
If yes, then I would certainly appreciate a code comment about it.

simskij (approved):

src/charm.py
```diff
@@ -239,16 +239,27 @@ def _loki_config(self) -> dict:
                 kvstore:
                   store: inmemory
 
+            storage_config:
+              boltdb:
+                directory: /loki/boltdb-shipper-active
```
sed-i, line 244: A previous commit changed `boltdb-shipper` to `boltdb`.
I wonder why the path has `shipper` in it?

sed-i (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1879,15 +1879,23 @@ def _on_relation_changed(self, event: RelationEvent) -> None:
...
+
             new_config = self._promtail_config
             if new_config != self._current_config:
                 self._container.push(
                     WORKLOAD_CONFIG_PATH, yaml.safe_dump(new_config), make_dirs=True
                 )
+
+            # Loki may send endpoints late. Don't necessarily start, there may be
+            # no clients
+            if new_config["clients"]:
```
Abuelodelanada, line 1894:
```python
if new_config.get("clients", None):
```

Abuelodelanada (approved):

src/charm.py
rbarry82: Because of the volume mount in `metadata.yaml`

PR #179, opened 2022-06-11
Use cos-tool to transform/validate, use JujuTopology from observability libs

rbarry82:
## Issue
Kind of a lot.

A mechanism was needed to validate alert rules _before_ Loki died if they were loaded, and `JujuTopology` is now in `observability-libs`, so removal of the custom code was slated.

## Solution
Add `cos-tool`, which does what `promql-transform` did for PromQL label tranformation, plus also handlingalert rule validation. Add tests for all of that, and an event fired back if an alert rule is invalid, which is then excluded from being loaded into Prometheus

Replace the "custom" `JujuTopology` with the one from `observability-libs`, which works well, but required a really annoying amount of updates to tests, especially unit tests to avoid `InvalidUUID` exceptions blowing them up.

## Testing Instructions
Test, run, be joyful.

## Release Notes
Use `JujuTopology` from observability-libs, validate alert rules, inject label matchers without `%%juju_topology%%`

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -441,18 +441,21 @@ import os
 import platform
 import re
 import socket
+import subprocess
+import tempfile
 import typing
-from collections import OrderedDict
+import uuid
 from copy import deepcopy
 from gzip import GzipFile
 from hashlib import sha256
 from io import BytesIO
 from pathlib import Path
-from typing import Dict, List, Optional, Union, cast
+from typing import Dict, List, Optional, Tuple, Union, cast
 from urllib import request
 from urllib.error import HTTPError
 
 import yaml
+from charms.observability_libs.v0.juju_topology import JujuTopology
 from ops.charm import (
     CharmBase,
     HookEvent,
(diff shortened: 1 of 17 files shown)


tests/integration/log-proxy-tester/lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,288 @@
+# Copyright 2022 Canonical Ltd.
```
sed-i, line 1: Was this committed by mistake? Should probably be copied into as part of the test.

tests/integration/loki-tester/lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,288 @@
+# Copyright 2022 Canonical Ltd.
```
sed-i, line 1: Was this committed by mistake? Should probably be copied into as part of the test.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -441,18 +441,21 @@ def _alert_rules_error(self, event):
...
 from hashlib import sha256
 from io import BytesIO
 from pathlib import Path
-from typing import Dict, List, Optional, Union, cast
+from typing import Dict, List, Optional, Tuple, Union, cast
 from urllib import request
 from urllib.error import HTTPError
 
 import yaml
+from charms.observability_libs.v0.juju_topology import JujuTopology
```
sed-i, line 458: Add to provider docstring that would need to fetch-lib?

simskij (dismissed):
Once Leon's comments are resolved, I'm happy to have this merged.

tests/integration/log-proxy-tester/lib/charms/observability_libs/v0/juju_topology.py
rbarry82: No, it was not. Since this lib isn't actually part of Loki, and a scenario of "this lib has changes in this patch, so make sure it's copied in for the tests" isn't a common use case, updates as-needed (when/if JujuTopology updates) would be expected/normal.

This also provides an additional validation point for when changes in `loki_push_api` may break/require updates to `JujuTopology` in users, which would get masked if we copied it.

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: Done

sed-i (dismissed):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -950,9 +760,15 @@ def _from_file(self, root_path: Path, file_path: Path) -> List[dict]:
...
+                        alert_rule["labels"].update(self.topology.label_matcher_dict)
                         # insert juju topology filters into a prometheus alert rule
-                        alert_rule["expr"] = self.topology.render(alert_rule["expr"])
+                        # logql doesn't like empty matchers, so add a job matcher which hits
+                        # any string as a "wildcard" which the topology labels will
+                        # filter down
+                        alert_rule["expr"] = self.tool.inject_label_matchers(
+                            re.sub(r"%%juju_topology%%(,?)", r'job=".+"\1', alert_rule["expr"]),
+                            self.topology.label_matcher_dict,
+                        )
```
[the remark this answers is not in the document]
rbarry82: It's required here (which is how it wound in Prometheus to stay in sync).

The difference in this case is that LogQL _cannot_ have empty matchers, or the expression will fail to parse, which is why a "wildcard" `job=".+"` matcher is injected. If a comma which was present is omitted, in, say, `{%%juju_topology%%, juju_unit="foo"}`, it would become `{job=".+" juju_unit="foo"}`, which wouldn't parse.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -441,18 +446,21 @@ def _alert_rules_error(self, event):
...
 from hashlib import sha256
 from io import BytesIO
 from pathlib import Path
-from typing import Dict, List, Optional, Union, cast
+from typing import Dict, List, Optional, Tuple, Union, cast
 from urllib import request
 from urllib.error import HTTPError
 
 import yaml
+from charms.observability_libs.v0.juju_topology import JujuTopology
```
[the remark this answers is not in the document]
rbarry82: With a patch in progress for grafana-agent to pass alerts to Loki also, the same dedupe task as Prometheus will apply here. 

The plan is to bump once the corresponding patch hits.

Abuelodelanada (dismissed):
I would update [`juju_topology.py` lib](https://github.com/canonical/observability-libs/blob/main/lib/charms/observability_libs/v0/juju_topology.py#L128) to its latest version. This way we can get rid of all `JujuTopology.is_valid_uuid` mocks.

Besides there is a [PR in ops](https://github.com/canonical/operator/pull/782) that solves the UUID issue and is ready to be merged

rbarry82:
> I would update [`juju_topology.py` lib](https://github.com/canonical/observability-libs/blob/main/lib/charms/observability_libs/v0/juju_topology.py#L128) to its latest version. This way we can get rid of all `JujuTopology.is_valid_uuid` mocks.
> 
> Besides there is a [PR in ops](https://github.com/canonical/operator/pull/782) that solves the UUID issue and is ready to be merged

Done :D

sed-i (approved):

PR #181, opened 2022-07-04
WIP: metrics based alert rules

Abuelodelanada:
## Issue

This PR adds metric based alert rules

### DIFF
--- a/src/prometheus_alert_rules/loki_process_too_many_restarts.rule
+++ b/src/prometheus_alert_rules/loki_process_too_many_restarts.rule
@@ -1,0 +1,9 @@
+# Based on https://awesome-prometheus-alerts.grep.to/rules.html#loki
+alert: LokiProcessTooManyRestarts
+expr: changes(process_start_time_seconds[15m]) > 2
+for: 0m
+labels:
+  severity: warning
+annotations:
+  summary: Loki process too many restarts (instance {{ $labels.instance }})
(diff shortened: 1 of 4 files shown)


sed-i (approved):

dstathis (approved):

balbirthomas (approved):

src/prometheus_alert_rules/loki_process_too_many_restarts.rule
```diff
@@ -0,0 +1,10 @@
+# Based on https://awesome-prometheus-alerts.grep.to/rules.html#loki
+# Given the way Loki charm behaves seems reasonable to warn if Loki restarts 5 times in 15 minutes
```
sed-i, line 2:
```suggestion
# The Loki charm may restart the workload on config-changed event, which means there would be
# two restarts just from deploying the charm.
# Changing the config soon after deploy (e.g. set resource limits) would cause another restart.
# Using 5 as the alert threshold to take into account restart (2 starts) + immediate config change (1 more start).
```

However, seems like `process_start_time_seconds` is not a restart count per se. Does "5" mean loki spent 5 seconds starting?

Abuelodelanada: Nop, `process_start_time_seconds` it is not a count per se. What transforms it in a count is the `changes` function:

```
changes(process_start_time_seconds[15m]) > 5
```

The result of this would be:

imagen

sed-i (approved):

rbarry82 (approved):

PR #188, opened 2022-07-20
Relabel structured syslog data

rbarry82:
## Issue
Promtail is able to parse structured syslog data into labels, but those labels are prefixed with underscores and may not be meaningful to users who are attempting to query or integrate them into dashboards.

## Solution
Transform `__syslog_message_foo` and `__syslog_message_sd_{id}_{key}` into more 

## Context
The fixed labels are culled from [here](https://github.com/grafana/loki/blob/b24a1430434770a5c4eddc19bfe8e1eddf498a15/pkg/promtail/targets/syslog/syslogtarget.go#L180)

## Testing Instructions
Run the integration tests, or connect a syslog target.

## Release Notes
Relabel structured syslog data

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -484,7 +484,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 11
+LIBPATCH = 12
 
 logger = logging.getLogger(__name__)
 
@@ -2123,6 +2123,14 @@ class LogProxyConsumer(ConsumerBase):
 
         # Syslog config
         if self._is_syslog:
+            relabel_mappings = [
+                "severity",
+                "facility",
+                "hostname",
+                "app_name",
+                "proc_id",
+                "msg_id",
+            ]
             syslog_labels = common_labels.copy()
             syslog_labels.update({"job": "{}_syslog".format(job_name)})
             syslog_config = {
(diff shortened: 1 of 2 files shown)


balbirthomas (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -2132,9 +2140,17 @@ def _scrape_configs(self) -> dict:
...
                 },
+                "relabel_configs": [
+                    {"source_labels": ["__syslog_message_{}".format(val)], "target_label": val}
+                    for val in relabel_mappings
+                ]
+                + [{"action": "labelmap", "regex": "__syslog_message_sd_(.+)"}],
             }
             scrape_configs.append(syslog_config)  # type: ignore
 
+        import pprint
```
Abuelodelanada: Is `pprint` such a large module that is preferable to import it here rather than at the beginning of the file?

Abuelodelanada (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: Gone now, but it's one of those "printf debugging" things in my workflow. If something is unclear, I import on the spot to log it so it's easier to remember to remove it later

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -2132,6 +2140,11 @@ def _scrape_configs(self) -> dict:
                     "label_structured_data": True,
                     "labels": syslog_labels,
                 },
+                "relabel_configs": [
```
sed-i, line 2143: We should link here to the reference you have [linked to in the pr description](https://github.com/grafana/loki/blob/b24a1430434770a5c4eddc19bfe8e1eddf498a15/pkg/promtail/targets/syslog/syslogtarget.go#L180), for future reference.

Issue #186, opened 2022-07-19
Collect logs from multiple containers in the same pod


arturo-seijas:
### Enhancement Proposal

There's no straightforward way to set up the loki `LogProxyConsumer` to work with multiple containers in the same charm. Instantiating it more than once is not supported either:
`RuntimeError: two objects claiming to be IndicoOperatorCharm/LogProxyConsumer[log-proxy] have been created`

It is my understanding that this is a very common use case and it should be supported.

simskij:
This is something we need to support. Maybe not in the GA, but at least in a point release soon thereafter. I envision that we'll keep the single instantiation and add a way for authors to pass multiple containers with multiple, different, log configs.

PR #191, opened 2022-07-21
Allow an instance of LogProxyConsumer per container [do not merge: experiment]

sed-i:
## Issue
Currently it is impossible to create multiple instances of LogProxyConsumer for the same relation.

## Solution
Allow an instance per relation per container.

- [x] add `container_name` to the `Object key`.
- [ ] rework relation data schema
- [ ] tests

Addresses #186.

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1371,8 +1371,10 @@ class ConsumerBase(Object):
         relation_name: str = DEFAULT_RELATION_NAME,
         alert_rules_path: str = DEFAULT_ALERT_RULES_RELATIVE_PATH,
         recursive: bool = False,
+        *,
+        key: Optional[str] = None,
     ):
-        super().__init__(charm, relation_name)
+        super().__init__(charm, key or relation_name)
         self._charm = charm
         self._relation_name = relation_name
         self.topology = JujuTopology.from_charm(charm)
@@ -1685,11 +1687,15 @@ class LogProxyConsumer(ConsumerBase):
         container_name: str = "",
         promtail_resource_name: Optional[str] = None,
     ):
-        super().__init__(charm, relation_name, alert_rules_path, recursive)
         self._charm = charm
         self._relation_name = relation_name
         self._container = self._get_container(container_name)
         self._container_name = self._get_container_name(container_name)
+
+        key = "{}_{}".format(relation_name, self._container_name)
+        logger.info("KEY %s", key)
+        super().__init__(charm, relation_name, alert_rules_path, recursive, key=key)
+
         self._log_files = log_files or []
         self._syslog_port = syslog_port
         self._is_syslog = enable_syslog
--- a/tests/unit/test_log_proxy_consumer.py
+++ b/tests/unit/test_log_proxy_consumer.py
@@ -75,6 +75,8 @@ class ConsumerCharm(CharmBase):
         containers:
           loki:
             resource: loki-image
+          loki2:
+            resource: loki-image
           promtail:
             resource: promtail-image
         requires:
@@ -90,6 +92,9 @@ class ConsumerCharm(CharmBase):
         self._stored.set_default(invalid_events=0)
         self.log_proxy = LogProxyConsumer(
             charm=self, container_name="loki", log_files=LOG_FILES, enable_syslog=True
+        )
+        self.log_proxy2 = LogProxyConsumer(
+            charm=self, container_name="loki2", log_files=LOG_FILES, enable_syslog=True
         )
 
         self.framework.observe(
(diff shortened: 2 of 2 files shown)


arturo-seijas:
This looks like a perfectly valid approach but I'm wondering if changing the constructor arguments as follows would be a better option:
```
    charm=self, log_files=[
        {
            container: "sample-container",
            files: ["file.log""]
        }
    ]
)
```
IMO this would provide a better development experience and will be aligned with the approach already in use for the Prometheus `MetricsEndpointProvider`:
```
self._metrics_endpoint = MetricsEndpointProvider(
    self, jobs=[
        {
            "metrics_path": "/my-metrics-path",
            "static_configs": [{"targets": ["*:9113", "*:9102"]}]
        }
    ]
)
```
Unfortunately, this would be a breaking change.

What are your thoughts on this?

sed-i:
By design, `class LogProxyConsumer` handles a single container. You can search occurrences of the following to get an idea how deeply that assumption is there:
- `self._container`
- `MultipleContainersFoundError`

so some major refactoring would be needed for this to work.

UX wise I imagine it's better for the user to not need to know the schema of a new dict we'd need to invent for this purpose.

---

Regardless of the above, there is another implication of sending logs from multiple containers: you won't be able to differentiate between them. Currently the promtool config file looks like this:
```bash
$ juju ssh --container workload flog/0 cat /etc/promtail/promtail_config.yaml
```
```yaml
clients:
- url: http://loki-0.loki-endpoints.m11.svc.cluster.local:3100/loki/api/v1/push
positions:
  filename: /opt/promtail/positions.yaml
scrape_configs:
- job_name: system
  static_configs:
  - labels:
      __path__: /bin/fake.log
      job: juju_m11_dfde3eb2_flog
      juju_application: flog
      juju_charm: flog-k8s
      juju_model: m11
      juju_model_uuid: dfde3eb2-4dda-4673-886e-c0fc705e0045
      juju_unit: flog/0
    targets:
    - localhost
server:
  grpc_listen_port: 9095
  http_listen_port: 9080
```

We may need to have `job_name` or an additional label reflect the container name.
WDYT? @rbarry82 @Abuelodelanada @dstathis @simskij

rbarry82:
Strictly speaking, `job` is per-pod, and per-container logs should use [`container_name`](https://github.com/grafana/loki/pull/229/files) to differentiate. This is standard practice with is used all over.

Similarly, the "standard" `promtail` deployment is a [`DaemonSet` which reads everything from `/var/lib/docker/containers` and `/var/log` (also including `/var/log/containers`)](https://github.com/grafana/loki/blob/b7a4035ba69f0ded089e523bf8370e0aa42cafba/production/helm/templates/promtail/daemonset.yaml#L46). We aren't going to do _that_, but the "canonical" way to do this is not by making any changes to the library this way at all, but instead documenting that a volume should be shared across all containers in the pod (maybe even `/var/log`), and pointing `promtail` at **that** instead.

If we are not going to go that route (documenting how to do it as part of the charm metadata using a shared volume, which wouldn't strictly include `container_name` anyway because it can't be inferred without the "raw" container info from `/var/lib/[containerd|docker|whatever]`, but we'd still get all of the topology), then we should explicitly add a `container_name` key

Abuelodelanada:
When we thought the implementation of `LogProxyConsumer` our use case was: "_one charm, one workload_"

This will certainly need a major refactor in the lib.

You are right @sed-i, we will need a new `container` label. 

An approach to LogProxyConsumer's `__init__` could be:


```python
        self._logging = LogProxyConsumer(
            self,
            relation_name="logging",
            containers = [
                {"container_name": "workload1", log_files=["file1", "file2"]},
                {"container_name": "workload2", log_files=["file1", "file2"]},
            ],
        )
```

But we need to think it better... I'm not so sure it is as good as I would like to be

sed-i:
> a volume should be shared across all containers in the pod (maybe even `/var/log`), and pointing `promtail` at **that** instead

I recall we always had reservations for using a shared volume (would it be shared across units?). @rbarry82 probably remembers better what those were.

As for multiplicity, @rbarry82 suggests "one job per pod", which probably means a single `LogProxyConsumer` for all the containers

One `LogProxyConsumer` instances per container seems easier = tempting.

rbarry82:
Well, it's a little different.

A shared volume (not across units, which is really terrible anyway -- shared filesystems are a danger zone without sanlock to orchestrate or using an actual clustered FS such as gluster, GFS2, lustre, cephfs, etc) across containers inside a pod is a common scenario, which is what I meant. But that's a "documentation-based" solution. It's one which more-or-less matches upstream best practice (to run as few promtails as possible and make communication with Loki efficient), but we aren't restricted to it.

To be particular, when I say one job per pod, I mean in the sense that Loki's default label (similar to Prometheus) is `job="foo"`. So we have one `job` as-in label per pod (`job="some-pod-logs"`), and each container has additional `container_name="foo"`, `container_name="bar"` label to add specificity. We should not re-use `job_name` as the identifier, because it's not the standard (`container_name` is), but even if there were not one, `job_name` is too similar to `job` and could be confusing.

simskij:
Closing this without merging. The recommendation for now is to use a shared volume. Long-term, we might consider adding support for targeting multiple containers in the same constructor call.

PR #199, opened 2022-09-12
Upkeep

dstathis:
## Issue
Upkeep

## Solution
Add Luca to CODEOWNERS.
Use binary to set workload verison.

## Testing Instructions
Deploy and check workload version is set correctly.

## Release Notes
loki-k8s sets workload version more reliably


Abuelodelanada (commented):
What is the benefit of executing the binary instead of querying the API?
Should not we keep these server responsibilities in `loki_server.py`?

rbarry82 (approved):
Also allows reporting the version even if the workload is down

simskij (approved):

PR #203, opened 2022-10-14
Also update the grafana source configuration when ingress is configured

rbarry82:
## Issue
Update the grafana source address when ingress is configured, which will change the subpath routing

## Solution
The same as `update_endpoint` for Loki

## Testing Instructions
This is already tested in the Grafana repo itself, and testing here would introduce a circular dependency

## Release Notes
Also update the grafana source configuration when ingress is configured

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -148,6 +148,7 @@ class LokiOperatorCharm(CharmBase):
     def _on_ingress_changed(self, _):
         self._configure()
         self.loki_provider.update_endpoint(url=self._external_url)
+        self.grafana_source_provider.update_source(source_url=self._external_url)
 
     def _on_logging_relation_changed(self, event):
         # If there is a change in logging relation, let's update Loki endpoint


src/charm.py
```diff
@@ -148,6 +148,7 @@ def _on_alertmanager_change(self, _):
     def _on_ingress_changed(self, _):
         self._configure()
         self.loki_provider.update_endpoint(url=self._external_url)
+        self.grafana_source_provider.update_source(source_url=self._external_url)
```
sed-i, line 151: This is certainly an improvement but iiuc this will become sufficient only when https://github.com/canonical/traefik-k8s-operator/issues/78 is fixed?

sed-i (approved):

src/charm.py
rbarry82: I mean, not really, for the most part. It doesn't _actually_ matter to Grafana whether the ingress is currently listening or whether it takes a while. The Grafana process (not even charm, actual workload) itself will handle marking it as available whenever it is.

The crucial part here is in making sure that the datasource configuration is updated, whether or not it's truly available yet or has to wait a bit.

PR #207, opened 2022-10-18
Refresh alert rules on pebble ready to account for potential pod churn

rbarry82:
## Issue
In case of a host restart, it's possible that no alert events will actually be fired, as there are no relation changes.

## Solution
Pre-emptively attempt to fetch and generate alert rules `on_loki_pebble_ready`. Even in the case of empty alerts, this is harmless.

## Testing Instructions
Deploy Loki. Relate something with alerts. Reboot the host. Ensure they are still present.

## Release Notes
Refresh alert rules on pebble ready to account for potential pod churn

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -135,6 +135,7 @@ class LokiOperatorCharm(CharmBase):
         self._configure()
 
     def _on_loki_pebble_ready(self, _):
+        self._ensure_alert_rules_path()
         self._regenerate_alert_rules()
         self._configure()
         version = self._loki_version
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -135,6 +135,8 @@ def _on_upgrade_charm(self, _):
         self._configure()
 
     def _on_loki_pebble_ready(self, _):
+        self._ensure_alert_rules_path()
+        self._regenerate_alert_rules()
```
Abuelodelanada, line 139: `_ensure_alert_rules_path` return a boolean.

Why do not we do the same we are doing in `_loki_push_api_alert_rules_changed` method?

```suggestion
        if self._ensure_alert_rules_path():
            self._regenerate_alert_rules()
```

Abuelodelanada (approved):

PR #208, opened 2022-10-18
Fix api call

### LINKED ISSUE #205: _check_alert_rules fails if a path_prefix_is set in the loki config
### Bug Description

_check_alert_rules fails if a path_prefix_is set in the loki config

### To Reproduce

deploy loki, relate to traefik, then relate to any workload charm.

### Environment

Ubuntu 22.04
Microk8s 1.25.2
Juju 2.9.35

### Relevant log output

```shell
-
```

### Additional context

_No response_

dstathis:
## Issue
Fixes #205

## Solution
Insert path_prefix in to the api call url.

## Testing Instructions
* Deploy Loki
* Relate to Traefik
* Relate to a logging client
* See if Loki is up

## Release Notes
Fixed a bug where Loki-k8s fails when used with Traefik-k8s

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -20,7 +20,7 @@ import textwrap
 from typing import Optional
 from urllib import request
 from urllib.error import HTTPError, URLError
-from urllib.parse import urljoin, urlparse, urlunparse
+from urllib.parse import urlparse
 
 import yaml
 from charms.alertmanager_k8s.v0.alertmanager_dispatch import AlertmanagerConsumer
@@ -88,7 +88,7 @@ class LokiOperatorCharm(CharmBase):
         )
 
         self.ingress_per_unit = IngressPerUnitRequirer(
-            self, relation_name="ingress", port=self._port
+            self, relation_name="ingress", port=self._port, strip_prefix=True
         )
         self.framework.observe(self.ingress_per_unit.on.ready_for_unit, self._on_ingress_changed)
         self.framework.observe(self.ingress_per_unit.on.revoked_for_unit, self._on_ingress_changed)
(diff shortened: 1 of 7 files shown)


simskij (approved):

src/charm.py
```diff
@@ -99,7 +99,15 @@ def __init__(self, *args):
             source_type="loki",
             source_url=self._external_url,
         )
-        scrape_jobs = [{"static_configs": [{"targets": [f"*:{self._port}"]}]}]
+
+        scrape_url = urlparse(self._external_url)
+        logger.debug("Using url {} for Prometheus scrape".format(urlunparse(scrape_url)))
```
Abuelodelanada, line 104: I [would remove](https://juju.is/docs/sdk/logging) the `.format(...`

```suggestion
        logger.debug("Using url %s for Prometheus scrape", urlunparse(scrape_url))
```

src/charm.py
```diff
@@ -419,9 +427,12 @@ def _remove_alert_rules_files(self) -> None:
 
     def _check_alert_rules(self):
         """Check alert rules using Loki API."""
-        url = "http://127.0.0.1:{}/loki/api/v1/rules".format(self.loki_provider.port)
+        path_prefix = urlparse(self._external_url).path
+        url_path = os.path.join(path_prefix, "loki/api/v1/rules")
+        url = urljoin("http://127.0.0.1:{}".format(self._port), url_path)
         req = request.Request(url)
         try:
+            logger.debug("Checking alert rules via {}.".format(req.full_url))
```
Abuelodelanada, line 435: The same here:


```suggestion
            logger.debug("Checking alert rules via %s.", req.full_url)
```

Abuelodelanada (approved):

rbarry82 (approved):

src/charm.py
rbarry82: Logger already does it, basically. It's an extremely marginal performance difference, but `logger.level("....")` has an implicit `printf`-style interpolation.

src/charm.py
```diff
@@ -99,7 +99,15 @@ def __init__(self, *args):
             source_type="loki",
             source_url=self._external_url,
         )
-        scrape_jobs = [{"static_configs": [{"targets": [f"*:{self._port}"]}]}]
+
+        scrape_url = urlparse(self._external_url)
+        logger.debug("Using url {} for Prometheus scrape".format(urlunparse(scrape_url)))
+        scrape_jobs = [
+            {
+                "metrics_path": "{}/{}".format(scrape_url.path, "metrics"),
```
sed-i, line 107: Because trailing slash is not automatically cleaned up
```
>>> "{}/{}".format(urlparse("http://a.b").path, "metrics")
'/metrics'
>>> "{}/{}".format(urlparse("http://a.b/").path, "metrics")
'//metrics'
```
I would
```suggestion
                "metrics_path": "{}/metrics".format(scrape_url.path.rstrip("/")),
```

src/charm.py
```diff
@@ -99,7 +99,15 @@ def __init__(self, *args):
             source_type="loki",
             source_url=self._external_url,
         )
-        scrape_jobs = [{"static_configs": [{"targets": [f"*:{self._port}"]}]}]
```
sed-i, line 102: This is star notation because we need it to automatically expand into all units.
This goes into app data so only the leader does that. The current change would only include the leader unit. I.e. currently the proposed change is buggy.

I.e. you must use star notation.

Add utest as well?

sed-i: This is also related to line 119
```
port=parsed_external_url.port or self._port
```
which is also buggy. It should probably be
```
port=parsed_external_url.port or 80
```

sed-i (dismissed):
We must continue using star notation.

src/charm.py
rbarry82: Yes, that's true. In this case, it's kind of a micro-optimization which is just down to style. The additional ops from "double interpolating" isn't going to be the end of the world

src/charm.py
rbarry82: Other than the fact Loki does not, in fact, listen on `80` by default. It's more of a concern that `urlparse.ParseResult` doesn't have a `port` field at all, and that IPU returns a "bare" URL which doesn't include a port anyway.

Probably, it will be necessary to either make `_external_url` a little more complex to actually figure out what's happening instead of just blindly returning `ingress_url`, to _assume_ that if `ingress_url` is valid (maybe an attribute on the class, or changing the type of `_external_url`, which would be more invasive) the port is 80, or other.

rbarry82:
We are consistently getting locked up trying to scale down at all. I've been trying to sift through Traefik's GH CI, but does that test work locally? If it doesn't, we may want to debate disabling it while we sort out the root cause as long as we think it's just CI

src/charm.py
dstathis: Closing this comment in the interest of releasing on time.

src/charm.py
dstathis: Fixed.

PR #215, opened 2022-11-15
Make loki_push_api machine charm compatible

dstathis:
## Issue
LokiPushApiProvider has an attribute `.container` which is k8s specific.

## Solution
Default the attribute to None.

## Context
It looks like Loki and Grafana Agent both do not use this attribute. Perhaps it would be safe to just remove it?

## Testing Instructions
Not really sure as I can't find anything that uses this.

## Release Notes
* Improvements toward making the LokiPushApiProvider machine charm compatible.

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1097,7 +1097,6 @@ class LokiPushApiProvider(Object):
         self._relation_name = relation_name
         self._tool = CosTool(self)
         self.port = int(port)
-
         self.scheme = scheme
         self.address = address
         self.path = path
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1098,13 +1098,6 @@ class LokiPushApiProvider(Object):
         self._tool = CosTool(self)
         self.port = int(port)
 
-        # This needs to return none if it does not exist for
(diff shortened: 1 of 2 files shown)


rbarry82 (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1097,7 +1097,14 @@ def __init__(
...
         self._tool = CosTool(self)
         self.port = int(port)
-        self.container = self._charm._container
+
+        # This needs to return none if it does not exist for
+        # compatibility with machine charms. Also, this should
+        # be removed in v1 of the library as we should not
+        # assume any charm attributes exist except those
+        # guaranteed by ops.
+        self.container = getattr(self._charm, "_container", None)
```
sed-i, line 1106: More likely than not  this is dead code and can probably be removed.
`container` comes from the charm. Why would any charm code access its container via the lib?
I vote for removal.

dstathis: I agree but I was hesitant due to the lack of an underscore. I think you are right though. I doubt anything is using this.

sed-i (approved):

sed-i:
Ok the merge button was disabled because of the unresolved conversation.
I included the changes of this PR in #218 an apparently CI passes without the no_implicit_optional addition.

dstathis:
Closing as this was rolled in to another PR.

PR #218, opened 2022-11-22
fetch-lib

sed-i:
## Issue
Outdated libs.

## Solution
- Fetch lib
- Run no_implicit_optional on loki files
- Drop self.container prop


Abuelodelanada (approved):

sed-i:
CI fails on the good old scaledown test. Merging red.

PR #232, opened 2023-01-13
fix ServicePort definition

lucabello:


rbarry82 (approved):
+1 (once CI passes -- I re-triggered it)

PR #243, opened 2023-01-30
Build against 20.04 and a compatible version of glibc

rbarry82:
## Issue
Also, bump golang to `1.19`.

`ubuntu-latest` is now 22.04, and is not ABI compatible with the tester images being used, so an appropriate version of `glibc` can't be found by the linker.

We could actually step around this and statically link `glibc` in also, but it would require modifying the `Makefile` for `grafana/loki` as part of an action, because it doesn't look for any meaningful env vars we could substitute.

With a `grafana-agent` machine snap/charm, do we actually have a need for an in-container rebuilt promtail which can read journalctl?

For that matter, is there a reason we're rebuilding this at all instead of grabbing the prebuilt binary from upstream? @simskij ?

### DIFF
--- a/.github/workflows/build-promtail-release.yaml
+++ b/.github/workflows/build-promtail-release.yaml
@@ -14,7 +14,7 @@ jobs:
 
   build-dynamic:
     name: Build dynamically-linked Promtail
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


rbarry82:
[This passes now](https://github.com/canonical/loki-k8s-operator/actions/runs/4049215963). I'm still not entirely convinced that we need it as a workflow anymore, though.

lucabello (approved):

PR #249, opened 2023-02-15
Type check the `log_files` arg

### LINKED ISSUE #226: Improve validation of parameters on LogProxyConsumer lib
### Enhancement Proposal

Passing a string instead of a list doesn't lead to any errors from the lib:
```
self.loki_push = LogProxyConsumer(self, log_files="my/log/file.log", ...)
```

sed-i:
## Issue
If a user mistakenly passed an str instead of a list for the `log_files` arg, the response is too silent and too late.

## Solution
Raise on type mismatch for `log_files`.

Fixes #226.

## Context
[Temptation](https://typeguard.readthedocs.io/en/latest/) was mild.

## Testing Instructions
Afaict, utest provides sufficient coverage.

## Release Notes
Type check the log_files arg.

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -483,7 +483,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 17
+LIBPATCH = 18
 
 logger = logging.getLogger(__name__)
 
@@ -1755,7 +1755,7 @@ class LogProxyConsumer(ConsumerBase):
     def __init__(
         self,
         charm,
-        log_files: Optional[list] = None,
+        log_files: Optional[Union[List[str], str]] = None,
         relation_name: str = DEFAULT_LOG_PROXY_RELATION_NAME,
         enable_syslog: bool = False,
         syslog_port: int = 1514,
@@ -1769,7 +1769,15 @@ class LogProxyConsumer(ConsumerBase):
         self._relation_name = relation_name
         self._container = self._get_container(container_name)
         self._container_name = self._get_container_name(container_name)
-        self._log_files = log_files or []
+
+        if not log_files:
+            log_files = []
+        elif isinstance(log_files, str):
+            log_files = [log_files]
+        else:
+            log_files = list(log_files)
+        self._log_files = log_files
+
         self._syslog_port = syslog_port
         self._is_syslog = enable_syslog
         self.topology = JujuTopology.from_charm(charm)
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -1774,8 +1774,10 @@ class LogProxyConsumer(ConsumerBase):
             log_files = []
         elif isinstance(log_files, str):
             log_files = [log_files]
-        else:
-            log_files = list(log_files)
+        elif not isinstance(log_files, list) or not all(
+            map(lambda x: isinstance(x, str), log_files)
+        ):
+            raise TypeError("The 'log_files' argument must be a list of strings.")
         self._log_files = log_files
 
         self._syslog_port = syslog_port
--- a/lib/charms/grafana_k8s/v0/grafana_dashboard.py
+++ b/lib/charms/grafana_k8s/v0/grafana_dashboard.py
@@ -219,7 +219,7 @@ LIBAPI = 0
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 
-LIBPATCH = 26
+LIBPATCH = 28
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


sed-i:
Are we merging on red when it is only `test_rerelate` that fails?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1689,7 +1689,15 @@ def __init__(
         self._relation_name = relation_name
         self._container = self._get_container(container_name)
         self._container_name = self._get_container_name(container_name)
-        self._log_files = log_files or []
+
+        if log_files is None:
+            log_files = []
```
dstathis, line 1694: The only difference here is that with the new way, if the user passes something like `""` or `{}` we now raise and exception rather than silently converting to a list. Which behavior is preferred? I'm not sure.

tox.ini
```diff
@@ -94,7 +94,7 @@ commands =
 description = Run integration tests
 deps =
     aiohttp
-    juju
+    juju~=3.0.0
```
dstathis, line 97: Should this maybe be `juju~=3.0`?

dstathis (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i: Correct, anything false-y would have become `[]`.
- Seems odd though that for `log_files=""` it works but then `log_files="file"` is a TypeError.
- We could accept `Union[list, set, tuple]` but just one is simpler and list seems most appropriate.

tox.ini
sed-i: `juju<3.1` would also work. All 3 options are equivalent afaict.

rbarry82: They aren't _quite_ equivalent, because there is actually `3.0.4`. `< 3.1` is clearly enough "the highest possible 3.0 release"

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1689,7 +1689,15 @@ def __init__(
...
+        elif type(log_files) is not list:
+            raise TypeError(
+                f"The `log_files` argument must be a list but it is '{type(log_files)}'"
+            )
```
rbarry82, line 1698: This cannot be merged. Yes, we _should_ accept a list, and yes, accepting `Union[List[str], str]` may seem incoherent, but suddenly raising an exception in a library update which doesn't also increment `LIBAPI` with arguments which used to pass is an unacceptable workflow/regression, regardless of whether that behavior was intentional or not.

Hyrum's Law and so on. The fact that the user who reported this as not being enforced noticed _while_ doing this is a fair enough indication of this. Even while iterating over a string would end up with `/` as a log path, it remains.

`v1` probably isn't far away, at least.

Also, `if type(...)` is almost never the right choice over `isinstance`

lib/charms/loki_k8s/v0/loki_push_api.py
rbarry82: Silently converting to a list is preferable so we don't raise brand-new exceptions on a `fetch-lib` with code that used to work.

```python
if log_files and isinstance(log_files, str):
  log_files = [log_files]
elif not isinstance(log_files, (list, set, tuple)) and log_files is None:
  log_files = []
else:
  logger.error("Some message about how it's invalid, and that we're going to zero it out, but it won't work")
  log_files = []
```

We can still warn the author in a manner far less disruptive than raising exceptions for previously acceptable behavior and making them run it down (and potentially rewriting their unit tests before they can continue).

It's not as cut-and-dry, but we live with it until v1.

rbarry82 (changes requested):

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i: Done

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1769,7 +1769,15 @@ def __init__(
...
         self._container = self._get_container(container_name)
         self._container_name = self._get_container_name(container_name)
-        self._log_files = log_files or []
+
+        if not log_files:
+            log_files = []
+        elif isinstance(log_files, str):
+            log_files = [log_files]
+        else:
+            log_files = list(log_files)
```
rbarry82, line 1778: I wonder if, since the presumptive fall-through case here is that it's _already_ a list, if we're gaining anything and we don't want to type check that it's actually a list (after checking whether it's a string), otherwise fail and log an error

rbarry82: I'd probably disagree with "try to cast stuff and jam it into a list and assume that if no exceptions come up, it makes sense as unsanitized input". No reason not to just throw our hands up and return a meaningful message, lest some other random type make it through that without throwing an exception and we end up with a user who's very dependent on an accidental API which we could change.

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1769,7 +1769,17 @@ def __init__(
...
+
+        if not log_files:
+            log_files = []
+        elif isinstance(log_files, str):
+            log_files = [log_files]
+        elif not isinstance(log_files, list) or not all(
+            map(lambda x: isinstance(x, str), log_files)
+        ):
+            raise TypeError("The 'log_files' argument must be a list of strings.")
+        self._log_files = log_files
```
sed-i, line 1781: @rbarry82 Wdyt about this stricter check?

PR #253, opened 2023-03-08
Find the uk8s group

rbarry82:
## Issue
Don't always assume it's `microk8s`, it's not when strict.


sed-i:
Revision 66 of charm 'loki-k8s' released to edge (attaching resources: 'loki-image' r50)

PR #264, opened 2023-03-24
Add an arg to skip topology label injection for Loki alert rules

rbarry82:
## Issue
Without going as far as `RulesProvider` in `prometheus_scrape`, since we don't need to skip other data, add an optional arg to bypass appending Juju topology labels to alert rules.

Since `metadata` is still provided in the relation, the "other side" is able to establish an identifier for the filename anyway, and `apply_label_matchers` will effectively NOOP without labels, as there's nothing to inject into expressions.

See: [this cos-config issue](https://github.com/canonical/cos-configuration-k8s-operator/issues/44) for reference

## Release Notes
Add an arg to skip topology label injection for Loki alert rules

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -483,7 +483,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 17
+LIBPATCH = 18
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1450,6 +1450,7 @@ def __init__(
         relation_name: str = DEFAULT_RELATION_NAME,
         alert_rules_path: str = DEFAULT_ALERT_RULES_RELATIVE_PATH,
         recursive: bool = False,
+        skip_alert_topology_labeling: bool = False,
```
sed-i, line 1453: **question:** Should we name it similar to grafana, e.g. `inject_topology_labels=True`?

**suggestion:** Should probably document that it's not just alert labels but also `expr` qualifiers.

PR #272, opened 2023-04-27
use ruff instead of flake8

lucabello:
The **flake8** plugins we use are re-implemented in **ruff**, which is why they have been removed from the dependencies.

**isort** is enabled (and used when formatting) through the `I001` rule.

The "exclude list" has been updated to omit folders that **ruff** excludes by default.

Everything else is the same, the only difference being one new rule being ignored (which wasn't checked before anyway):
* `RET504: Unnecessary variable assignment before return statement`, which we do everywhere as it increases readability.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,9 +27,6 @@ per-file-ignores = {"tests/*" = ["D100","D101","D102","D103"]}
 [tool.ruff.pydocstyle]
 convention = "google"
 
-[tool.ruff.mccabe]
-max-complexity = 10
-
 # Static analysis tools configuration
 [tool.mypy]
 pretty = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,26 +13,22 @@ show_missing = true
 line-length = 99
 target-version = ["py38"]
 
-[tool.isort]
-profile = "black"
-
 # Linting tools configuration
-[tool.flake8]
-max-line-length = 99
-max-doc-length = 99
-max-complexity = 10
-exclude = [".git", "__pycache__", ".tox", "build", "dist", "*.egg_info", "venv"]
-select = ["E", "W", "F", "C", "N", "R", "D", "H"]
-# Ignore W503, E501 because using black creates errors with this
+[tool.ruff]
+line-length = 99
+extend-exclude = ["__pycache__", "*.egg_info"]
+select = ["E", "W", "F", "C", "N", "R", "D", "I001"]
+# Ignore E501 because using black creates errors with this
 # Ignore D107 Missing docstring in __init__
-ignore = ["W503", "E501", "D107"]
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -13,26 +13,22 @@ show_missing = true
...
-select = ["E", "W", "F", "C", "N", "R", "D", "H"]
-# Ignore W503, E501 because using black creates errors with this
+[tool.ruff]
+line-length = 99
+extend-exclude = ["__pycache__", "*.egg_info"]
+select = ["E", "W", "F", "C", "N", "R", "D", "I001"]
+# Ignore E501 because using black creates errors with this
 # Ignore D107 Missing docstring in __init__
-ignore = ["W503", "E501", "D107"]
+ignore = ["E501", "D107", "RET504"]
```
sed-i, line 23: BTW what's the difference between ignore and extend-ignore?

pyproject.toml
```diff
@@ -13,26 +13,22 @@ show_missing = true
...
-max-complexity = 10
-exclude = [".git", "__pycache__", ".tox", "build", "dist", "*.egg_info", "venv"]
-select = ["E", "W", "F", "C", "N", "R", "D", "H"]
-# Ignore W503, E501 because using black creates errors with this
+[tool.ruff]
+line-length = 99
+extend-exclude = ["__pycache__", "*.egg_info"]
+select = ["E", "W", "F", "C", "N", "R", "D", "I001"]
+# Ignore E501 because using black creates errors with this
 # Ignore D107 Missing docstring in __init__
```
sed-i, line 22: So the google style says the construction docstring can be either in init or the class level docstring.
Can we tell ruff to look for either?

tox.ini
```diff
@@ -31,29 +31,21 @@ passenv =
 description = Apply coding style standards to code
 deps =
     black
-    isort
+    ruff
```
sed-i, line 34: With ruff we still need black?
For example: https://beta.ruff.rs/docs/settings/#docstring-quotes

pyproject.toml
lucabello: `extend-ignore` technically add elements to the default list of ignores; however, **ruff** [deprecates it](https://beta.ruff.rs/docs/settings/#extend-ignore) in favor of `ignore`

pyproject.toml
lucabello: Not really, no; their implementation of "google convention" is relatively straightforward and not customizable :)

tox.ini
lucabello: Actually we most likely don't; I didn't think about it because I was keeping the changes as minimal as possible, but the `fmt` has been run by **ruff** only in all the PRs already; **black** seems to be redundant indeed!

PR #277, opened 2023-05-10
Doc update

sed-i:
## Issue
Intended usage of LokiPushApiConsumer is not clear enough.

## Solution
Improve doc.

## Release Notes
Doc update.


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1532,25 +1530,27 @@ def __init__(
...
 
-        The `LokiPushApiConsumer` object provides configurations to a Loki client charm.
-        A charm instantiating this object needs Loki information, for instance the
-        Loki API endpoint to push logs.
-        The `LokiPushApiConsumer` can be instantiated as follows:
+        The `LokiPushApiConsumer` object provides configurations to a Loki client charm, such as
+        the Loki API endpoint to push logs. It is intended for workloads that can speak
+        loki_push_api (https://grafana.com/docs/loki/latest/api/#push-log-entries-to-loki), such
+        as grafana-agent.
+        (If you only need to forward a few workload log files, then use LogProxyConsumer.)
```
sed-i, line 1537: @bencekov is this better?

bencekov: yes, I think it's good

bencekov (approved):
Thanks for the update. I think the purpose of the different classes is clearer now.

lib/charms/loki_k8s/v0/loki_push_api.py
Abuelodelanada:
```
It is intended for workloads that can speak loki_push_api (https://grafana.com/docs/loki/latest/api/#push-log-entries-to-loki), such as grafana-agent.
```

This is not 100% correct.

We use `LogProxyConsumer` in [Zinc charm](https://github.com/jnsgruk/zinc-k8s-operator/blob/main/src/charm.py#L53) because Zinc **cannot** speak loki_push_api by itself, and `LogProxyConsumer` provides that feature. After instrumenting `LogProxyConsumer` the charm is able to send logs to Loki or Grafana Agent.

Abuelodelanada (changes requested):

Abuelodelanada (approved):

PR #284, opened 2023-07-10
Implement TLS and enable it when a cert is available

Abuelodelanada:
## Issue

This PR implements TLS as part of this tracking issue https://github.com/canonical/cos-lite-bundle/issues/75

## How to test it

1. Deploy Loki
2. Deploy `self-signed-certificates`
3. Deploy `prometheus-k8s`
4. Relate Loki to `self-signed-certificates` and `prometheus-k8s`
5. You can check HTTP endpoint in Prometheus:
imagen

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -10,6 +10,7 @@ bases:
       channel: "20.04"
 parts:
   charm:
+    charm-binary-python-packages: [cryptography, jsonschema]
     build-packages: [git]
   cos-tool:
     plugin: dump
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -10,7 +10,6 @@ bases:
       channel: "20.04"
 parts:
   charm:
-    charm-binary-python-packages: [cryptography, jsonschema]
     build-packages: [git]
   cos-tool:
     plugin: dump
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -56,6 +56,16 @@ requires:
   ingress:
     interface: ingress_per_unit
     limit: 1 # Since this is the ingress there is no point in having more than one.
+  certificates:
+    interface: tls-certificates
+    limit: 1
+    description: |
+      Certificate and key files for the loki server.
+
+peers:
+  replicas:
+    interface: loki_replica
+    # assumed network type: private
 
 resources:
   loki-image:
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -65,7 +65,6 @@ requires:
 peers:
   replicas:
     interface: loki_replica
-    # assumed network type: private
 
 resources:
   loki-image:
--- a/src/charm.py
+++ b/src/charm.py
@@ -291,7 +291,13 @@ class LokiOperatorCharm(CharmBase):
         current_layer = self._container.get_plan()
         new_layer = self._build_pebble_layer
         restart = True if current_layer.services != new_layer.services else False
-        config = ConfigBuilder(self).build()
+
+        config = ConfigBuilder(
+            instance_addr=self.hostname,
+            alertmanager_url=self._alerting_config(),
+            external_url=self._external_url,
+            http_tls=self.server_cert.cert,
+        ).build()
 
         try:
             self._push_certs()
--- a/src/charm.py
+++ b/src/charm.py
@@ -251,7 +251,12 @@ class LokiOperatorCharm(CharmBase):
 
     @property
     def scrape_jobs(self) -> List[Dict[str, Any]]:
-        """Generate scrape jobs."""
+        """Generate scrape jobs.
+        
+        Note: We're generating a scrape job for the leader only because loki is not intended to
+        be scaled beyond one unit. If we wanted to scrape all potential units, we would need
+        to collect all the peer addresses over peer relation data.
+        """
         if not self.ingress_per_unit.url:
             job: Dict[str, Any] = {
                 "static_configs": [{"targets": [f"{self.hostname}:{self._port}"]}]
--- a/src/charm.py
+++ b/src/charm.py
@@ -74,8 +74,6 @@ class LokiOperatorCharm(CharmBase):
         # https://grafana.com/docs/loki/latest/operations/storage/filesystem/
         # https://grafana.com/docs/loki/latest/rules/#ruler-storage
         tenant_id = "fake"
-        self.loki_cert_path = CERT_FILE
-        self.loki_key_path = KEY_FILE
         self.rules_dir_tenant = os.path.join(RULES_DIR, tenant_id)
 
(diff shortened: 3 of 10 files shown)


charmcraft.yaml
```diff
@@ -10,6 +10,7 @@ bases:
       channel: "20.04"
 parts:
   charm:
+    charm-binary-python-packages: [cryptography, jsonschema]
```
dstathis, line 13: We should not be doing this in production code. Out charms are supposed to be built from source.

(This is the blocking comment)

metadata.yaml
```diff
@@ -56,6 +56,16 @@ requires:
...
+  certificates:
+    interface: tls-certificates
+    limit: 1
+    description: |
+      Certificate and key files for the loki server.
+
+peers:
+  replicas:
+    interface: loki_replica
+    # assumed network type: private
```
dstathis, line 68: I'm not sure I understand this comment.

src/config_builder.py
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+    @property
+    def _server(self) -> dict:
+        _server = {
+            "http_listen_address": "0.0.0.0",
+            "http_listen_port": self._charm._port,
+        }
+
+        if self._charm.server_cert.cert:
```
dstathis, line 90: I do not think this is a safe way to check if TLS should be enabled. What if the relation exists but the cert has not been generated yet? In that case we don't want the server running over HTTP while we wait.

I am going to open a PR adding [this method](https://github.com/canonical/observability-libs/compare/main...enabled_method) to cert_manager. Until then, I think you should do something similar.

src/config_builder.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Reference: https://grafana.com/docs/loki/latest/configuration/
+    """
+
+    _target: str = "all"
+    _auth_enabled: bool = False
+
+    def __init__(self, charm, *args):
+        """Init method."""
+        self._charm = charm
```
dstathis, line 23: Why add `*args` to the arguments? Do you expect anything else to be passed in?

dstathis (changes requested):

charmcraft.yaml
Abuelodelanada: You are right, I forget this

src/config_builder.py
Abuelodelanada: Sounds reasonable. 
That is a simple PR, let's wait for it to be merged, and then I'll use that method here.

src/charm.py
```diff
@@ -225,26 +258,18 @@ def _external_url(self) -> str:
         # are routable virtually exclusively inside the cluster (as they rely)
         # on the cluster's DNS service, while the ip address is _sometimes_
         # routable from the outside, e.g., when deploying on MicroK8s on Linux.
-        return f"http://{self.hostname}:{self._port}"
+        scheme = "https" if self.server_cert.cert else "http"
+        return f"{scheme}://{self.hostname}:{self._port}"
 
     @property
-    def scrape_jobs(self) -> list:
+    def scrape_jobs(self) -> List[Dict[str, Any]]:
```
sed-i, line 265: The new scrape job looks quite different:
- Iirc, there's a leader guard elsewhere surrounding this function call, so without the `*`-notation we'd only get this unit.
- If we do end up having ingress in place, we will have a path component and we will need to set `metrics_path` as before. Unless you refactor the charm to use `strip_prefix=True` in the ingress lib.
- `self.hostname` is `socket.getfqdn()`, so we're losing `self.ingress_per_unit.url` compared to before. When grafana agent pushes to loki. Perhaps add a comment that clearly states that we assume self-mon is always within the model?

src/config_builder.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Reference: https://grafana.com/docs/loki/latest/configuration/
+    """
+
+    _target: str = "all"
+    _auth_enabled: bool = False
+
+    def __init__(self, charm):
+        """Init method."""
+        self._charm = charm
```
sed-i, line 23: Should we replace it with c'tor args?
I'm inclined to having a dep-free config builder.

src/config_builder.py
```diff
@@ -0,0 +1,105 @@
...
+            "http_listen_address": "0.0.0.0",
+            "http_listen_port": self._charm._port,
+        }
+
+        if self._charm.server_cert.cert:
+            _server["http_tls_config"] = {
+                "cert_file": self._charm.loki_cert_path,  # HTTP server cert path.
+                "key_file": self._charm.loki_key_path,  # HTTP server key path.
+                "client_auth_type": "",  # HTTP TLS Client Auth type.
+                "client_ca_file": "",  # HTTP TLS Client CA path.
```
sed-i, line 95: Seems like this is mTLS.
Can't we omit them then?

src/config_builder.py
sed-i: I'd prefer something like
```python
config = ConfigBuilder(tls_enabled=is_tls_enabled(), ...).build()
```
or
```python
config = ConfigBuilder(...).set_http_tls({"cert_path"=..., "key_path"=...}).build()
```

src/config_builder.py
Abuelodelanada: c'tor args??

I also inclined to dep-free config builder.The problem here is that we need a lot of information from the charm to build the config, that why I'm injecting it.

src/charm.py
Abuelodelanada: Until we have OB031 approved, let's keep self.ingress_per_unit.url

src/config_builder.py
dstathis: Sounds good to me.

dstathis (approved):

src/charm.py
Abuelodelanada: Yes, you are right.

So, we have another issue here then, because when the relation to Traefik is established (`ingress_per_unit`), we set this scrape job and we are always scrapping the leader unit: 


```python
        return [
            {
                "metrics_path": f"{external_url.path}/metrics",
                "static_configs": [{"targets": [f"{external_url.hostname}:{external_url.port}"]}],
            }
        ]
```

For instance, without Traefik we have:

imagen

imagen


With Traefik:

imagen

imagen

sed-i: I imagine the peer data trick should do it.

src/charm.py
```diff
@@ -264,18 +290,18 @@ def _configure(self):  # noqa: C901
 
         current_layer = self._container.get_plan()
         new_layer = self._build_pebble_layer
+        restart = True if current_layer.services != new_layer.services else False
```
sed-i, line 293: Now we have `replan` so we don't need to track the `restart`.

sed-i (approved):

Issue #293, opened 2023-07-26
Add scheme to loki client


sed-i:
Currently the client code has `http` hard-coded:
https://github.com/canonical/loki-k8s-operator/blob/5a826f1d823b14934ac17046d8dee10b29e7c951/src/loki_server.py#L41

Ref: https://github.com/canonical/prometheus-k8s-operator/blob/d33f51f39d990de3b8dcb9436bf69291a8e8b891/src/prometheus_client.py#L20

Abuelodelanada:
We do not need TLS for this.
This is almost dead code. Right now we are using `self.base_url` to get the `build_info` through a request:

https://github.com/canonical/loki-k8s-operator/blob/5a826f1d823b14934ac17046d8dee10b29e7c951/src/loki_server.py#L44

And with that `build_info` we get Loki's `version`:

https://github.com/canonical/loki-k8s-operator/blob/5a826f1d823b14934ac17046d8dee10b29e7c951/src/loki_server.py#L62


We can get Loki's version using the binary. An example of this in [Grafana agent](https://github.com/canonical/grafana-agent-k8s-operator/blob/main/src/k8s_charm.py#L134):

```python
    def agent_version_output(self) -> str:
        """Runs `agent -version` and returns the output.

        Returns:
            Output of `agent -version`
        """
        version_output, _ = self._container.exec(["/bin/agent", "-version"]).wait_output()
        return version_output
```

Abuelodelanada:
This is definitely dead code, we are in fact using:


https://github.com/canonical/loki-k8s-operator/blob/main/src/charm.py#L509

```python
    @property
    def _loki_version(self) -> Optional[str]:
        """Returns the version of Loki.

        Returns:
            A string equal to the Loki version
        """
        if not self._container.can_connect():
            return None
        version_output, _ = self._container.exec(["/usr/bin/loki", "-version"]).wait_output()
        # Output looks like this:
        # loki, version 2.4.1 (branch: HEAD, ...
        result = re.search(r"version (\d*\.\d*\.\d*)", version_output)
        if result is None:
            return result
        return result.group(1)
```

PR #296, opened 2023-07-26
src/loki_server.py removed

Abuelodelanada:
## Issue

This PR fixes https://github.com/canonical/loki-k8s-operator/issues/293
`src/loki_server.py` removed since it is dead code.


sed-i (approved):
So it wasn't even in use!

PR #306, opened 2023-09-20
Charm logging

PietroPasotti:
charm instrumentation library using loki push api to forward charm logs (same logs that go to the juju debug-log) to a loki instance. 

Similar to what tempo's charm_tracing does, the library assumes the presence of a loki_push_api integration (that the charm might be using already to push workload logs).

image

This library exposes a decorator the user will put in front of a charm class. 
The user will pass to the decorator the name of a charm method that returns a list of loki endpoint urls to which the charm logs should be forwarded. Presumably, the user will get those endpoints from `LokiPushApiConsumer` (but in practice we don't care how the url is obtained).

### DIFF
--- a/lib/charms/loki_k8s/v0/charm_logging.py
+++ b/lib/charms/loki_k8s/v0/charm_logging.py
@@ -14,6 +14,8 @@ from ops import Framework, CharmBase
 if TYPE_CHECKING:
     from ops.model import _ModelBackend  # noqa
 
+import logging_loki
+
 
 # The unique Charmhub library identifier, never change it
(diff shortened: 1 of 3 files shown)


sed-i:
Would you mind updating the PR description with the mechanism? Just guessing:
- User needs to pass a logging.getLogger instance so this lib could tag a long
- The lib relies on the loki_push_api relation schema to obtain loki server url
- User adds a class level decorator to the charm, utilizing some ops stuff

simskij:
Closing this due to staleness. If this is still relevant, please reopen or create a new PR once it's ready for review.

PR #314, opened 2023-10-11
upkeep: use refactored ci and pyright

lucabello:


dstathis (approved):

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -866,20 +868,20 @@ def _from_dir(self, dir_path: Path, recursive: bool) -> List[dict]:
 
         return alert_groups
 
-    def add_path(self, path: str, *, recursive: bool = False):
+    def add_path(self, path_str: str, *, recursive: bool = False):
```
sed-i, line 871: Just curious what was wrong with `path`?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1885,7 +1889,7 @@ def _on_relation_departed(self, _: RelationEvent) -> None:
             self._container.stop(WORKLOAD_SERVICE_NAME)
         self.on.log_proxy_endpoint_departed.emit()
 
-    def _get_container(self, container_name: str = "") -> Container:
+    def _get_container(self, container_name: str = "") -> Container:  # pyright: ignore
```
sed-i, line 1892: Just curious why the ignore here?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1959,7 +1963,9 @@ def _add_pebble_layer(self, workload_binary_path: str) -> None:
                 }
             },
         }
-        self._container.add_layer(self._container_name, pebble_layer, combine=True)
+        self._container.add_layer(
+            self._container_name, pebble_layer, combine=True  # pyright: ignore
```
sed-i, line 1967: Just curious why the ignore here?

lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -1996,7 +2002,11 @@ def _push_binary_to_workload(self, binary_path: str, workload_binary_path: str)
         """
         with open(binary_path, "rb") as f:
             self._container.push(
-                workload_binary_path, f, permissions=0o755, encoding=None, make_dirs=True
+                workload_binary_path,
+                f,
+                permissions=0o755,
+                encoding=None,  # pyright: ignore
```
sed-i, line 2008: Do we need to raise an issue against ops?

src/charm.py
```diff
@@ -507,7 +507,7 @@ def _resource_reqs_from_config(self) -> ResourceRequirements:
         return adjust_resource_requirements(limits, requests, adhere_to_requests=True)
 
     def _on_k8s_patch_failed(self, event: K8sResourcePatchFailedEvent):
-        self.unit.status = BlockedStatus(event.message)
+        self.unit.status = BlockedStatus(str(event.message))
```
sed-i, line 510: Just curious why the ignore here?

lib/charms/loki_k8s/v0/loki_push_api.py
lucabello: It was being immediately rewritten with `path = Path(path)`, which the linter didn't like

src/charm.py
lucabello: There's no ignore here :) I guess it was meant for the other line

lib/charms/loki_k8s/v0/loki_push_api.py
lucabello: Might be a good idea!

lib/charms/loki_k8s/v0/loki_push_api.py
lucabello: The pebble layer should be of `LayerDict` type, but I wanted this PR to change as little as possible to avoid breaking anything

lib/charms/loki_k8s/v0/loki_push_api.py
lucabello: Because the function technically returns an `Optional[Container]`; however, changing the return type means adding an `if` EVERY time we call the `self._container` property, which is a lot; for the purpose of keeping the PR small, I just ignored it here

If we want to refactor this specifically, I'd suggest a new issue, because it's a bit of code which will need to be tested better

PR #324, opened 2023-11-22
Add config options for ingestion rate limits

sed-i:
## Issue
The charm does not expose config options for ingestion rate limits.

## Solution
Add config options for ingestion rate limits.

Fixes #323.

## Context
- https://github.com/canonical/cos-lite-bundle/pull/84

PR #325, opened 2023-11-22
Modify loki config to address HTTP 429 code (too many reqs)

### LINKED ISSUE #283: Configuration plans for Loki charm
### Enhancement Proposal

[Read this post](https://discourse.charmhub.io/t/configuration-plans-for-loki-charm/10991):

> The context is that after instrumenting my machine charm with grafana-agent and setting up a local COS-lite environment, I was able to see logs in Grafana and proceeded to setup a few panels that displayed logs for different sources - syslog, model logs, unit logs, etc. Mainly to observe what the difference was between them. That’s when I ran into a bit of an issue, my Grafana panels all timed out and displayed an error “too many outstanding requests”.
> 
> Further investigation landed me on a few [threads](https://github.com/grafana/loki/issues/8981) - [like](https://github.com/grafana/loki/issues/4613) - [this](https://github.com/grafana/loki/issues/5123) suggesting changes to Loki config.
>
> The [Loki charm ](https://charmhub.io/loki-k8s/configure) currently doesn’t offer much in the way of config so my question is how will users of the COS-lite stack be able to tune their deployment. My understanding is that sensible charm defaults should get one most of the way there but predictably there might be a need to configure parameters for large and small scale operations.

sed-i:
## Issue
Following up on #324, as the problem still persists.

- Getting HTTP codes 429 and others throughout the load test.
- Loki logs `2023-11-22T20:28:31.034Z [loki] level=warn ts=2023-11-22T20:28:30.939304741Z caller=cache.go:114 msg="fifocache config is deprecated. use embedded-cache instead"`

## Solution
- Enable embedded-cache.
- Additional tweaks, per refs inline.

With this change, at high load (`increase(loki_distributor_lines_received_total[1m])` of up to 320,000), all of the following issues are now gone:
- 429 / GET / loki_api_v1_query_range
- 499 / GET / loki_api_v1_query_range
- 500 / GET / loki_api_v1_query_range
- 503 / GET / ready
- 500 / POST / loki_api_v1_push
- cancel / gRPC / /logproto.Querier/Query
- cancel / gRPC / /schedulerpb.SchedulerForQuerier/QuerierLoop

Fixes #283.

## Context
- https://github.com/canonical/cos-lite-bundle/pull/84

## Testing Instructions
```bash
juju refresh --switch loki-k8s --channel=edge/325 loki  # Expires at 2023-12-30T00:00:00Z
```

### DIFF
--- a/src/config_builder.py
+++ b/src/config_builder.py
@@ -64,6 +64,8 @@ class ConfigBuilder:
             "limits_config": self._limits_config,
             "query_range": self._query_range,
             "chunk_store_config": self._chunk_store_config,
+            "frontend": self._frontend,
+            "querier": self._querier,
         }
 
     @property
@@ -154,6 +156,7 @@ class ConfigBuilder:
     def _query_range(self) -> dict:
         # Ref: https://grafana.com/docs/loki/latest/configure/#query_range
         return {
+            "parallelise_shardable_queries": False,
             "results_cache": {
                 "cache": {
                     "embedded_cache": {
@@ -161,7 +164,7 @@ class ConfigBuilder:
                         "enabled": True
                     }
                 }
-            }
+            },
         }
 
     @property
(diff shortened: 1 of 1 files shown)


src/config_builder.py
```diff
@@ -143,4 +147,53 @@ def _limits_config(self) -> dict:
...
+                }
+            }
+        }
+
+    @property
+    def _frontend(self) -> dict:
+        # Ref: https://grafana.com/docs/loki/latest/configure/#frontend
+        return {
+            # Maximum number of outstanding requests per tenant per frontend; requests beyond this error with HTTP 429.
+            # Default is 2048, but 8cpu16gb can ingest ~3 times more, so set to 4x.
```
simskij, line 187: This does not add up. If it can ingest 3 times more, why are we then multiplying it by 4? 😅

simskij:
Didn't we say we also wanted to expose config options, allowing users to override our defaults?

src/config_builder.py
sed-i: 8cpu16gb is just what we use for the load test. In the wild this will obviously be sized differently.
I think at this point taking 4x the default would be sufficient.

- Generally the value should vary with vm size because for smaller vm some users may prefer getting 4xx and 5xx for dropped instructions instead of increase in latency.
- At the moment we have "one size fits all" so with this change here I'm effectively proposing to be opinionated favoring latency over failures for smaller VMs, and failures over latency for higher load.

I can delete the comment so it's less confusing :)

sed-i: Tested the same config with 4cpu8gb and 180k loglines/min and it looks good - not much change in resource use compared to before the change.

PR #332, opened 2024-01-12
Add support for pebble log forwarding

### LINKED ISSUE #333: Add Pebble log forwarding support in loki_push_api
### Enhancement Proposal

The `loki_push_api` library should support Pebble log forwarding.

We should implement a new class that a charm can instantiate to configure Pebble log forwarding. That class should be able to receive the Loki endpoints to push to in two ways:
1. via explicitly provided `loki_endpoints`;
2. when those are missing, via the `logging` relation.

The option 1 allows charms like **mimir-worker-k8s** to fetch the charm library and configure Pebble log forwarding without a charm relation.

lucabello:
## Issue

Closes #333.

## Solution

The solution has changed so many times, you'll probably have a better understanding by reading the documentation additions in the PR.


lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2313,60 @@ def _containers(self) -> Dict[str, Container]:
         return {cont: self._charm.unit.get_container(cont) for cont in self._logs_scheme.keys()}
 
 
+class PebbleLogForwarder:
```
dstathis, line 2316: What do you think about calling it something like `StdoutForwarder`?

lucabello: I don't mind; it probably makes what it does more clear

lucabello:
TODO: make sure that the class is observing logging relation (or `relation_name`) changed and if there's a loki endpoint it should enable log forwarding automatically.

A client charm that instantiate this and relates to Loki directly, should be able to only call the constructor and have things work out of the box.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2313,92 @@ def _containers(self) -> Dict[str, Container]:
...
+                "location": endpoint,
+                "services": services_value,
+                "labels": {
+                    "product": "Juju",
+                    "charm_name": self.topology._charm_name,
+                    "model": self.topology._model,
+                    "model_uuid": self.topology._model_uuid,
+                    "application": self.topology._application,
+                    "unit": self.topology._unit,
+                },
```
lucabello, line 2353: Minor change to make the labels consistent with our usual charm metrics:

```suggestion
                "labels": {
                    "product": "Juju",
                    "charm": self.topology._charm_name,
                    "juju_model": self.topology._model,
                    "juju_model_uuid": self.topology._model_uuid,
                    "juju_application": self.topology._application,
                    "juju_unit": self.topology._unit,
                },
```

lib/charms/loki_k8s/v1/loki_push_api.py
lucabello: We just agreed to name it `LogForwarder`.

IbraAoad:
Unit tests are still missing validations of the actual pebble layer due to this https://github.com/canonical/operator/pull/1111

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -20,6 +20,9 @@
 send telemetry, such as logs, to Loki through a Log Proxy by implementing the consumer side of the
 `loki_push_api` relation interface.
 
+- 'LogForwarder': This object can be used by any Charmed Operator which needs to send its
```
PietroPasotti, line 23:
```suggestion
- 'LogForwarder': This object can be used by any Charmed Operator which wants any stdout emitted by a workload container to be captured and automatically forwarded to Loki. 
```

PietroPasotti: as it's written, it gives the impression we're talking about the charm process' own stdout

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,72 @@ def _promtail_error(self, event):
     )
 ```
 
+## LogForwarder Library Usage
+
+Let's say that we have a workload charm that produces logs to the standard output (stdout),
```
PietroPasotti, line 357:
```suggestion
Let's say that we have a charm's workload that writes logs to the standard output (stdout),
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,72 @@ def _promtail_error(self, event):
...
+and we need to send those logs to a workload implementing the `loki_push_api` interface,
```
PietroPasotti, line 358:
```suggestion
and we need to send those logs to a Loki instance.
```

if you like, you can add something like: "to know how to reach a loki instance, a charm typically would use the `loki_push_api` interface".

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,72 @@ def _promtail_error(self, event):
...
+such as `Loki` or `Grafana Agent`.
+
+Use the `LogForwarder` class by instantiating it in the `__init__` method of the charmed
```
PietroPasotti, line 361: I know technically they're synonyms, but usually we write 'charm', not 'charmed operator'. reads more easily

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,72 @@ def _promtail_error(self, event):
...
+operator. There are two ways to provide Loki endpoint(s) to the forwarder. You can let the
```
PietroPasotti, line 362: I'd say:
"The easiest way to pass a Loki endpoint to the forwarder is to provide it with the name of an endpoint which uses the loki_push_api interface. The forwarder will automatically extract from there all the data it needs.
Otherwise, you can manually pass a Loki endpoint to which the logs should be forwarded."

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,72 @@ def _promtail_error(self, event):
...
+object extract the endpoint(s) from relation data, or you can explicitly pass the Loki
+endpoint(s) to forward your logs to.
+
+1. If you want your charm to relate to another implementing the `loki_push_api` interface,
```
PietroPasotti, line 366:
```suggestion
1. To use the LogForwarder, the first thing you need to do is instantiate it in your charm's constructor:
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,72 @@ def _promtail_error(self, event):
...
+          self._log_forwarder = LogForwarder(self)
+   ```
+
+   The `LogForwarder` will listen to relation events out-of-the-box and enable or disable
+   the log forwarding accordingly. Next, modify the `metadata.yaml` file to add:
+
+   - The `log-forwarding` relation in the `requires` section:
+     ```yaml
+     requires:
+       log-forwarding:
```
PietroPasotti, line 385: in this case, the example you give above will not work, because the LogForwarder's endpoint defaults to 'logging', not 'log-forwarding'.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,72 @@ def _promtail_error(self, event):
...
+   from charms.loki_k8s.v1.loki_push_api import LogForwarder
+
+   ...
+
+      def __init__(self, *args):
+          ...
+          self._log_forwarder = LogForwarder(self)
+   ```
+
+   The `LogForwarder` will listen to relation events out-of-the-box and enable or disable
```
PietroPasotti, line 379:
```suggestion
   The `LogForwarder` by default will observe relation events on the `logging` endpoint and enable/disable log forwarding automatically.
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,72 @@ def _promtail_error(self, event):
...
+             loki_endpoints=self.get_loki_endpoints,
+         )
+         self.framework.observe(self.on["foo"].relation_joined, self._log_forwarder.enable())
+         self.framework.observe(self.on["foo"].relation_changed, self._log_forwarder.enable())
+         self.framework.observe(self.on["foo"].relation_departed, self._log_forwarder.disable())
+         self.framework.observe(self.on["foo"].relation_broken, self._log_forwarder.disable())
+
+
+     def get_loki_endpoints(self) -> List[str]:
+         return self.model.relations["foo"].get("loki_endpoints", None)
```
PietroPasotti, line 414: `model.relations[x]` returns List[Relation], not Relation.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,72 @@ def _promtail_error(self, event):
...
+
+   ...
+
+     def __init__(self, *args):
+         ...
+         self._log_forwarder = LogForwarder(
+             self,
+             loki_endpoints=self.get_loki_endpoints,
+         )
+         self.framework.observe(self.on["foo"].relation_joined, self._log_forwarder.enable())
```
PietroPasotti, line 407: I think this API could be improved by: passing to logforwarder the endpoint which it should listen to, and a Callable[[Relation], endpoint].
Since it seems that what you're doing with the relation events is generic, and what is specific is only 'how to extract the data I need from a relation'. 

not sure I'm explaining myself well here, contact me for more.
In short:

```
  self._log_forwarder = LogForwarder(
             self,
             binding="other-endpoint", # defaults to 'logging'
             loki_endpoint_getter=self._get_loki_endpoints, # defaults to whatever we need to do to get the endpoint from logging:loki_push_api
         )

def _get_loki_endpoints(self, relation:Relation) -> str:
    return relation.data[relation.app]['very-specific-way-to-store-an-url']

```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,72 @@ def _promtail_error(self, event):
...
+         self.framework.observe(self.on["foo"].relation_departed, self._log_forwarder.disable())
+         self.framework.observe(self.on["foo"].relation_broken, self._log_forwarder.disable())
+
+
+     def get_loki_endpoints(self) -> List[str]:
+         return self.model.relations["foo"].get("loki_endpoints", None)
+
+   ```
+
+Once the library is implemented in a Charmed Operator, and log forwarding is enabled, the library
```
PietroPasotti, line 418:
```suggestion
Once the library is implemented in a Charmed Operator, and a logging relation (loki_push_api) is active and healthy, the library
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,72 @@ def _promtail_error(self, event):
...
+will inject a Pebble layer in the workload container to send logs.
```
PietroPasotti, line 419:
```suggestion
will inject a Pebble layer in the workload container to configure pebble's log-forwarding feature and start sending logs to Loki.
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2382,123 @@ def _containers(self) -> Dict[str, Container]:
         return {cont: self._charm.unit.get_container(cont) for cont in self._logs_scheme.keys()}
 
 
+class LogForwarder(Object):
+    """Forward the StdOut output to one or multiple Loki endpoints."""
```
PietroPasotti, line 2386:
```suggestion
    """Forward the standard outputs of all workloads operated by a charm to one or multiple Loki endpoints."""
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2382,123 @@ def _containers(self) -> Dict[str, Container]:
...
+                    # This is a peer unit
+                    continue
+
+                endpoint = relation.data[unit].get("endpoint")
+                if endpoint:
+                    deserialized_endpoint = json.loads(endpoint)
+                    url = deserialized_endpoint.get("url")
+                    if url:
+                        endpoints.append(url)
+        endpoints.sort()
```
PietroPasotti, line 2498: why do we do this? test repeatability?

PietroPasotti (changes requested):
I see some issues with the API and a bunch of improvements that can be made on the documentation side. Take a look and let me know!

dstathis (changes requested):

lib/charms/loki_k8s/v1/loki_push_api.py
lucabello: This is cool, done!

lib/charms/loki_k8s/v1/loki_push_api.py
lucabello: I think so!

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2382,123 @@ def _containers(self) -> Dict[str, Container]:
...
+
+    def _on_logging_relation_changed(self, _):
+        self.loki_endpoints = self._fetch_endpoints
+        self.enable()
+
+    def _on_logging_relation_departed(self, _):
+        self.disable()
+
+    def _on_logging_relation_broken(self, _):
+        self.disable()
```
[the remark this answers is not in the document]
lucabello: True, that would incorrectly disable both; I think it made more sense on a previous iteration of the code, but what you're suggesting could be the way to go

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,77 @@ def _promtail_error(self, event):
     )
 ```
 
+## LogForwarder Library Usage
```
PietroPasotti, line 355:
```suggestion
## LogForwarder object Usage
```

PietroPasotti: or 'class' as you say below

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,77 @@ def _promtail_error(self, event):
...
+              self,
+              relation_name="logging"  # optional, defaults to `logging`
+          )
+   ```
+
+   The `LogForwarder` by default will observe relation events on the `logging` endpoint and
+   enable/disable log forwarding automatically.
+   Next, modify the `metadata.yaml` file to add:
+
+   - The `log-forwarding` relation in the `requires` section:
```
PietroPasotti, line 387:
```suggestion
   - The `logging` relation in the `requires` section:
```

PietroPasotti: also, this is a list with a single bullet. I'd remove the list

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,77 @@ def _promtail_error(self, event):
...
+     ```yaml
+     requires:
+       logging:
+         interface: loki_push_api
+         optional: true
+     ```
+
+2. If you don't want to relate your charm to another implementing the `loki_push_api` interface,
```
PietroPasotti, line 395:
```suggestion
2. Alternatively,
```

(maybe it's not that they don't want to, but they can't. Reason doesn't matter.)

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,77 @@ def _promtail_error(self, event):
...
+   you need to provide two things: the `relation_name` of the relation containing the endpoint(s),
+   and a `loki_endpoints_getter` function that extracts the endpoint(s) from that relation.
```
PietroPasotti, line 397:
```suggestion
   and a `loki_endpoints_getter` function to define how to extract the Loki endpoint(s) from that relation.
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,77 @@ def _promtail_error(self, event):
...
+         self._log_forwarder = LogForwarder(
+             self,
+             relation_name="foo",
+             loki_endpoints_getter=self.get_loki_endpoints("foo"),
+         )
+
+     def get_loki_endpoints(self, relation_name: str) -> List[str]:
+         endpoints = []
+         for relation in self.model.relations[relation_name]:
+             endpoint = some_way_to_retrieve_the_endpoint()
```
PietroPasotti, line 416:
```suggestion
             endpoint = some_way_to_retrieve_the_endpoint(relation.data)
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,77 @@ def _promtail_error(self, event):
...
+             endpoints.append(endpoint)
+
+         return endpoints
```
PietroPasotti, line 419: is there anything we should be clear about what kind of endpoints these can be? 
List[str] is a bit broad.
Do they need to be fully resolvable URLS? Hostnames only? `hostname:port`?

I'm thinking we need something like:
 "resolvable url of a `loki-push-api`-compliant endpoint"

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,77 @@ def _promtail_error(self, event):
...
+   ```
+
+Once the library is implemented in a charm and a logging relation (loki_push_api) is
```
PietroPasotti, line 422:
```suggestion
Once the LogForwarder class is implemented in your charm and the relation (`logging`, or else a custom one)
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +352,77 @@ def _promtail_error(self, event):
...
+active and healthy, the library will inject a Pebble layer in the workload container
```
PietroPasotti, line 423:
```suggestion
active and healthy, the library will inject a Pebble layer in each workload container the charm has access to (only if they can connect(?))
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2387,120 @@ def _containers(self) -> Dict[str, Container]:
...
+        charm: CharmBase,
+        *,
+        relation_name: str = DEFAULT_RELATION_NAME,
+        loki_endpoint_getter: Optional[Callable[[str], List[str]]] = None,
+    ):
+        super().__init__(charm, relation_name)
+        self._charm = charm
+        self._relation_name = relation_name
+        self.topology = JujuTopology.from_charm(charm)
+        self.endpoints_getter: Callable = (
```
PietroPasotti, line 2404: this should be private IMHO. Also `topology`, and `loki_endpoints`

Take a close look at the public signature of this class and remove all that shouldn't be visible.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2387,120 @@ def _containers(self) -> Dict[str, Container]:
...
+
+    def _build_log_targets(self, endpoints: Optional[List[str]], enable=False):
+        """Build the targets for the log forwarding Pebble layer."""
+        targets = {}
+        if endpoints:
+            for i, endpoint in enumerate(endpoints):
+                targets.update(self._build_log_target(endpoint, i, enable))
+        return targets
+
+    def handle_logging(self, enabled=False):
```
PietroPasotti, line 2465: also private?

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2387,120 @@ def _containers(self) -> Dict[str, Container]:
...
+    def _on_logging_relation_changed(self, _):
+        self.enable()
+
+    def _on_logging_relation_departed(self, _):
+        self.disable()
+
+    def _on_logging_relation_broken(self, _):
+        self.disable()
+
+    def enable(self):
```
PietroPasotti, line 2427: Perhaps enable/disable should also be private? There could be a use case for manually managing the status of this patch but until it's clear I say let's make this private. It's easier to make it public again if we change our mind than do the opposite.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2387,120 @@ def _containers(self) -> Dict[str, Container]:
...
+        """Disable log forwarding."""
+        self.handle_logging(enabled=False)
+
+    def _build_log_target(self, endpoint, index, enabled=False):
+        """Build a log target for the log forwarding Pebble layer."""
+        dest_name = f"loki{index}"
+        services_value = ["all"] if enabled else ["-all"]
+
+        return {
+            dest_name: {
```
PietroPasotti, line 2441: are you using dest_name in only one place? Then I'd not give it a name
```suggestion
            f"loki{index}": {
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2387,120 @@ def _containers(self) -> Dict[str, Container]:
...
+        self.handle_logging(enabled=True)
+
+    def disable(self):
+        """Disable log forwarding."""
+        self.handle_logging(enabled=False)
+
+    def _build_log_target(self, endpoint, index, enabled=False):
+        """Build a log target for the log forwarding Pebble layer."""
+        dest_name = f"loki{index}"
+        services_value = ["all"] if enabled else ["-all"]
```
PietroPasotti, line 2438:
```suggestion
        services_to_monitor = ["all"] if enabled else ["-all"]
```

Also I'd add maybe a link to the reference/docs that explains the syntax of this minilanguage
And a comment clarifying that with `-all` + override semantics we're disabling all log-forwarding configs applied by previous layers

lib/charms/loki_k8s/v1/loki_push_api.py
PietroPasotti: That'd not be a good reason IMHO.
I think it's actually because otherwise the index in `_build_log_targets` might refer to different services every time, while we want the index to be persistent and always assigned to the same service.
This should be made clear in a comment otherwise it's not obvious why we're sorting that list. Also perhaps we shouldn't sort in this method but in `_build_log_targets`, where it's actually clear why we need the sorting.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2387,120 @@ def _containers(self) -> Dict[str, Container]:
...
+                container.add_layer(f"{container_name}-log-forwarding", layer, combine=True)
+        else:
+            logger.warning("No loki endpoints available")
+
+    def _fetch_endpoints(self, relation_name: str) -> List[str]:
+        """Fetch Loki Push API endpoints through relation data.
+
+        Returns:
+            A list of Loki Push API endpoint URLs, for instance:
+            [
```
PietroPasotti, line 2481: this example data should also be in the module docstring

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2387,120 @@ def _containers(self) -> Dict[str, Container]:
...
+            layer_config = {"log-targets": self._build_log_targets(self.loki_endpoints, enabled)}
+            layer = Layer(layer_config)  # pyright: ignore
+
+            for container_name, container in self._charm.unit.containers.items():
+                container.add_layer(f"{container_name}-log-forwarding", layer, combine=True)
+        else:
+            logger.warning("No loki endpoints available")
+
+    def _fetch_endpoints(self, relation_name: str) -> List[str]:
+        """Fetch Loki Push API endpoints through relation data.
```
PietroPasotti, line 2477:
```suggestion
        """Fetch Loki Push API endpoints from relation data.
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2382,123 @@ def _containers(self) -> Dict[str, Container]:
...
+                "http://loki2:3100/loki/api/v1/push",
+            ]
+        """
+        endpoints = []  # type: list
+
+        for relation in self._charm.model.relations[self._relation_name]:
+            for unit in relation.units:
+                if unit.app == self._charm.app:
+                    # This is a peer unit
+                    continue
```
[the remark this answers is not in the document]
PietroPasotti: IIRC there are. relation.units is 'all units involved in the relation' i.e. also yourself and your peers. Which is fun because you can't actually read your peer's databags (if it's not a peer relation)

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2387,120 @@ def _containers(self) -> Dict[str, Container]:
...
+        endpoints = []  # type: list
+
+        for relation in self._charm.model.relations[relation_name]:
+            for unit in relation.units:
+                if unit.app == self._charm.app:
+                    # This is a peer unit
+                    continue
+
+                endpoint = relation.data[unit].get("endpoint")
+                if endpoint:
```
PietroPasotti, line 2495: 1) if there is no "endpoint", the relation is bork and we should at least log it, and surface it somehow to the charm so it has a chance to set blocked status.

I'm thinking to expose a `self.is_ready()` method like we do in other charm libs that verifies that 1) there are relations 2) they are healthy i.e. there is data and the data has the format we expect. Maybe [DatabagModel](https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_k8s/v2/ingress.py#L86) is too much for this use case, but maybe not. Else, use charm-relation-interfaces' models.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2387,120 @@ def _containers(self) -> Dict[str, Container]:
...
+                    deserialized_endpoint = json.loads(endpoint)
```
PietroPasotti, line 2496: 2) if this fails, do we just raise an uncaught error? This will bork the charm's initialization

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2387,120 @@ def _containers(self) -> Dict[str, Container]:
...
+                    url = deserialized_endpoint.get("url")
+                    if url:
```
PietroPasotti, line 2498: if there's no url, relation data is bork and this should be surfaced

tests/unit/test_log_forwarder.py
```diff
@@ -0,0 +1,151 @@
...
+
+        requires:
+          logging:
+            interface: loki_push_api
+        """
+    )
+
+    def __init__(self, *args):
+        super().__init__(*args)
+        self.log_forwarder = LogForwarder(self)
```
PietroPasotti, line 31: we also need to test the 'non-default loki_push_api endpoint path'

perhaps parametrize? And, we need to test the get_loki_endpoint callback

What's the coverage of the lib?

PietroPasotti (changes requested):
We're heading in the right direction, but I still have some thoughts on the API and the docs

lib/charms/loki_k8s/v1/loki_push_api.py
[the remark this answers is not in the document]
lucabello: I'll leave this in then :)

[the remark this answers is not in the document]
dstathis: This shouldn't be the case. It should just be remote units. Can we verify this and, if it's true, create an operator bug.

lib/charms/loki_k8s/v1/loki_push_api.py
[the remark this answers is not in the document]
PietroPasotti: holistically rebuilding the layer on every event might not work if we use the merge strategy:

- receive an event with remotes loki/0, loki/1 --> create layer containing `["all"]` and endpoints from `loki/0, loki/1`
- receive an event with remotes loki/0 --> create layer containing `["all"]` and endpoints from `loki/0`, but previous layer still has loki/1.

if we use the replace strategy, maybe things can be simplified.

lib/charms/loki_k8s/v1/loki_push_api.py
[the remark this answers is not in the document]
PietroPasotti: you're right, it was fixed some time ago
We can remove all those checks

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2395,155 @@ def _containers(self) -> Dict[str, Container]:
         return {cont: self._charm.unit.get_container(cont) for cont in self._logs_scheme.keys()}
 
 
+class LogForwarder(Object):
+    """Forward the standard outputs of all workloads operated by a charm to one or multiple Loki endpoints."""
```
dstathis, line 2399: This needs to inherit from `ConsumerBase` so that alert rules can be handled.

lib/charms/loki_k8s/v1/loki_push_api.py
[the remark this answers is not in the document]
lucabello: Unfortunately my understanding is that:
* you can't specify multiple URLs in one target;
* the "override" strategy (e.g., `merge | replace`) is at the same level of the endpoint, so you can't replace all of them with only one.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2395,155 @@ def _containers(self) -> Dict[str, Container]:
...
+
+        return {unit_name: log_target}
+
+    def _build_log_targets(self, endpoints: Optional[Dict[str, str]], enable=False):
+        """Build the targets for the log forwarding Pebble layer."""
+        targets = {}
+        if endpoints:
+            for unit_name, endpoint in endpoints.items():
+                targets.update(self._build_log_target(unit_name, endpoint, enable))
+        return targets
```
Abuelodelanada, line 2475: I would remove one indentation level:

```suggestion
        targets = {}
        if not endpoints:
            return targets

        for unit_name, endpoint in endpoints.items():
            targets.update(self._build_log_target(unit_name, endpoint, enable))
        return targets
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2396,156 @@ def _containers(self) -> Dict[str, Container]:
...
+            loki_endpoints.update(self._fetch_endpoints(relation))
+
+        if loki_endpoints:
+            layer_config = {"log-targets": self._build_log_targets(loki_endpoints, enable=True)}
+            layer = Layer(layer_config)  # pyright: ignore
+
+            for container_name, container in self._charm.unit.containers.items():
+                container.add_layer(f"{container_name}-log-forwarding", layer, combine=True)
+        else:
+            logger.warning("No loki endpoints available")
```
Abuelodelanada, line 2490: Use guard clause instead of if-else

```suggestion
    def _enable_logging(self) -> None:
        """Enable the log forwarding."""
        loki_endpoints = {}
        for relation in self._charm.model.relations[self._relation_name]:
            loki_endpoints.update(self._fetch_endpoints(relation))

        if not loki_endpoints:
            logger.warning("No loki endpoints available")
            return
            
        layer_config = {"log-targets": self._build_log_targets(loki_endpoints, enable=True)}
        layer = Layer(layer_config)  # pyright: ignore

        for container_name, container in self._charm.unit.containers.items():
            container.add_layer(f"{container_name}-log-forwarding", layer, combine=True)
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2396,156 @@ def _containers(self) -> Dict[str, Container]:
...
+        self._topology = JujuTopology.from_charm(charm)
+        self._endpoints_getter: Callable = (
+            self._extract_urls if loki_endpoints_getter is None else loki_endpoints_getter
+        )
+
+        on_relation = self._charm.on[self._relation_name]
+        self.framework.observe(on_relation.relation_joined, self._on_logging_relation_joined)
+        self.framework.observe(on_relation.relation_changed, self._on_logging_relation_changed)
+        self.framework.observe(on_relation.relation_departed, self._on_logging_relation_departed)
+        self.framework.observe(on_relation.relation_broken, self._on_logging_relation_broken)
```
Abuelodelanada, line 2426: `_relation` is redundant

```suggestion
        on = self._charm.on[self._relation_name]
        self.framework.observe(on.relation_joined, self._on_logging_relation_joined)
        self.framework.observe(on.relation_changed, self._on_logging_relation_changed)
        self.framework.observe(on.relation_departed, self._on_logging_relation_departed)
        self.framework.observe(on.relation_broken, self._on_logging_relation_broken)
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2396,156 @@ def _containers(self) -> Dict[str, Container]:
...
 
+class LogForwarder(ConsumerBase):
+    """Forward the standard outputs of all workloads operated by a charm to one or multiple Loki endpoints."""
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        *,
+        relation_name: str = DEFAULT_RELATION_NAME,
+        loki_endpoints_getter: Optional[Callable[[Relation], List[str]]] = None,
```
lucabello, line 2407: This should be reverted to just a Dict[str,str] in the form of `{unit_name: url}` instead of a Callable. And it should be called `loki_endpoints`.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2396,156 @@ def _containers(self) -> Dict[str, Container]:
...
+        alert_rules_path: str = DEFAULT_ALERT_RULES_RELATIVE_PATH,
+        recursive: bool = True,
+        skip_alert_topology_labeling: bool = False,
+    ):
+        super().__init__(
```
lucabello, line 2412: We should call this `super().__init__()` ONLY when `loki_endpoints is None`.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2396,156 @@ def _containers(self) -> Dict[str, Container]:
...
+    def _on_logging_relation_changed(self, _):
+        self._enable_logging()
+
+    def _on_logging_relation_departed(self, event: RelationDepartedEvent):
+        self._disable_logging(unit=event.unit)
+
+    def _on_logging_relation_broken(self, event: RelationBrokenEvent):
+        for unit in event.relation.units:
+            if unit.app == self._charm.app:
+                continue
```
lucabello, line 2440: This should be removed.

dstathis (approved):

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2396,156 @@ def _containers(self) -> Dict[str, Container]:
         return {cont: self._charm.unit.get_container(cont) for cont in self._logs_scheme.keys()}
 
 
+class LogForwarder(ConsumerBase):
+    """Forward the standard outputs of all workloads operated by a charm to one or multiple Loki endpoints."""
+
+    def __init__(
```
lucabello, line 2402: This should check if the Juju version is at least 3.4:
* something like [this](https://github.com/canonical/observability-libs/blob/main/lib/charms/observability_libs/v1/cert_handler.py#L413)
* compare JujuVersion objects explained [here](https://github.com/canonical/operator/blob/dc47640f8ea915049d9e20a1fe4ea189c96ae8d1/ops/jujuversion.py#L24)

lib/charms/loki_k8s/v1/loki_push_api.py
lucabello: Actually, this is done because `RelationBrokenEvent` doesn't indicate a single unit, so we need to iterate over all the units in that relation.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +357,107 @@ def _promtail_error(self, event):
...
+   ...
+
+       def __init__(self, *args):
+           ...
+           self._log_forwarder = ManualLogForwarder(
+                self,
+                loki_endpoints=get_endpoints_from_somewhere()
+            )
+
+           self.framework.observe(self.on.some_event, self._log_forwarder.update_logging())
```
PietroPasotti, line 421: probably this is what you need:

```suggestion
           self.framework.observe(self.on.some_event, self._log_forwarder.update_logging)
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +357,107 @@ def _promtail_error(self, event):
...
+   automatically be called whenever that relation triggers an event:
+
+   ```python
+   ...
+
+       def __init__(self, *args):
+           ...
+           self._log_forwarder = ManualLogForwarder(
+                self,
+                loki_endpoints=get_endpoints_from_somewhere(),
```
PietroPasotti, line 445:
```suggestion
                loki_endpoints=self.get_endpoints_from_somewhere(),
```

is that a callback, or a function call?

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +357,107 @@ def _promtail_error(self, event):
...
+                loki_endpoints=get_endpoints_from_somewhere()
+            )
+
+           self.framework.observe(self.on.some_event, self._log_forwarder.update_logging())
+
+       def get_endpoints_from_somewhere():
+           # Example function that gets the endpoints from Juju config.
+           #
+           # Returns:
+           #     A dictionary of remote units and respective Loki endpoint.
```
PietroPasotti, line 427: This is confusing. does the dict key really need to be a juju unit name, or is it fine if it's a unique name?

The example below maps `loki/0` to a url starting with `loki1`. What's going on?

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +357,107 @@ def _promtail_error(self, event):
...
+           #     {
+           #         "loki/0": "http://loki1:3100/loki/api/v1/push",
+           #     }
+           #
+           return self.config["loki_endpoints"]
+   ```
+
+2. Alternatively, you can pass a `relation_name` to the constructor; `update_logging` will
+   automatically be called whenever that relation triggers an event:
```
PietroPasotti, line 436:
```suggestion
   be emitted whenever that relation triggers an event:
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -349,6 +357,107 @@ def _promtail_error(self, event):
...
+   ```python
+   ...
+
+       def __init__(self, *args):
+           ...
+           self._log_forwarder = ManualLogForwarder(
+                self,
+                loki_endpoints=get_endpoints_from_somewhere(),
+                relation_name="some-relation"
+            )
```
PietroPasotti, line 447: I'd like to see in this example what the intended usage of `update_logging` handlers is

PietroPasotti (changes requested):
Still some unclarity left in the updated API.

lib/charms/loki_k8s/v1/loki_push_api.py
lucabello: It needs to be unique, so the unit name fits; `loki1` is put there as a placeholder, I can change it to just `loki` if it's less confusing

lib/charms/loki_k8s/v1/loki_push_api.py
dstathis: Should be a function call if I understand correctly.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2313,6 +2423,265 @@ def _containers(self) -> Dict[str, Container]:
...
 
 
+class _LogForwarderHelpers:
+    @staticmethod
+    def check_juju_version() -> bool:
+        """Make sure the Juju version supports Log Forwarding."""
+        juju_version = JujuVersion.from_environ()
+        if not juju_version > JujuVersion(version=str("3.3")):
+            msg = f"Juju version {juju_version} does not support Pebble log forwarding. Juju >= 3.4 is needed."
+            logger.warn(msg)
```
dstathis, line 2433: `warn` is deprecated. Use `logger.warning` instead.
```suggestion
            logger.warning(msg)
```

dstathis (approved):

lib/charms/loki_k8s/v1/loki_push_api.py
lucabello: Done!

Abuelodelanada:
Shouldn't we have itest that verifies this feature?

For instance would be nice to have an itest that checks if you [have a charm with 2 containers](https://github.com/canonical/loki-k8s-operator/blob/main/tests/integration/log-proxy-tester/metadata.yaml#L11) everything works

PR #349, opened 2024-02-13
Handle restart and status better

sed-i:
## Issue
While reproducing https://github.com/canonical/grafana-agent-k8s-operator/issues/275, I encountered the following issue:
1. After `juju deploy`, loki wasn't running. Charm was in blocked: "Error connecting to Loki. Check juju debug-log". From debug-log: `Checking alert rules: 400 - Bad Request`.
2. After manually starting loki via cli from the workload container, `/usr/bin/loki -config.file=/etc/loki/loki-local-config.yaml`, it was using `/tmp/loki/` instead of `/loki`, and indeed, `cat /etc/loki/loki-local-config.yaml | grep directory` showed the default values, which means the charm didn't write the config file up to that point. 

This could have happened because loki was deployed without `--trust` and a later status overwrote a previous status, giving the wrong idea.

## Solution
- [x] Use the `restart` variable correctly.
- [x] Centralized status setting.

Drive-by fixes:
- Use `Unit.set_ports` instead of KubernetesServicePatch.


lib/charms/observability_libs/v1/kubernetes_service_patch.py
PietroPasotti, line 1: :partying_face:

src/charm.py
```diff
@@ -341,39 +391,23 @@ def _configure(self):  # noqa: C901
...
-        ):
+        if isinstance(to_status(self._stored.status["rules"]), BlockedStatus):
             # Wait briefly for Loki to come back up and re-check the alert rules
             # in case an upgrade/refresh caused the check to occur when it wasn't
-            # ready yet
+            # ready yet. TODO: use custom pebble notice for "workload ready" event.
             time.sleep(2)
             self._check_alert_rules()
-            return
+            return  # TODO: why do we have a return here?
```
PietroPasotti, line 410: that's a very good question :D

PietroPasotti (approved):
Looks like a change for the better. Not convinced about the copy-paste of the composite status thingy in all of our repos. Let's publish a charm lib?

PR #330, opened 2024-01-10
Support juju proxy settings

nrobinaubertin:
## Issue
Fixes https://github.com/canonical/loki-k8s-operator/issues/326

## Solution
I declared a ProxyHandler for the urllib request that is made to download the promtail binary.

## Testing Instructions
I used the discourse-k8s charm to test the change.  
1. Add the cilium addon to microk8s
2. Deploy squid on microk8s like proposed [here](https://hub.docker.com/r/ubuntu/squid)
3. Create a model
4. apply a network-policy to that model only allowing local networking (like [this](https://gist.github.com/nrobinaubertin/f583409700bced7edf03827d3fd48134)) 
5. Apply the correct juju-http{s,}-proxy settings to the model
6. relate all the charms to discourse-k8s, it should be operational in the end.

## Release Notes
Add support for JUJU_CHARM_{HTTP,HTTPS}_PROXY variables in the loki_push_api lib.

PR #352, opened 2024-02-23
Fix logic of handling proxies

nrobinaubertin:
## Issue
Support for juju proxy settings was added with the following PR: https://github.com/canonical/loki-k8s-operator/pull/330.
Unfortunately, a logic issue in the code went unnoticed and the proxy settings were therefore never applied.

## Solution
Instead of testing the boolean value of the `proxies` variable (which is `False` in case of an empty `dict`), we test that `proxies` is an instance of a `dict`.

## Context
N/A

## Testing Instructions
N/A

## Upgrade Notes
N/A

### DIFF
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -2116,17 +2116,15 @@ class LogProxyConsumer(ConsumerBase):
                - "binsha": sha256 sum of unpacked promtail binary
         """
         # Check for Juju proxy variables and fall back to standard ones if not set
-        proxies: Optional[Dict[str, str]] = {}
-        if isinstance(proxies, dict) and os.environ.get("JUJU_CHARM_HTTP_PROXY"):
-            proxies.update({"http": os.environ["JUJU_CHARM_HTTP_PROXY"]})
-        if isinstance(proxies, dict) and os.environ.get("JUJU_CHARM_HTTPS_PROXY"):
-            proxies.update({"https": os.environ["JUJU_CHARM_HTTPS_PROXY"]})
-        if isinstance(proxies, dict) and os.environ.get("JUJU_CHARM_NO_PROXY"):
-            proxies.update({"no_proxy": os.environ["JUJU_CHARM_NO_PROXY"]})
         # If no Juju proxy variable was set, we set proxies to None to let the ProxyHandler get
         # the proxy env variables from the environment
-        if isinstance(proxies, dict) and not proxies:
-            proxies = None
+        proxies: Optional[Dict[str, str]] = {
+            "https_proxy": os.environ.get("JUJU_CHARM_HTTPS_PROXY", ""),
+            "http_proxy": os.environ.get("JUJU_CHARM_HTTP_PROXY", ""),
+            "no_proxy": os.environ.get("JUJU_CHARM_NO_PROXY", ""),
+        }
+        if isinstance(proxies, dict):
+            proxies = {k: v for k, v in proxies.items() if v != ""} or None
 
         proxy_handler = request.ProxyHandler(proxies)
         opener = request.build_opener(proxy_handler)
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -2123,7 +2123,9 @@ class LogProxyConsumer(ConsumerBase):
             proxies.update({"https": os.environ["JUJU_CHARM_HTTPS_PROXY"]})
         if isinstance(proxies, dict) and os.environ.get("JUJU_CHARM_NO_PROXY"):
             proxies.update({"no_proxy": os.environ["JUJU_CHARM_NO_PROXY"]})
-        else:
+        # If no Juju proxy variable was set, we set proxies to None to let the ProxyHandler get
+        # the proxy env variables from the environment
+        if isinstance(proxies, dict) and not proxies:
             proxies = None
 
         proxy_handler = request.ProxyHandler(proxies)
--- a/lib/charms/loki_k8s/v0/loki_push_api.py
+++ b/lib/charms/loki_k8s/v0/loki_push_api.py
@@ -480,7 +480,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 26
+LIBPATCH = 27
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


lib/charms/loki_k8s/v0/loki_push_api.py
```diff
@@ -2117,11 +2117,11 @@ def _download_and_push_promtail_to_workload(self, promtail_info: dict) -> None:
...
+        if isinstance(proxies, dict) and os.environ.get("JUJU_CHARM_HTTP_PROXY"):
             proxies.update({"http": os.environ["JUJU_CHARM_HTTP_PROXY"]})
-        if proxies and os.environ.get("JUJU_CHARM_HTTPS_PROXY"):
+        if isinstance(proxies, dict) and os.environ.get("JUJU_CHARM_HTTPS_PROXY"):
             proxies.update({"https": os.environ["JUJU_CHARM_HTTPS_PROXY"]})
-        if proxies and os.environ.get("JUJU_CHARM_NO_PROXY"):
+        if isinstance(proxies, dict) and os.environ.get("JUJU_CHARM_NO_PROXY"):
             proxies.update({"no_proxy": os.environ["JUJU_CHARM_NO_PROXY"]})
         else:
             proxies = None
```
mthaddon, line 2127: This means that if you set JUJU_CHARM_HTTP_PROXY and JUJU_CHARM_HTTPS_PROXY but not JUJU_CHARM_NO_PROXY `proxies` will be `None`. Is that correct?

nrobinaubertin: No this is not correct. Thanks for pointing that out. We want `proxies` to be `None` only if nothing was set (I can't think of a sound case where JUJU_HTTP_PROXY is set on one side and HTTPS_PROXY on the other)

dstathis (commented):
Is there a reason  we are checking the proxies variable at all? It was explicitly instantiated to empty dict on the previous line.

nrobinaubertin:
> Is there a reason we are checking the proxies variable at all? It was explicitly instantiated to empty dict on the previous line.

I know but our lord and savior mypy wants me to follow its path

lib/charms/loki_k8s/v0/loki_push_api.py
sed-i: Would it work to have
```python
            proxies = {
                "https_proxy": os.environ.get("JUJU_CHARM_HTTPS_PROXY", ""),
                "http_proxy": os.environ.get("JUJU_CHARM_HTTP_PROXY", ""),
                "no_proxy": os.environ.get("JUJU_CHARM_NO_PROXY", ""),
            }
```

If you confirmed that an empty `""` value breaks things, then perhaps add:
```python
            proxies = {
                "https_proxy": os.environ.get("JUJU_CHARM_HTTPS_PROXY"),
                "http_proxy": os.environ.get("JUJU_CHARM_HTTP_PROXY"),
                "no_proxy": os.environ.get("JUJU_CHARM_NO_PROXY"),
            }
            proxies = {k: v for k, v in proxies.items() if v is not None} or None
```

nrobinaubertin: The second would logically work (it needs to be `None` if we don't want to mask the usual env variables).  
But mypy wants `Optional[Dict[str, str]]` as an argument for the proxyHandler.  
So I could try to do this:
```
        proxies: Optional[Dict[str, str]] = {
            "https_proxy": os.environ.get("JUJU_CHARM_HTTPS_PROXY"),
            "http_proxy": os.environ.get("JUJU_CHARM_HTTP_PROXY"),
            "no_proxy": os.environ.get("JUJU_CHARM_NO_PROXY"),
        }
        if isinstance(proxies, dict):
            proxies = {k: v for k, v in proxies.items() if v is not None} or None

        proxy_handler = request.ProxyHandler(proxies)
```

But it doesn't work because here `proxies` is in fact `Optional[Dict[str, Optional[str]]]`. So mypy complains again (did I mentioned that I love mypy ?).  
At that point I find my proposal easier to read and understand.

nrobinaubertin: Just after writing that, I understood that I could use the idea from your first version to write it in a way that mypy accepts (using an empty string instead of None)

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2094,15 +2094,15 @@ def _download_and_push_promtail_to_workload(
...
-            proxies = None
+        # If no Juju proxy variable was set, we set proxies to None to let the ProxyHandler get
+        # the proxy env variables from the environment
+        proxies: Optional[Dict[str, str]] = {
+            "https_proxy": os.environ.get("JUJU_CHARM_HTTPS_PROXY", ""),
+            "http_proxy": os.environ.get("JUJU_CHARM_HTTP_PROXY", ""),
+            "no_proxy": os.environ.get("JUJU_CHARM_NO_PROXY", ""),
+        }
+        if isinstance(proxies, dict):
+            proxies = {k: v for k, v in proxies.items() if v != ""} or None
```
sed-i, line 2105:
```suggestion
        proxies = {k: v for k, v in proxies.items() if v} or None
```

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2094,15 +2094,15 @@ def _download_and_push_promtail_to_workload(
...
-            proxies.update({"http": os.environ["JUJU_CHARM_HTTP_PROXY"]})
-        if proxies and os.environ.get("JUJU_CHARM_HTTPS_PROXY"):
-            proxies.update({"https": os.environ["JUJU_CHARM_HTTPS_PROXY"]})
-        if proxies and os.environ.get("JUJU_CHARM_NO_PROXY"):
-            proxies.update({"no_proxy": os.environ["JUJU_CHARM_NO_PROXY"]})
-        else:
-            proxies = None
+        # If no Juju proxy variable was set, we set proxies to None to let the ProxyHandler get
+        # the proxy env variables from the environment
+        proxies: Optional[Dict[str, str]] = {
```
sed-i, line 2099: The dict is inferred, and Optional is technically wrong here.
```suggestion
        proxies = {
```

lib/charms/loki_k8s/v1/loki_push_api.py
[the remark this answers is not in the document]
nrobinaubertin: pyright seems indeed smarter than mypy.  
I made the changes you suggested. Thank you

Abuelodelanada (approved):
Looks good to me

sed-i (approved):
Thank you for this important fix!

PR #374, opened 2024-04-16
add support for promtail arm binary

lucabello:
## Issue

We need to build grafana-agent-k8s for ARM; to do so, the ARM promtail binary needs to be downloaded when deploying on that architecture.

## Solution

Download the ARM binary when the architecture is **arm64**.

## Testing Instructions
Please download the promtail files yourself from [here (v2.9.7)](https://github.com/canonical/loki-k8s-operator/releases/tag/promtail-v2.9.7) and run `sha256sum` on both the zip and the binary, to make sure the ones I pasted are correct.

### DIFF
--- a/lib/charms/loki_k8s/v1/loki_push_api.py
+++ b/lib/charms/loki_k8s/v1/loki_push_api.py
@@ -519,7 +519,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 8
+LIBPATCH = 9
 
 PYDEPS = ["cosl"]
 
--- a/lib/charms/loki_k8s/v1/loki_push_api.py
+++ b/lib/charms/loki_k8s/v1/loki_push_api.py
@@ -534,12 +534,17 @@ PROMTAIL_BASE_URL = "https://github.com/canonical/loki-k8s-operator/releases/dow
 # To update Promtail version you only need to change the PROMTAIL_VERSION and
 # update all sha256 sums in PROMTAIL_BINARIES. To support a new architecture
 # you only need to add a new key value pair for the architecture in PROMTAIL_BINARIES.
-PROMTAIL_VERSION = "v2.5.0"
+PROMTAIL_VERSION = "v2.9.7"
 PROMTAIL_BINARIES = {
     "amd64": {
         "filename": "promtail-static-amd64",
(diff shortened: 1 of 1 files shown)


IbraAoad (approved):
Validated [from (v2.9.7)](https://github.com/canonical/loki-k8s-operator/releases/tag/promtail-v2.9.7)

```
sha256sum *                                                                                                         
28da9b99f81296fe297831f3bc9d92aea43b4a92826b8ff04ba433b8cb92fb50  promtail-static-amd64
6873cbdabf23062aeefed6de5f00ff382710332af3ab90a48c253ea17e08f465  promtail-static-amd64.gz
4cd055c477a301c0bdfdbcea514e6e93f6df5d57425ce10ffc77f3e16fec1ddf  promtail-static-arm64
c083fdb45e5c794103f974eeb426489b4142438d9e10d0ae272b2aff886e249b  promtail-static-arm64.gz
```

Abuelodelanada (approved):

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -534,12 +534,17 @@ def _alert_rules_error(self, event):
...
+PROMTAIL_VERSION = "v2.9.7"
 PROMTAIL_BINARIES = {
     "amd64": {
         "filename": "promtail-static-amd64",
-        "zipsha": "543e333b0184e14015a42c3c9e9e66d2464aaa66eca48b29e185a6a18f67ab6d",
-        "binsha": "17e2e271e65f793a9fbe81eab887b941e9d680abe82d5a0602888c50f5e0cac9",
+        "zipsha": "6873cbdabf23062aeefed6de5f00ff382710332af3ab90a48c253ea17e08f465",
+        "binsha": "28da9b99f81296fe297831f3bc9d92aea43b4a92826b8ff04ba433b8cb92fb50",
+    },
+    "arm64": {
```
sed-i, line 544: How is this picked up? Note that sometimes arm64 is coded as aarch64.
We have code somewhere (gaganet?) that handles it.

lucabello: Good point, I've added that code section you mentioned (which was in cos-proxy-operator) here:

```python
if arch in ["x86_64", "amd64"]:
    self._arch = "amd64"
elif arch in ["aarch64", "arm64", "armv8b", "armv8l"]:
    self._arch = "arm64"
else:
    self._arch = arch
```

Issue #369, opened 2024-04-03
Allow configuration of loki


honghan-wong:
### Bug Description

With regards to bug #362, we would like to have a configurable loki where we can modify the retention or compactor.
We dont see any retention in the current loki deployment and would like to have option to set the compactor + retention.

The loki is running on v2.7.4
```bash
$ curl localhost:3100/loki/api/v1/status/buildinfo
{"version":"2.7.4","revision":"98421b0c0","branch":"HEAD","buildUser":"root@rockcraft-loki-785655","buildDate":"2023-03-26T12:09:41Z","goVersion":""}
```

https://grafana.com/docs/loki/v2.7.x/operations/storage/retention/#grafana-loki-storage-retention

### To Reproduce

See #362
```
  loki:
    charm: loki-k8s
    channel: latest/stable
    revision: 105
    resources:
      loki-image: 88
    scale: 1
    constraints: arch=amd64
    storage:
      active-index-directory: kubernetes,1,2048M
      loki-chunks: kubernetes,1,409600M
    trust: true
```

### Environment

running on microk8s + microceph

```
series: jammy
applications:
  microceph:
    charm: microceph
    channel: edge
    revision: 9
    num_units: 3
    constraints: arch=amd64
  microk8s:
    charm: microk8s
    channel: legacy/stable
    revision: 121
    num_units: 3
    options:
      addons: dns ingress rbac
      channel: latest/stable
      skip_verify: true
    constraints: arch=amd64
```

### Relevant log output

```shell
See #362
```

Issue #362, opened 2024-03-08
Loki silently rejects data when chunks dir is too large


sed-i:
### Bug Description

When the chunks dir contains too many files, loki stops ingesting new data. (Setting `tune2fs -O large_dir /dev/vda2` resolved it.)

- Should have an alert fired in that scenario. If loki doesn't have metrics for this then think about pebble checks/notices or node exporter.
- Need to think if there's anything generic we could modify with the chunks config.

### To Reproduce

See https://github.com/canonical/cos-proxy-operator/issues/130.

### Environment

See https://github.com/canonical/cos-proxy-operator/issues/130.

### Relevant log output

```shell
See https://github.com/canonical/cos-proxy-operator/issues/130.
```

### Additional context

Observed by @dnegreira.

dnegreira:
@sed-i maybe relevant: https://github.com/grafana/loki/issues/364

lathiat:
Some context from https://github.com/grafana/loki/issues/1502

Firstly, we need to be careful not to enable large_dir as a workaround on a /boot device (including where /boot is stored on /)

> As [@wschoot](https://github.com/wschoot) points out EXERCISE CAUTION before you run tune2fs -O large_dir /dev/... if the device uses grub to boot it will fail to boot on the next reboot. This is true for any system using < grub-2.12 (which was only released (Wed, 20 Dec 2023) and will take a while to rollout via the distros.

Secondly they moved away from this single large directory in v2.5, in these issues:
https://github.com/grafana/loki/issues/364

However to take advantage of that we need to use Schema v12, and looking at the charm we are possibly still using v11:
https://github.com/canonical/loki-k8s-operator/blob/6c26adf5a7c850e5ea520eee40f2adc5a2865f56/src/config_builder.py#L104

```
    @property
    def _schema_config(self) -> dict:
        return {
            "configs": [
                {
                    "from": "2020-10-24",
                    "index": {"period": "24h", "prefix": "index_"},
                    "object_store": "filesystem",
                    "schema": "v11",
                    "store": "boltdb",
                }
            ]
        }
```

mmkay:
We need to check if we can still reproduce this once we have compaction. We should likely do `tune2fs ...` in our charm if so.

IbraAoad:
To help narrow doing our options on this issue, here are some potential approaches we could explore

## Approach 1: Configuring ingester config
[...]

PR #377, opened 2024-04-17
Add Support for Time-Based Log Retention in Loki

### LINKED ISSUE #278: Expose retention tuning knobs
### Enhancement Proposal

Clouds come in various sizes, and customer requirements can vary wildly. If Loki is to become the default centralized logging solution for all charmed deployments I feel it would be important to allow operators to control their retention period and log storage consumption.

IbraAoad:
## Issue
Fixes #278 and might potentially be complementary to the fix of #369 & #362

This PR adds support for having `time` based log retention in Loki and it also points loki to use `"store": "boltdb-shipper"` instead of the [currently invalid option](https://grafana.com/docs/loki/v2.9.x/configure/#period_config) `"store": "boltdb"` which might help in fixing #362 since data compaction will be enabled

Testing has been conducted to validate the following:

- Deployments that were previously functioning without a compactor ("store": "boltdb") can safely transition to using boltdb-shipper without losing any old logs.
- Old logs are retained and properly compacted upon startup with the new configuration.

```
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.167917681Z caller=shipper.go:165 index-store=boltdb-shipper-2020-10-24 msg="starting index shipper in RW mode"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.167338945Z caller=table_manager.go:136 index-store=boltdb-shipper-2020-10-24 msg="uploading tables"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.168149888Z caller=table_manager.go:220 index-store=boltdb-shipper-2020-10-24 msg="found a legacy file index_19830, moving it to folder with same name"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.168235527Z caller=table_manager.go:240 index-store=boltdb-shipper-2020-10-24 msg="loading table index_19830"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.16856759Z caller=table.go:318 msg="handing over indexes to shipper index_19830"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.168603427Z caller=table.go:334 msg="finished handing over table index_19830"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.168626094Z caller=shipper_index_client.go:76 index-store=boltdb-shipper-2020-10-24 msg="starting boltdb shipper in RW mode"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.168728076Z caller=table_manager.go:171 index-store=boltdb-shipper-2020-10-24 msg="handing over indexes to shipper"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.168812945Z caller=table.go:318 msg="handing over indexes to shipper index_19830"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.168820671Z caller=table.go:334 msg="finished handing over table index_19830"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.180144521Z caller=module_service.go:82 msg=initialising module=compactor
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.180241431Z caller=ring.go:273 msg="ring doesn't exist in KV store yet"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.180314544Z caller=basic_lifecycler.go:297 msg="instance not found in the ring" instance=loki-0 ring=compactor
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.180391138Z caller=compactor.go:395 msg="waiting until compactor is JOINING in the ring"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.180879172Z caller=manager.go:995 user=fake msg="Starting rule manager..."
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.182992457Z caller=ingester.go:447 msg="recovered WAL checkpoint recovery finished" elapsed=2.881953ms errors=false
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.183016236Z caller=ingester.go:453 msg="recovering from WAL"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.184145693Z caller=ingester.go:469 msg="WAL segment recovery finished" elapsed=4.03537ms errors=false
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.184164644Z caller=ingester.go:417 msg="closing recoverer"
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.184173336Z caller=ingester.go:425 msg="WAL recovery finished" time=4.063365ms
2024-04-17T14:53:40.246Z [loki] level=info ts=2024-04-17T14:53:40.18419343Z caller=wal.go:156 msg=started component=wal
2024-04-17T14:53:40.346Z [loki] level=info ts=2024-04-17T14:53:40.314000526Z caller=compactor.go:399 msg="compactor is JOINING in the ring"
2024-04-17T14:53:40.346Z [loki] level=info ts=2024-04-17T14:53:40.314112138Z caller=compactor.go:409 msg="waiting until compactor is ACTIVE in the ring"
2024-04-17T14:53:40.446Z [loki] level=info ts=2024-04-17T14:53:40.425459505Z caller=compactor.go:413 msg="compactor is ACTIVE in the ring"
2024-04-17T14:53:40.446Z [loki] level=info ts=2024-04-17T14:53:40.425499846Z caller=loki.go:505 msg="Loki started"
2024-04-17T14:53:43.346Z [loki] level=info ts=2024-04-17T14:53:43.296600331Z caller=worker.go:209 msg="adding connection" addr=loki-0.loki-endpoints.cos.svc.cluster.local:9095
2024-04-17T14:53:43.346Z [loki] level=info ts=2024-04-17T14:53:43.297812733Z caller=scheduler.go:615 msg="this scheduler is in the ReplicationSet, will now accept requests."
2024-04-17T14:53:45.446Z [loki] level=info ts=2024-04-17T14:53:45.425633613Z caller=compactor.go:474 msg="this instance has been chosen to run the compactor, starting compactor"
2024-04-17T14:53:45.446Z [loki] level=info ts=2024-04-17T14:53:45.425675246Z caller=compactor.go:503 msg="waiting 10m0s for ring to stay stable and previous compactions to finish before starting compactor"
2024-04-17T14:54:40.245Z [loki] level=info ts=2024-04-17T14:54:40.168043675Z caller=table_manager.go:136 index-store=boltdb-shipper-2020-10-24 msg="uploading tables"
2024-04-17T14:54:40.245Z [loki] level=info ts=2024-04-17T14:54:40.168074092Z caller=index_set.go:86 msg="uploading table index_19830"
2024-04-17T14:54:40.245Z [loki] level=info ts=2024-04-17T14:54:40.169169813Z caller=table_manager.go:171 index-store=boltdb-shipper-2020-10-24 msg="handing over indexes to shipper"
2024-04-17T14:54:40.245Z [loki] level=info ts=2024-04-17T14:54:40.169240404Z caller=table.go:318 msg="handing over indexes to shipper index_19830"
2024-04-17T14:54:40.245Z [loki] level=info ts=2024-04-17T14:54:40.169247974Z caller=table.go:334 msg="finished handing over table index_19830"
2024-04-17T14:54:40.245Z [loki] level=info ts=2024-04-17T14:54:40.187572448Z caller=index_set.go:107 msg="finished uploading table index_19830"
2024-04-17T14:54:40.245Z [loki] level=info ts=2024-04-17T14:54:40.187608966Z caller=index_set.go:185 msg="cleaning up unwanted indexes from table index_19830"
```

## Solution
Exposing `retention_period` as a config option in the charm which will set a retention period on all streams with this duration; however, setting retention periods for individual streams is not currently supported and may be considered in future work.

## Testing Instructions
1. juju deploy cos-lite
2. juju deploy flog-k8s
3. juju relate loki flog-k8s
4. juju config loki retention-period=24
5. validate /etc/loki/loki-local-config.yaml and check Loki logs after 24h to valdidate log deletion 

## Important Considerations

- Since "boltdb-shipper" is also deprecated in [loki v3](https://grafana.com/docs/loki/latest/release-notes/v3-0/#deprecations) it might wise to look into changing the boltdb store to the [recommended tsdb store](https://grafana.com/docs/loki/latest/configure/storage/#tsdb-recommended) a separate issue can be opened for this to investigate the impact   (Issue Created #380)
- The Loki compactor [exposes various configurations](https://grafana.com/docs/loki/v2.9.x/configure/#compactor) related to retention, in addition to compaction itself. It could be beneficial to open an issue to investigate the possibility of having some of these opinionated/exposed as a charm configs. (Issue Created #381)


mmkay (approved):
LGTM as well.

src/charm.py
```diff
@@ -391,6 +391,11 @@ def _configure(self):  # noqa: C901
             self._stored.status["config"] = to_tuple(MaintenanceStatus("Configuring Loki"))
             return
 
+        if 0 < int(self.config["retention-period"]) < 24:
```
simskij, line 394: is this an upstream limitation? Could we use a duration string (i.e. `2d 4h`, `2M`, '4w' etc.) here instead to improve the UX for the operator?

simskij (changes requested):

src/charm.py
IbraAoad: So loki allows setting duration strings with the exception of `M`, but my initial thinking to this was to guard from any input errors for example people putting `1h` or `3600s` cause Loki itself doesn't really panic if value is less than `24h` and automatically defaults to `744h` and putting duration string validation in the charm itself didn't really seem like the best thing to do.

the upstream uses [Prometheus type duration](https://github.com/prometheus/common/blob/e54e4df4b9182e2fceac62aa447b507a53e9114d/model/time.go#L189)

```
var unitMap = map[string]struct {
	pos  int
	mult uint64
}{
	"ms": {7, uint64(time.Millisecond)},
	"s":  {6, uint64(time.Second)},
	"m":  {5, uint64(time.Minute)},
	"h":  {4, uint64(time.Hour)},
	"d":  {3, uint64(24 * time.Hour)},
	"w":  {2, uint64(7 * 24 * time.Hour)},
	"y":  {1, uint64(365 * 24 * time.Hour)},
}
```

IbraAoad: this however can be avoided and changed to string without putting the validation in the charm if we maybe just add it in the config docstrings that if any value is put less than `24h` workload won't panic yet won't honor it and will default to `744h`

IbraAoad: Also when I say validation inside the charm I mean smth like below

```
def duration_to_hours(duration_str):
    total_hours = 0
    matches = re.findall(r'(\d+)([a-zA-Z]+)', duration_str)
    conversion_factors = {
        "ms": 1 / (1000 * 3600), 
        "s": 1 / 3600,            
        "m": 1 / 60,               
        "h": 1,                   
        "d": 24,                  
        "w": 24 * 7, 
        "y": 24 * 365             
    }
    supported_types = ", ".join(conversion_factors.keys())
    for match in matches:
        print(match)
        value = float(match[0])
        unit = match[1].lower()
        if unit in conversion_factors:
            total_hours += value * conversion_factors[unit]
        else:
            return f"Unsupported duration type: {unit}. Supported types are: {supported_types}"
    return int(total_hours)

retention_period = "1y1w1d4h61m"
```

config.yaml
```diff
@@ -37,3 +37,11 @@ options:
...
     type: int
     default: 15
+  retention-period:
+    description: |
+      Sets a global retention period for log streams in Loki. A value of 0 disables retention (default).
+      Minimum retention period is 24h.
+      Specify the period in hours. For example, to set a 48-hour retention period, use `48`.
+      Specifying retention periods for individual streams is not currently supported.
+    type: int
+    default: 0
```
sed-i, line 47: I thought Loki's default is 30 days.
If so, then to remain consistent on upgrade, we would need to set if there.

src/charm.py
sed-i: An int can also be negative. To we need the `0 <`?

src/charm.py
```diff
@@ -391,6 +391,11 @@ def _configure(self):  # noqa: C901
...
+            self._stored.status["config"] = to_tuple(
```
sed-i: We already have a `config` used elsewhere. Strictly speaking, a section should be "edited" only within one context, to avoid unexpected overwrites. For example, you could:

```suggestion
            self._stored.status["retention"] = to_tuple(
```

src/charm.py
```diff
@@ -391,6 +391,11 @@ def _configure(self):  # noqa: C901
             self._stored.status["config"] = to_tuple(MaintenanceStatus("Configuring Loki"))
             return
```
sed-i: I would add a comment here like this:

> The config validity check does not `return` on error because if a lifecycle event
> comes in after a config change, we still want Loki to continue to function even
> with the invalid config.

src/config_builder.py
```diff
@@ -131,14 +136,23 @@ def _server(self) -> dict:
...
+                "shared_store": "filesystem",
+                "cache_location": BOLTDB_CACHE_DIR,
+            },
             "filesystem": {"directory": CHUNKS_DIR},
         }
 
     @property
     def _limits_config(self) -> dict:
         # Ref: https://grafana.com/docs/loki/latest/configure/#limits_config
+        if self.retention_period != 0:
```
sed-i, line 152: Add comment explaining the conversion?

tests/integration/helpers.py
```diff
@@ -62,6 +62,80 @@ async def loki_rules(ops_test, app_name) -> dict:
...
+    try:
+        response = requests.get(url)
+        if response.status_code == 200:
+            services = {}
+            # Parse the response and populate the services dictionary
+            # Each line represents a service name and its status separated by " => "
+            # We split each line by " => " and store the key-value pairs in the services dictionary
+            for line in response.text.split("\n"):
+                if line.strip():
+                    key, value = line.strip().split(" => ")
```
sed-i, line 102: Is there an API call we could use instead? This seems brittle.

tests/integration/test_loki_configs.py
```diff
@@ -0,0 +1,57 @@
...
+    logger.debug("deploy local charm")
+
+    await ops_test.model.deploy(
+        loki_charm, application_name=app_name, resources=resources, trust=True
+    )
+    await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=1000)
+    assert await is_loki_up(ops_test, app_name)
+
+    services = await loki_services(ops_test, app_name)
+    assert all(status == "Running" for status in services.values()), "Not all services are running"
```
sed-i, line 32: I wonder if there's a way we could simplify this. For example `assert "Not running" not in output`.

src/charm.py
IbraAoad: Kept as an `int` config yet switched the workload to switch it to days instead of hours

config.yaml
IbraAoad: default is that they live forever cause `retention_enabled` in compactor should be set to True, if set to true and no value is provided or set to `0` then 744h (30days) retention is used.

src/config_builder.py
IbraAoad: conversion actually seems redundant after another test, setting it to `0d` will yield same result so I removed the check

PR #392, opened 2024-04-29
charm logging lib

PietroPasotti:
Loki integration for charm logs.

waiting on https://github.com/canonical/cos-lib/pull/26

### DIFF
--- a/lib/charms/loki_k8s/v0/charm_logging.py
+++ b/lib/charms/loki_k8s/v0/charm_logging.py
@@ -53,8 +53,12 @@ The ``log_charm`` decorator will take these endpoints and set up the root logger
 logging module root logger) to forward all logs to these loki endpoints.
 
 ## TLS support
-If your charm integrates with a tls provider which is also trusted by Loki, you can configure TLS by
-passing to charm logging a `server_cert` parameter.
+If your charm integrates with a tls provider which is also trusted by the logs receiver, you can
+configure TLS by passing to charm logging a `server_cert` parameter.
+
+If you're not using the same CA as the loki-push-api endpoint you are sending logs to,
+you'll need to implement a cert-transfer relation to obtain the CA certificate from the same
+CA that Loki is using.
 
 ```
 @log_charm(loki_push_api_endpoint="my_logging_endpoint", server_cert="my_server_cert")
@@ -122,6 +126,10 @@ class InvalidEndpointError(CharmLoggingError):
     """Raised if an endpoint is invalid."""
 
 
+class InvalidEndpointsError(CharmLoggingError):
+    """Raised if an endpoint is invalid."""
+
+
 @contextmanager
 def charm_logging_disabled():
     """Contextmanager to temporarily disable charm logging.
@@ -162,14 +170,17 @@ def _get_logging_endpoints(
 
     errors = []
     sanitized_logging_endponts = []
-    for endpoint in logging_endpoints:
-        if isinstance(endpoint, str):
-            sanitized_logging_endponts.append(endpoint)
-        else:
-            errors.append(f"invalid endpoint: expected string, got {endpoint!r}")
+    if isinstance(logging_endpoints, str):
+        errors.append("invalid return value: expected Iterable[str], got str")
+    else:
+        for endpoint in logging_endpoints:
+            if isinstance(endpoint, str):
+                sanitized_logging_endponts.append(endpoint)
+            else:
+                errors.append(f"invalid endpoint: expected string, got {endpoint!r}")
 
     if errors:
-        logger.error(
+        raise InvalidEndpointsError(
             f"{charm}.{logging_endpoints_getter} should return an iterable of Loki push-api "
             "(-compatible) endpoints (strings); "
             f"ERRORS: {errors}"
@@ -186,9 +197,12 @@ def _get_server_cert(
     else:  # method or callable
         server_cert = server_cert_getter(self)
 
+    # we're assuming that the ca cert that signed this unit is the same that has signed loki's
     if server_cert is None:
+        logger.debug(f"{charm.__name__}.{server_cert_getter} returned None.")
         logger.warning(
-            f"{charm}.{server_cert_getter} returned None; sending logs over INSECURE connection."
+            "Charm logs are being sent over insecure http because a ca cert is "
+            "not provided to the charm_logging module."
         )
         return None
 
@@ -224,17 +238,7 @@ def _setup_root_logger_initializer(
             logger.debug("Charm logging DISABLED by env: skipping root logger initialization")
             return
 
-        try:
-            logging_endpoints = _get_logging_endpoints(logging_endpoints_getter, self, charm)
-        except Exception:
-            # if anything goes wrong with retrieving the endpoint, we go on with logging disabled.
-            # better than breaking the charm.
-            logger.exception(
-                f"exception retrieving the logging "
-                f"endpoint from {charm}.{logging_endpoints_getter}; "
-                f"proceeding with charm_logging DISABLED. "
-            )
-            return
+        logging_endpoints = _get_logging_endpoints(logging_endpoints_getter, self, charm)
 
         if not logging_endpoints:
             return
@@ -258,12 +262,9 @@ def _setup_root_logger_initializer(
                 labels=labels,
                 cert=str(server_cert) if server_cert else None,
             )
-
             root_logger.addHandler(handler)
-            root_logger.debug(
-                "Initialized charm logger",
-                extra={"tags": {"endpoint": url}},
-            )
+
+        logger.debug("Initialized LokiHandler and set up root logging for charm code.")
         return
 
     charm.__init__ = wrap_init
--- a/lib/charms/loki_k8s/v0/charm_logging.py
+++ b/lib/charms/loki_k8s/v0/charm_logging.py
@@ -53,39 +53,23 @@ class MyCharm(...):
         if self.tls_is_enabled():
             return "/path/to/my/server_cert.crt"
 ```
-
 """
-import copy
 import functools
-import json
 import logging
 import os
-import string
-import time
-import urllib.error
 from contextlib import contextmanager
-from logging.config import ConvertingDict
 from pathlib import Path
 from typing import (
-    Any,
     Callable,
-    Dict,
     Optional,
-    Tuple,
     Type,
     TypeVar,
     Union,
-    cast,
 )
-from urllib import request
 
-from cosl import JujuTopology
+from cosl import JujuTopology, LokiHandler
 from ops.charm import CharmBase
 from ops.framework import Framework
-
-# prevent infinite recursion because on failure urllib3 will push more logs
-# https://github.com/GreyZmeem/python-logging-loki/issues/18
-logging.getLogger("urllib3").setLevel(logging.INFO)
 
 # The unique Charmhub library identifier, never change it
 LIBID = "52ee6051f4e54aedaa60aa04134d1a6d"
@@ -102,140 +86,6 @@ PYDEPS = ["cosl"]
 logger = logging.getLogger("charm_logging")
 _GetterType = Union[Callable[[CharmBase], Optional[str]], property]
 CHARM_LOGGING_ENABLED = "CHARM_LOGGING_ENABLED"
-
-
-# from https://github.com/GreyZmeem/python-logging-loki, which seems to be dead
-class LokiEmitter:
-    """Base Loki emitter class."""
-
-    #: Success HTTP status code from Loki API.
-    success_response_code: int = 204
-
-    #: Label name indicating logging level.
-    level_tag: str = "severity"
-    #: Label name indicating logger name.
-    logger_tag: str = "logger"
-
-    #: String contains chars that can be used in label names in LogQL.
-    label_allowed_chars: str = "".join((string.ascii_letters, string.digits, "_"))
-    #: A list of pairs of characters to replace in the label name.
-    label_replace_with: Tuple[Tuple[str, str], ...] = (
-        ("'", ""),
-        ('"', ""),
-        (" ", "_"),
-        (".", "_"),
-        ("-", "_"),
-    )
-
-    def __init__(self, url: str, tags: Optional[dict] = None, cert: Optional[str] = None):
-        """Create new Loki emitter.
-
-        Arguments:
-            url: Endpoint used to send log entries to Loki (e.g. `https://my-loki-instance/loki/api/v1/push`).
-            tags: Default tags added to every log record.
-            cert: Absolute path to a ca cert for TLS authentication.
-
-        """
-        #: Tags that will be added to all records handled by this handler.
-        self.tags = tags or {}
-        #: Loki JSON push endpoint (e.g `http://127.0.0.1/loki/api/v1/push`)
-        self.url = url
-        #: Optional cert for TLS auth
-        self.cert = cert
-        #: only notify once on push failure, to avoid spamming error logs
-        self._error_notified_once = False
-
-    def __call__(self, record: logging.LogRecord, line: str):
-        """Send log record to Loki."""
-        payload = self.build_payload(record, line)
-        req = request.Request(self.url, method="POST")
-        req.add_header("Content-Type", "application/json; charset=utf-8")
-        jsondata_encoded = json.dumps(payload).encode("utf-8")
-
-        try:
-            resp = request.urlopen(req, jsondata_encoded, capath=self.cert)
-        except urllib.error.HTTPError as e:
-            if not self._error_notified_once:
-                logger.error(f"error pushing logs to {self.url}: {e.status, e.reason}")
-                self._error_notified_once = True
-            return
-
-        if resp.getcode() != self.success_response_code:
-            raise ValueError(
-                "Unexpected Loki API response status code: {0}".format(resp.status_code)
-            )
-
-    def build_payload(self, record: logging.LogRecord, line) -> dict:
-        """Build JSON payload with a log entry."""
-        labels = self.build_tags(record)
-        ns = 1e9
-        ts = str(int(time.time() * ns))
-        stream = {
-            "stream": labels,
-            "values": [[ts, line]],
-        }
-        return {"streams": [stream]}
-
-    @functools.lru_cache(256)
-    def format_label(self, label: str) -> str:
-        """Build label to match prometheus format.
-
-        `Label format <https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels>`_
-        """
-        for char_from, char_to in self.label_replace_with:
-            label = label.replace(char_from, char_to)
-        return "".join(char for char in label if char in self.label_allowed_chars)
-
-    def build_tags(self, record: logging.LogRecord) -> Dict[str, Any]:
-        """Return tags that must be send to Loki with a log record."""
-        tags = dict(self.tags) if isinstance(self.tags, ConvertingDict) else self.tags
-        tags = cast(Dict[str, Any], copy.deepcopy(tags))
-        tags[self.level_tag] = record.levelname.lower()
-        tags[self.logger_tag] = record.name
-
-        extra_tags = getattr(record, "tags", {})
-        if not isinstance(extra_tags, dict):
-            return tags
-
-        for tag_name, tag_value in extra_tags.items():
-            cleared_name = self.format_label(tag_name)
-            if cleared_name:
-                tags[cleared_name] = tag_value
-
-        return tags
-
-
-class LokiHandler(logging.Handler):
-    """Log handler that sends log records to Loki.
-
-    `Loki API <https://github.com/grafana/loki/blob/mas<#NOWOKE>ter/docs/api.md>`
-    """
-
-    def __init__(
-        self,
-        url: str,
-        tags: Optional[dict] = None,
-        # username, password tuple
-        cert: Optional[str] = None,
-    ):
-        """Create new Loki logging handler.
-
-        Arguments:
-            url: Endpoint used to send log entries to Loki (e.g. `https://my-loki-instance/loki/api/v1/push`).
-            tags: Default tags added to every log record.
-            cert: Optional absolute path to cert file for TLS auth.
-
-        """
-        super().__init__()
-        self.emitter = LokiEmitter(url, tags, cert)
-
-    def emit(self, record: logging.LogRecord):
-        """Send log record to Loki."""
-        # noinspection PyBroadException
-        try:
-            self.emitter(record, self.format(record))
-        except Exception:
-            self.handleError(record)
 
 
 def is_enabled() -> bool:
@@ -326,10 +176,10 @@ def _get_server_cert(server_cert_getter, self, charm):
 
 
 def _setup_root_logger_initializer(
-    charm: Type[CharmBase],
-    logging_endpoints_getter: _GetterType,
-    server_cert_getter: Optional[_GetterType],
-    service_name: Optional[str] = None,
+        charm: Type[CharmBase],
+        logging_endpoints_getter: _GetterType,
+        server_cert_getter: Optional[_GetterType],
+        service_name: Optional[str] = None,
 ):
     """Patch the charm's initializer and inject a call to set up root logging."""
     original_init = charm.__init__
@@ -389,9 +239,9 @@ def _setup_root_logger_initializer(
 
 
 def log_charm(
-    logging_endpoints: str,
-    server_cert: Optional[str] = None,
-    service_name: Optional[str] = None,
+        logging_endpoints: str,
+        server_cert: Optional[str] = None,
+        service_name: Optional[str] = None,
 ):
     """Set up the root logger to forward any charm logs to one or more Loki push API endpoints.
 
@@ -437,10 +287,10 @@ def log_charm(
 
 
 def _autoinstrument(
-    charm_type: Type[CharmBase],
-    logging_endpoints_getter: _GetterType,
-    server_cert_getter: Optional[_GetterType] = None,
-    service_name: Optional[str] = None,
+        charm_type: Type[CharmBase],
+        logging_endpoints_getter: _GetterType,
+        server_cert_getter: Optional[_GetterType] = None,
+        service_name: Optional[str] = None,
 ) -> Type[CharmBase]:
     """Set up logging on this charm class.
 
--- a/lib/charms/loki_k8s/v0/charm_logging.py
+++ b/lib/charms/loki_k8s/v0/charm_logging.py
@@ -33,27 +33,21 @@ the certificate file.
 """
 
 import functools
-import inspect
 import logging
 import os
 from contextlib import contextmanager
 from pathlib import Path
 from typing import (
-    Any,
     Callable,
-    Generator,
     Optional,
-    Sequence,
     Type,
     TypeVar,
     Union,
-    cast, List,
 )
 
 import logging_loki
 from ops.charm import CharmBase
 from ops.framework import Framework
-
 
 # The unique Charmhub library identifier, never change it
 LIBID = "52ee6051f4e54aedaa60aa04134d1a6d"
@@ -99,7 +93,6 @@ def charm_logging_disabled():
     os.environ[CHARM_LOGGING_ENABLED] = previous
 
 
-
 _C = TypeVar("_C", bound=Type[CharmBase])
 _T = TypeVar("_T", bound=type)
 _F = TypeVar("_F", bound=Type[Callable])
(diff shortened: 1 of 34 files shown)


PietroPasotti:
Note: this only works in combination with #394 
Or when https://github.com/canonical/traefik-k8s-operator/issues/347 is fixed.

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,463 @@
...
+            "stream": labels,
+            "values": [[ts, line]],
+        }
+        return {"streams": [stream]}
+
+    @property
+    def session(self) -> requests.Session:
+        """Create HTTP(s) session."""
+        if self._session is None:
+            self._session = requests.Session()
```
sed-i, line 145: Iirc, requests is a heavy dep to build from source.
Would it be easy to replace with urllib?

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,463 @@
...
+        # noinspection PyBroadException
+        try:
+            self.emitter(record, self.format(record))
+        except Exception:
+            self.handleError(record)
+
+
+def is_enabled() -> bool:
+    """Whether charm logging is enabled."""
+    return os.getenv(CHARM_LOGGING_ENABLED, "1") == "1"
```
sed-i, line 228: Charms have bursts of logs on event hooks. Should this be opt-in rather than opt-out?

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,463 @@
...
+
+class InvalidEndpointError(CharmLoggingError):
+    """Raised if an endpoint is invalid."""
+
+
+@contextmanager
+def charm_logging_disabled():
+    """Contextmanager to temporarily disable charm logging.
+
+    For usage in tests.
```
sed-i, line 243: Would it make sense to place fixtures in e.g. a conftest file next to charm_logging.py?

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,463 @@
...
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+PYDEPS = []
+
+logger = logging.getLogger("charm_logging")
```
sed-i, line 75: This is a neat capability, well done!
I wonder if we could simplify this:
In charm.py we always have `logger = logging.getLogger(__name__)`. Would it make sense / be possible to somehow update this logger, or create a mirror one, that sends to loki? I.e. without a decorator.

lib/charms/loki_k8s/v0/charm_logging.py
PietroPasotti: Mmmm yes except these fixtures are also for requirer charm testing.
For example, traefik uses charm_tracing_disabled to test its tracing requirer integration without sending any charm traces to the nonexistent tempo backend.
So it felt handy to ship them with the charm lib. Same pattern is present in `charms.tempo_k8s.v1.charm_testing.py`.

lib/charms/loki_k8s/v0/charm_logging.py
PietroPasotti: The thing is, and it's the same with `charm_tracing`, that the charm instance (not the type, nor the context) know what endpoint to send the data to.
So we first need to initialize the charm, then we need to ask the charm 'what endpoint do you want me to send telemetry to', and then we can send stuff.

We _could_ do all of this without the charm instance knowing about it, but we'd need to (hard)code plenty of raw hook tool calls to figure out what the endpoint is. 
For example say we're instrumenting traefik. We want to set up a logger to send logs to a loki instance. The endpoint is in the 'logging' endpoint relation databag. 
We'd need to, in charm.py:

```
charm_logging.start(
loki_endpoint = Popen("relation-get <ID of the relation on 'logging' if present>").from_json()["endpoint"]["url"]
)
```

And then we're not even considering TLS. It gets pretty complex to do this without the API the charm has access to. 
[Which is one more argument for separating ops API and charm API as far as I'm concerned.
](https://github.com/canonical/operator/issues/1075)

PietroPasotti: TLDR: that is possible but IMHO not desirable and too complex, It would require a redesign of charm tracing and charm logging and further speccing work. Not something for this PR.

lib/charms/loki_k8s/v0/charm_logging.py
PietroPasotti: same goes for traces. what is the concern exactly? 
It's an ephemeral process so it will naturally have activity when it runs otherwise no activity.
If it's opt-in, who would opt-in?

lib/charms/loki_k8s/v0/charm_logging.py
PietroPasotti: good point. If it's a little work I'll do this otherwise we can have a follow-up iteration to thin it down

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,463 @@
...
+    if errors:
+        if sanitized_logging_endponts:
+            logger.error(
+                f"{charm}.{logging_endpoints_getter} returned some invalid endpoint strings: {errors}"
+            )
+        else:
+            logger.error(
+                f"{charm}.{logging_endpoints_getter} should return an iterable of Loki push-api (compatible) endpoints (strings); "
+                f"got {errors} instead."
+            )
```
Abuelodelanada, line 287: This way is clearer and shorter:


```suggestion
        message = f"{charm}.{logging_endpoints_getter} should return an iterable of Loki push-api (compatible) endpoints (strings); got {errors} instead."

        if sanitized_logging_endponts:
            message = f"{charm}.{logging_endpoints_getter} returned some invalid endpoint strings: {errors}"

        logger.error(message)
```

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,463 @@
...
+
+        juju_topology = {
+            "juju_unit": self.unit.name,
+            "juju_application": self.app.name,
+            "juju_model": self.model.name,
+            "juju_model_uuid": self.model.uuid,
+            "service_name": service_name or self.app.name,
+            "charm_type_name": type(self).__name__,
+            "dispatch_path": os.getenv("JUJU_DISPATCH_PATH", ""),
+        }
```
Abuelodelanada, line 351: Why don't we use the `JujuTopology.from_charm()` object from cosl?

src/charm.py
```diff
@@ -717,10 +727,17 @@ def tracing_endpoint(self) -> Optional[str]:
         """Tempo endpoint for charm tracing."""
         return self.tracing.otlp_http_endpoint()
 
+    @property
+    def logging_endpoints(self) -> Optional[List[str]]:
+        """Loki endpoint for charm logging."""
+        if self._loki_container.get_service(self._name).current is ops.pebble.ServiceStatus.ACTIVE:
+            return [self._internal_url + "/loki/api/v1/push"]
```
Abuelodelanada, line 734: I would use a constant for `/loki/api/v1/push` value.

lib/charms/loki_k8s/v0/charm_logging.py
PietroPasotti: didn't want to assume the dependency or force other charms to install that. Or is it fine @simskij ?

simskij: Use PYDEPS to make sure it's there :)

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,463 @@
...
+2) add to your charm a "my_logging_endpoint" (you can name this attribute whatever you like) **property**
+that returns an http/https endpoint url. If you are using the `LogProxyConsumer` as
+`self.logging = LogProxyConsumer(self, ...)`, the implementation could be:
+
+```
+    @property
+    def my_logging_endpoint(self) -> List[str]:
+        '''Loki push API endpoints for charm logging'''
+        return self.logging.loki_endpoints:
+```
```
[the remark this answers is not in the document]
PietroPasotti: that feels like a good idea, but out of scope for this PR.

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,463 @@
...
+
+The ``log_charm`` decorator will take these endpoints and set up the root logger (as in python's
+logging module root logger) to forward all logs to these loki endpoints.
+
+3) If you were passing a certificate using `server_cert`, you need to change it to provide an *absolute* path to
+the certificate file.
```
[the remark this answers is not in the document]
PietroPasotti: refactored the section, it was a leftover from the charm logging doc

lib/charms/loki_k8s/v0/charm_logging.py
PietroPasotti: cleaned it up, thanks for the tip

lib/charms/loki_k8s/v0/charm_logging.py
PietroPasotti: fixed

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,470 @@
...
+2) add to your charm a "my_logging_endpoint" (you can name this attribute whatever you like) **property**
+that returns an http/https endpoint url. If you are using the `LogProxyConsumer` as
+`self.logging = LogProxyConsumer(self, ...)`, the implementation could be:
+
+```
+    @property
+    def my_logging_endpoint(self) -> List[str]:
+        '''Loki push API endpoints for charm logging'''
+        return self.logging.loki_endpoints:
+```
```
dstathis, line 26: Can we pass a relation name in instead of the endpoints? Then just get the endpoints from relation data. Seems like a better solution to me.

dstathis (approved):

src/charm.py
```diff
@@ -717,10 +729,17 @@ def tracing_endpoint(self) -> Optional[str]:
         """Tempo endpoint for charm tracing."""
         return self.tracing.otlp_http_endpoint()
 
+    @property
+    def logging_endpoints(self) -> Optional[List[str]]:
+        """Loki endpoint for charm logging."""
+        if self._loki_container.get_service(self._name).current is ops.pebble.ServiceStatus.ACTIVE:
+            return ["http://localhost:3100" + self._loki_push_api_endpoint]
```
mmkay, line 790: Will it work when TLS is enabled?

lib/charms/loki_k8s/v0/charm_logging.py
PietroPasotti: We made the same decision for charm tracing. It's a matter of separation of concerns: we don't want the charm logging lib to make assumptions about the relation interface. Also take a look at loki's own usage: in loki's case, we don't get the endpoint(s) from relation data, we simply return `"http://localhost:3100/api/v1/push"`.

The library shouldn't have to worry about databag parsing.

src/charm.py
```diff
@@ -108,12 +110,15 @@ def to_status(tpl: Tuple[str, str]) -> StatusBase:
         MetricsEndpointProvider,
     ],
 )
+@log_charm(logging_endpoints="logging_endpoints", server_cert="server_ca_cert_path")
```
sed-i, line 113: In [loki docs](https://grafana.com/docs/loki/latest/send-data/), they list a few loki clients in python, and two of them work directly with the logging module:

- https://pypi.org/project/python-logging-loki/
- https://pypi.org/project/nextlog/

Could we simplify by a call to `logger.addHandler(...)` from the class's `__init__`?

PietroPasotti: I tried out python-logging-loki and
1) it wasn't working out of the box, also it missed tls integration capability
2) it was using requests

so I ended up copying most of the code from it and fixing what was broken/missing. Instead of adding an external dependency I'd rather vendor what we need in the charm lib, since it's not much code anyway.

PietroPasotti: Secondly, which class are you referring to? the charm? 
Just like in charm tracing, we want to set this up before the charm's init runs, so that we can also capture more ops logs and avoid any races in init

lib/charms/loki_k8s/v0/charm_logging.py
PietroPasotti: another example: we'd also have to automatically detect if the relation you're passing is to a grafana-agent instead of a loki, and fetch the push api endpoint in a different way.

Abuelodelanada (dismissed):
LGTM.
Just update lib and fix linting

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,466 @@
...
+# See LICENSE file for licensing details.
+
+"""This charm library contains utilities to automatically forward your charm logs to a loki-push-api endpoint.
+
+(yes! charm code, not workload code!)
+
+This means that, if your charm is related to, for example, COS' Loki charm (or a Grafana Agent),
+you will be able to inspect in real time from the Grafana dashboard the logs emitted by your charm.
+
+To start using this library, you need to do two things:
```
mmkay, line 12: I think that the dependence on `loki_push_api` could be more emphasized. Maybe something like:

1) if your charm isn't already related to Loki using any of the consumers/forwarders from `loki_push_api` library, you need to fetch the library:

`charmcraft fetch-lib charms.loki_k8s.v1.loki_push_api`

and add the logging consumer that matches your use case. See https://charmhub.io/loki-k8s/libraries/loki_push_apihttps://charmhub.io/loki-k8s/libraries/loki_push_api for more information.

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,466 @@
...
+        # username, password tuple
+        cert: Optional[str] = None,
+    ):
+        """Create new Loki logging handler.
+
+        Arguments:
+            url: Endpoint used to send log entries to Loki (e.g. `https://my-loki-instance/loki/api/v1/push`).
+            tags: Default tags added to every log record.
+
+            # FIXME: Session expects a .pem file it says
```
mmkay, line 213: Anything still needs to be done here?

mmkay (commented):
A few nits on the documentation part, can't wait to use it in action :D

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,466 @@
...
+
+2) add to your charm a "my_logging_endpoint" (you can name this attribute whatever you like) **property**
+that returns an http/https endpoint url. If you are using the `LogProxyConsumer` as
+`self.logging = LogProxyConsumer(self, ...)`, the implementation could be:
+
+```
+    @property
+    def my_logging_endpoint(self) -> List[str]:
+        '''Loki push API endpoints for charm logging'''
+        return self.logging.loki_endpoints:
```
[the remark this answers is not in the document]
PietroPasotti: the push api consumer doesn't expose any kind of `is_ready()` API, but it seems that the `loki_endpoints` method is quite tolerant and will simply return an empty list on all 'unhappy paths', and never raise. 
So we should be safe. But I agree it's kind of implicit.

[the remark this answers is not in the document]
PietroPasotti: added a comment to clarify this point

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,466 @@
...
+        """Send log record to Loki."""
+        payload = self.build_payload(record, line)
+        req = request.Request(self.url, method="POST")
+        req.add_header("Content-Type", "application/json; charset=utf-8")
+        jsondata_encoded = json.dumps(payload).encode("utf-8")
+
+        try:
+            resp = request.urlopen(req, jsondata_encoded, capath=self.cert)
+        except urllib.error.HTTPError as e:
+            logger.error(f"error pushing logs to {self.url}: {e.status, e.reason}")
```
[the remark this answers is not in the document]
PietroPasotti: mmm true, we'd get an error logline for each logline. Will add a little bit of state to the emitter to only log that once.

mmkay (approved):

lib/charms/loki_k8s/v0/charm_logging.py
dstathis: I don't understand. We know exactly what the interface looks like. It is always loki_push_api. Grafana-agent uses the exact same interface. We don't need to care at all whether it is grafana-agent or loki on the other end.

src/charm.py
PietroPasotti: resolved in conversations: moved lokihandler and related code to cosl in https://github.com/canonical/cos-lib/pull/26

lib/charms/loki_k8s/v0/charm_logging.py
PietroPasotti: resolved in face to face convo, agreed that the charm lib can and should remain interface agnostic

PietroPasotti:
Waiting for cosl to release 0.0.12

src/charm.py
```diff
@@ -528,7 +538,8 @@ def _configure(self):  # noqa: C901
             scheme="https" if self._certs_on_disk else "http", port=self._port
         )
         self.metrics_provider.update_scrape_job_spec(self.scrape_jobs)
-        self.grafana_source_provider.update_source(source_url=self._external_url)
+        # self.grafana_source_provider.update_source(source_url=self._external_url)
+        # self.grafana_source_provider.update_source(source_url=self.hostname)
```
sed-i, line 542: Is this intentional?

src/charm.py
```diff
@@ -770,11 +782,23 @@ def tracing_endpoint(self) -> Optional[str]:
             return self.tracing.get_endpoint("otlp_http")
         return None
 
+    @property
+    def logging_endpoints(self) -> Optional[List[str]]:
+        """Loki endpoint for charm logging."""
+        container = self._loki_container
+        if (
+            container.can_connect()
+            and container.get_service(self._name).current is ops.pebble.ServiceStatus.ACTIVE
```
sed-i, line 791: - Would it be easier to use [`is_running`](https://github.com/canonical/operator/blob/5582dbbb3713f9e51077d7ab8553adc15c626b0b/ops/pebble.py#L1049)?
- This is an interesting condition. Other charms using charm_logging can't do this check.
What would happen if a loki relation is in place but loki hasn't started/finished refresh yet?

sed-i (dismissed):

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,327 @@
...
+    yield
+    os.environ[CHARM_LOGGING_ENABLED] = previous
+
+
+_C = TypeVar("_C", bound=Type[CharmBase])
+_T = TypeVar("_T", bound=type)
+_F = TypeVar("_F", bound=Type[Callable])
+
+
+def _get_logging_endpoints(logging_endpoints_getter, self, charm):
```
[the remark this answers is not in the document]
PietroPasotti: yup, there's a _GetterType above. Will use that.

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,327 @@
...
+    if isinstance(logging_endpoints_getter, property):
+        logging_endpoints = logging_endpoints_getter.__get__(self)
+    else:  # method or callable
+        logging_endpoints = logging_endpoints_getter(self)
+
+    if logging_endpoints is None:
+        logger.debug(
+            f"{charm}.{logging_endpoints_getter} returned None; quietly disabling "
+            f"charm_logging for the run."
```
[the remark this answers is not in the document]
PietroPasotti: I think mentioning the hook name is not necessary, there'll be a log above telling you what event we're processing.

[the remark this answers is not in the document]
PietroPasotti: simplified the message though.

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,327 @@
...
+            sanitized_logging_endponts.append(endpoint)
+        else:
+            errors.append(f"invalid endpoint: expected string, got {endpoint!r}")
+
+    if errors:
+        logger.error(
+            f"{charm}.{logging_endpoints_getter} should return an iterable of Loki push-api "
+            "(-compatible) endpoints (strings); "
+            f"ERRORS: {errors}"
+        )
```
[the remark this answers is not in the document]
PietroPasotti: mmm, my main concern here is to make sure the charm_logging lib raises as little as possible during runtime. Given it executes on each event before the charm has had a chance to process the event, if this lib is broken then the charm is super broken. 
It feels risky.
Unittests should use `caplog` and filter for this error instead.

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,327 @@
...
+
+    return sanitized_logging_endponts
+
+
+def _get_server_cert(server_cert_getter, self, charm):
```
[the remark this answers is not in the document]
PietroPasotti: done, good point

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,327 @@
...
+    if isinstance(server_cert_getter, property):
+        server_cert = server_cert_getter.__get__(self)
+    else:  # method or callable
+        server_cert = server_cert_getter(self)
+
+    if server_cert is None:
+        logger.warning(
+            f"{charm}.{server_cert_getter} returned None; sending logs over INSECURE connection."
```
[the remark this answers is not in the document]
PietroPasotti: do you have a proposal? I'm not sure what'd be helpful for a cloud admin who's got no clue what charm tracing is and how to solve this :/

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,327 @@
...
+        )
+        return None
+
+    if not Path(server_cert).is_absolute():
+        raise ValueError(
+            f"{charm}.{server_cert_getter} should return a valid tls cert absolute path (string | Path)); "
+            f"got {server_cert} instead."
+        )
```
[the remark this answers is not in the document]
PietroPasotti: I don't recall why I was strict in requiring an absolute path, I guess I didn't want to worry about the CWD. but assuming the charm will do that for me, I'm fine with raising if the file is not found.

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,327 @@
...
+):
+    """Patch the charm's initializer and inject a call to set up root logging."""
+    original_init = charm.__init__
+
+    @functools.wraps(original_init)
+    def wrap_init(self: CharmBase, framework: Framework, *args, **kwargs):
+        original_init(self, framework, *args, **kwargs)
+
+        if not is_enabled():
+            logger.info("Charm logging DISABLED by env: skipping root logger initialization")
```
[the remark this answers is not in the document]
PietroPasotti: I'm fine with downgrading to debug, but I think it's an important one

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,327 @@
...
+
+        if not logging_endpoints:
+            return
+
+        juju_topology = JujuTopology.from_charm(self)
+        tags = {
+            **juju_topology.as_dict(),
+            "service_name": service_name or self.app.name,
+            "charm_type_name": type(self).__name__,
+            "dispatch_path": os.getenv("JUJU_DISPATCH_PATH", ""),
```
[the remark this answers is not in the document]
PietroPasotti: 1) done.

2) `service_name` is a standard label that's picked up by grafana and has [a special meaning](https://community.grafana.com/t/how-to-get-value-of-service-name/113534), so we can't change that.
charm_type_name we could.

3) +1 for hook_name instead of dispatch_path

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,327 @@
...
+        }
+        server_cert: Optional[Union[str, Path]] = (
+            _get_server_cert(server_cert_getter, self, charm) if server_cert_getter else None
+        )
```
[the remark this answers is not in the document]
PietroPasotti: atm `_get_server_cert` can't take a None, and I'd rather keep it that way, it's simpler.

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,327 @@
...
+            )
+        return
+
+    charm.__init__ = wrap_init
+
+
+def log_charm(
+    logging_endpoints: str,
+    server_cert: Optional[str] = None,
+    service_name: Optional[str] = None,
```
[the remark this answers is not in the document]
PietroPasotti: it's not the method itself,  it's the name of a method (so a string).
can't use the method itself because we're forward-referencing it (it's a decorator).

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,327 @@
...
+            logging_endpoints_getter=getattr(charm_type, logging_endpoints),
+            server_cert_getter=getattr(charm_type, server_cert) if server_cert else None,
+            service_name=service_name,
+        )
+        return charm_type
+
+    return _decorator
+
+
+def _autoinstrument(
```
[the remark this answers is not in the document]
PietroPasotti: it's not public. It's useful in testing, and in the future, if we want to allow

```
if __name__=="__main__":
    charm_tracing.autoinstrument(MyCharm,
        loki_endpoints=MyCharm.get_loki_endpoints
        )
    ops.main(MyCharm)
```
it's just an underscore away.

src/charm.py
PietroPasotti: huh, no. good catch.

src/charm.py
PietroPasotti: - TIL about is_running()!
- it would fail to push logs. Not much we can do about that, same as with all other push models in cos

lib/charms/loki_k8s/v0/charm_logging.py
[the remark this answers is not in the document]
PietroPasotti: do you agree or have a counterproposal?

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,327 @@
...
+        for url in logging_endpoints:
+            handler = LokiHandler(
+                url=url,
+                tags=tags,
+                cert=str(server_cert) if server_cert else None,
+            )
+
+            root_logger.addHandler(handler)
+            root_logger.debug(
+                "Initialized charm logger",
```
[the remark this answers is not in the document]
PietroPasotti: let's have a chat about this 'for admin eyes', I'm not sure I see how charm logging differs from regular 'for developer eyes' logging.

lib/charms/loki_k8s/v0/charm_logging.py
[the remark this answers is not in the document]
PietroPasotti: agreed: we are more strict in validation and raise.

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,360 @@
...
+        logging_endpoints = _get_logging_endpoints(logging_endpoints_getter, self, charm)
+
+        if not logging_endpoints:
+            return
+
+        juju_topology = JujuTopology.from_charm(self)
+        labels = {
+            **juju_topology.as_dict(),
+            "service_name": service_name or self.app.name,
+            "charm_type": type(self).__name__,
```
sed-i, line 250: Why do we need charm type if we already have charm name from the topology class?
If the class name changes across an upgrade, data will be fragmented.
Happens rarely, but changes in metadata `.name` are even rarer.

PietroPasotti: +1

lib/charms/loki_k8s/v0/charm_logging.py
```diff
@@ -0,0 +1,360 @@
...
+See https://charmhub.io/loki-k8s/libraries/loki_push_apihttps://charmhub.io/loki-k8s/libraries/loki_push_api
+for more information.
+
+Once your charm is related to, for example, COS' Loki charm (or a Grafana Agent),
+you will be able to inspect in real time from the Grafana dashboard the logs emitted by your charm.
+
+## Labels
+
+The library will inject the following labels into the records sent to grafana:
+- ``model``: name of the juju model this charm is deployed to
```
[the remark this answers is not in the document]
PietroPasotti: I think monospacing them makes it more clear that they're not arbitrary strings but technical tags

Issue #380, opened 2024-04-19
Switch to TSDB instead of boltdb-shipper


IbraAoad:
### Enhancement Proposal

Currently we're relying on [boltdb](https://github.com/canonical/loki-k8s-operator/blob/fc9d4b90260aeef2408582bcb32d91c9fc5040ed/src/config_builder.py#L135) for storing index, this will be [deprecated](https://grafana.com/docs/loki/latest/configure/storage/#tsdb-recommended) once we upgrade to Loki V3.X, we need to start looking into the risks and benefits of migrating to TSDB.

### Key Considerations
Primary focus should be on the potential challenges related to data migration between the two storage solutions, this should be thoroughly tested. 
[ref to migration](https://grafana.com/docs/loki/latest/setup/migrate/migrate-to-tsdb/)

PR #393, opened 2024-04-30
Adding TSBD storage to Loki and switching to v12 schema

### LINKED ISSUE #362: Loki silently rejects data when chunks dir is too large

IbraAoad:
## Issue
Fixes #362 & #380

## Solution

To avoid this becoming a breaking change in loki a few workarounds were implemented

1. TSDB data will live inside the active-index-directory originally created to accommodate boltdb, this is because adding a new PVC for it might disrupt the upgrade path as juju doesn't currently support this, this adjustment won't lead to storage overload, as all chunks, regardless of their schema, will still be directed to `/loki/chunks`
```
ERROR Juju on containers does not support updating storage on a statefulset.
The new charm's metadata contains updated storage declarations.
You'll need to deploy a new charm rather than upgrading if you need this change.
```
2. V12 schema migration date (The date for when loki will actually start to write logs in v12) will be set dynamically, with different values depending on whether it's a fresh installation or a charm upgrade. In case of a fresh installation loki will set the date to the day of installation and will set it to the day after in case of upgrade. This is achieved by dumping the loki config in the chunks persistent storage `/loki/chunks` so that next time we go through startup/upgrade we check if the file is there and render the v12 "from" to be today/tomorrow

## Context
Refer to https://github.com/canonical/loki-k8s-operator/issues/362#issuecomment-2074912569

## Upgrade Notes

**Storage Schema Bump** 
This PR bumps the Loki's storage schema to v12, which manages the `filesystem` storage far better than v11, the filesystem will use `<user>/<fingerprint>/base64Encode(<start>:<end>:<checksum>)` instead of dumping everything in the flat `chunks` directory.

**Migration to TSDB**
Starting with Loki v2.8, TSDB is the recommended Loki index. It is heavily inspired by the Prometheus’s TSDB [sub-project](https://github.com/prometheus/prometheus/tree/main/tsdb). For a deeper explanation you can read Loki maintainer Owen’s [blog post](https://lokidex.com/posts/tsdb/). The short version is that this new index is more efficient, faster, and more scalable. It also resides in object storage like the [boltdb-shipper](https://grafana.com/docs/loki/latest/operations/storage/boltdb-shipper/) index which preceded it.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -21,7 +21,7 @@ import subprocess
 import time
 import urllib.request
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple, TypedDict, cast
+from typing import Any, Dict, List, Optional, Tuple, TypedDict, Union, cast
 from urllib.error import HTTPError, URLError
 from urllib.parse import urlparse
 
@@ -50,7 +50,7 @@ from config_builder import (
     HTTP_LISTEN_PORT,
     KEY_FILE,
     LOKI_CONFIG,
-    LOKI_CONFIG_PERSISTED,
+    LOKI_CONFIG_BACKUP,
     RULES_DIR,
     ConfigBuilder,
 )
@@ -208,7 +208,7 @@ class LokiOperatorCharm(CharmBase):
             path=f"{external_url.path}/loki/api/v1/push",
         )
 
-        self.v12_migration_date = self._get_v12_migration_date()
+        self.v12_migration_date = self._get_v12_migration_date
 
         self.dashboard_provider = GrafanaDashboardProvider(self)
 
@@ -242,11 +242,13 @@ class LokiOperatorCharm(CharmBase):
             event.add_status(to_status(status))
 
     def _on_config_changed(self, _):
+        # If migration date isn't present then this is a fresh loki installation, hence set the v12 schema date to today
         if not self.v12_migration_date:
             self.v12_migration_date = (datetime.date.today()).strftime("%Y-%m-%d")
         self._configure()
 
     def _on_upgrade_charm(self, _):
+        # If migration date isn't present on upgrade, set the v12 schema to tomorrow to avoid losing today's logs written in v11
         if not self.v12_migration_date:
             self.v12_migration_date = (
                 datetime.date.today() + datetime.timedelta(days=1)
@@ -517,7 +519,7 @@ class LokiOperatorCharm(CharmBase):
         if self._running_config() != config:
             config_as_yaml = yaml.safe_dump(config)
             self._loki_container.push(LOKI_CONFIG, config_as_yaml, make_dirs=True)
-            self._loki_container.push(LOKI_CONFIG_PERSISTED, config_as_yaml, make_dirs=True)
+            self._loki_container.push(LOKI_CONFIG_BACKUP, config_as_yaml, make_dirs=True)
             logger.info("Pushed new configuration")
             return True
 
@@ -578,13 +580,21 @@ class LokiOperatorCharm(CharmBase):
 
         return ",".join(alertmanagers)
 
+    @property
     def _get_v12_migration_date(self) -> str:
-        """Get the 'from' date from the v12 schema in Loki config."""
+        """Get the 'from' date from the v12 schema in Loki config.
+
+        In 2024-05 we changed the storage backend the charm configures from v11/boltdb to v12/tsdb.
+
+        Returns:
+        str: If v12 schema is found, returns the 'from' date of the v12 schema
+                        in ISO 8601 format (YYYY-MM-DD), otherwise returns empty string.
+        """
         if not self._loki_container.can_connect():
             return ""
         try:
             running_config = yaml.safe_load(
-                self._loki_container.pull(LOKI_CONFIG_PERSISTED, encoding="utf-8").read()
+                self._loki_container.pull(LOKI_CONFIG_BACKUP, encoding="utf-8").read()
             )
         except PathError:
             return ""
--- a/src/charm.py
+++ b/src/charm.py
@@ -11,6 +11,7 @@ develop a new k8s charm using the Operator Framework:
 
     https://discourse.charmhub.io/t/4208
 """
+import datetime
 import logging
 import os
 import re
@@ -406,6 +407,9 @@ class LokiOperatorCharm(CharmBase):
         current_layer = self._container.get_plan()
         new_layer = self._build_pebble_layer
         restart = current_layer.services != new_layer.services
+        v12_migration = self._get_v12_from_date() or (
+            datetime.date.today() + datetime.timedelta(days=1)
+        ).strftime("%Y-%m-%d")
 
         config = ConfigBuilder(
             instance_addr=self.hostname,
@@ -415,6 +419,7 @@ class LokiOperatorCharm(CharmBase):
             ingestion_burst_size_mb=int(self.config["ingestion-burst-size-mb"]),
             retention_period=int(self.config["retention-period"]),
             http_tls=(self.server_cert.server_cert is not None),
+            v12_from=v12_migration,
         ).build()
 
         # At this point we're already after the can_connect guard, so if the following pebble operations fail, better
@@ -507,6 +512,15 @@ class LokiOperatorCharm(CharmBase):
             return alerting_config
 
         return ",".join(alertmanagers)
+
+    def _get_v12_from_date(self) -> str:
+        """Get the 'from' date from the v12 schema in Loki config."""
+        running_config = self._running_config()
+        if running_config:
+            for config in running_config.get("schema_config", {}).get("configs", []):
+                if config.get("schema") == "v12":
+                    return config.get("from", "")
+        return ""
 
     def _running_config(self) -> Dict[str, Any]:
         """Get the on-disk Loki config."""
--- a/src/charm.py
+++ b/src/charm.py
@@ -21,7 +21,7 @@ import subprocess
 import time
 import urllib.request
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple, TypedDict, Union, cast
+from typing import Any, Dict, List, Optional, Tuple, TypedDict, cast
 from urllib.error import HTTPError, URLError
 from urllib.parse import urlparse
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -129,6 +129,7 @@ class LokiOperatorCharm(CharmBase):
                 retention=to_tuple(ActiveStatus()),
             )
         )
+        self._stored.set_default(is_upgrade=False)
 
         self._loki_container = self.unit.get_container(self._name)
         self._node_exporter_container = self.unit.get_container("node-exporter")
@@ -243,7 +244,8 @@ class LokiOperatorCharm(CharmBase):
         self._configure()
 
     def _on_upgrade_charm(self, _):
-        self._configure(False)
+        self._stored.is_upgrade = True
+        self._configure()
 
     def _on_server_cert_changed(self, _):
         self._update_cert()
@@ -435,7 +437,7 @@ class LokiOperatorCharm(CharmBase):
     ##############################################
     #             UTILITY METHODS                #
     ##############################################
-    def _configure(self, is_fresh_installation: bool = True):  # noqa: C901
+    def _configure(self):  # noqa: C901
         """Configure Loki charm."""
         # "is_ready" is a racy check, so we do it once here (instead of in collect-status)
         if self.resources_patch.is_ready():
@@ -467,25 +469,24 @@ class LokiOperatorCharm(CharmBase):
         new_layer = self._build_pebble_layer
         restart = current_layer.services != new_layer.services
 
-        v12_migration_date = self._v12_migration_date
-
         # If v12_migration_date isn't set (due to missing or failed retrieval),
         # we determine the migration date for v12 schema. This occurs once
-        # during initial setup, as subsequent hooks persist the info.
+        # during initial setup, as subsequent hooks will get the value from the persisted backup config.
 
         # If it's a fresh Loki installation, it's safe to set the v12 schema date to today.
         # This ensures that logs are correctly managed in v12 from the beginning of Loki's operation.
         # If it's an upgrade scenario, we set the date to tomorrow to accommodate today's logs
         # that might have been written in the previous v11 schema format.
 
-        # By default, we assume it's a fresh installation unless explicitly set to False
+        # By default, we assume it's a fresh installation unless state explicitly set to True
         # by the upgrade hook, indicating an upgrade
-        if not v12_migration_date:
-            v12_migration_date = (datetime.date.today() + datetime.timedelta(days=1)).strftime(
+
+        if not (v12_migration_date := self.get_v12_migration_date_from_backup()):
+            today = datetime.date.today()
+            tomorrow = today + datetime.timedelta(days=1)
+            v12_migration_date = (tomorrow if self._stored.is_upgrade else today).strftime(
                 "%Y-%m-%d"
             )
-            if is_fresh_installation:
-                v12_migration_date = datetime.date.today().strftime("%Y-%m-%d")
 
         config = ConfigBuilder(
             instance_addr=self.hostname,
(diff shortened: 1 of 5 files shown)


src/config_builder.py
```diff
@@ -16,9 +16,12 @@
 
 LOKI_DIR = "/loki"
 CHUNKS_DIR = os.path.join(LOKI_DIR, "chunks")
+LOKI_CONFIG_PERSISTED = os.path.join(CHUNKS_DIR, "loki-local-config.yaml")
```
sed-i, line 19: Should we make it more obvious that this is a backup, for reference only?

```python
# Path to a persisted config backup, for reference purposes.
# This is needed to e.g. determine the storage backend used before an upgrade too place.
LOKI_CONFIG_BACKUP = os.path.join(CHUNKS_DIR, "loki-local-config.yaml.bak")
```

---

Off-topic:
This variable isn't used by the config builder but relies on some dir being persisted in metadata.yaml.
We should probably refactor these "consts" in a future PR.

src/config_builder.py
```diff
@@ -16,9 +16,12 @@
...
 COMPACTOR_DIR = os.path.join(LOKI_DIR, "compactor")
 BOLTDB_DIR = os.path.join(LOKI_DIR, "boltdb-shipper-active")
 BOLTDB_CACHE_DIR = os.path.join(LOKI_DIR, "boltdb-shipper-cache")
+TSDB_DIR = os.path.join(BOLTDB_DIR, "tsdb-index")
```
sed-i, line 23: As part of this PR, wdyt about adding something along the lines of the following comment to metadata.yaml?

```diff
    mounts:
+     # We do not need separate storages. TODO: In the next breaking change for Loki,
+     # switch to having just one persisted storate, e.g. `/loki/persisted`.
      - storage: active-index-directory
        location: /loki/boltdb-shipper-active
      - storage: loki-chunks
        location: /loki/chunks
```

src/config_builder.py
```diff
@@ -143,6 +155,11 @@ def _storage_config(self) -> dict:
                 "shared_store": "filesystem",
                 "cache_location": BOLTDB_CACHE_DIR,
             },
+            "tsdb_shipper": {
+                "active_index_directory": TSDB_DIR,
+                "shared_store": "filesystem",
+                "cache_location": TSDB_CACHE_DIR,
+            },
```
sed-i, line 162: Should we add a dedicated integration test?
For example:
- Name: `test_boltdb_v11_to_tsdb_v12_migration.py`
- Have a literal bundle that deploys the current loki by revision number (not channel), then upgrade to the current loki.
- Then upgrade again to the current loki.
- Then upgrade again, but override the resource to an old oci-image from before v12 (if this is a use case we want to address).

sed-i (commented):
Nice work @IbraAoad!

src/config_builder.py
```diff
@@ -112,24 +118,28 @@ def _ruler(self) -> dict:
...
+            {
+                "from": "2020-10-24",
+                "index": {"period": "24h", "prefix": "index_"},
+                "object_store": "filesystem",
+                "schema": "v11",
+                "store": "boltdb-shipper",
+            }
+        ]
+
+        if self.v12_migration_date:
```
sed-i, line 131: Rendering it conditionally is a nice touch!
Add a comment? Iiuc:
```suggestion
        # Only render the v12/tsdb section when we can reliably obtain a "from" date.
        # This is needed to avoid code ordering issues during upgrade, related to "can_connect".
        if self.v12_migration_date:
```

src/config_builder.py
```diff
@@ -148,20 +158,23 @@ def _server(self) -> dict:
...
             "boltdb_shipper": {
                 "active_index_directory": BOLTDB_DIR,
                 "shared_store": "filesystem",
                 "cache_location": BOLTDB_CACHE_DIR,
             },
-            "tsdb_shipper": {
+            "filesystem": {"directory": CHUNKS_DIR},
+        }
+
+        if self.v12_migration_date:
```
sed-i, line 170: Does this need to be rendered conditionally? I.e. does loki care that there's an unused section?

tests/integration/test_boltdb_v11_to_tsdb_v12_migration.py
```diff
@@ -0,0 +1,113 @@
...
+    tsdb_shipper_config = post_upgrade_config["storage_config"]["tsdb_shipper"]
+    tsdb_schema_configs = post_upgrade_config["schema_config"]["configs"][1]
+
+    expected_from = (
+        (datetime.date.today()).strftime("%Y-%m-%d")
+        if is_fresh_installation
+        else (datetime.date.today() + datetime.timedelta(days=1)).strftime("%Y-%m-%d")
+    )
+
+    assert all(
```
sed-i, line 103: Iirc, in a test failure it would be easier to figure out what's wrong if we use individual assert statements.

tests/integration/test_boltdb_v11_to_tsdb_v12_migration.py
```diff
@@ -0,0 +1,113 @@
...
+        [
+            tsdb_shipper_config["active_index_directory"]
+            == "/loki/boltdb-shipper-active/tsdb-index",
+            tsdb_shipper_config["shared_store"] == "filesystem",
+            tsdb_shipper_config["cache_location"] == "/loki/tsdb-cache",
+            tsdb_schema_configs["store"] == "tsdb",
+            tsdb_schema_configs["schema"] == "v12",
+            tsdb_schema_configs["from"] == expected_from,
```
sed-i, line 111: Do we need all these assertions?
At first glance, seems like keeping "tsdb", "v12", "expected_from" would make the test less brittle.

tests/integration/test_boltdb_v11_to_tsdb_v12_migration.py
```diff
@@ -0,0 +1,113 @@
...
+    assert await is_loki_up(ops_test, app_name)
+    post_upgrade_config = await loki_config(ops_test, app_name)
+    tsdb_shipper_config = post_upgrade_config["storage_config"]["tsdb_shipper"]
+    tsdb_schema_configs = post_upgrade_config["schema_config"]["configs"][1]
+
+    expected_from = (
+        (datetime.date.today()).strftime("%Y-%m-%d")
+        if is_fresh_installation
+        else (datetime.date.today() + datetime.timedelta(days=1)).strftime("%Y-%m-%d")
+    )
```
sed-i, line 101: Hopefully this itest never runs across midnight :sweat_smile:

src/charm.py
```diff
@@ -578,13 +580,21 @@ def _alerting_config(self) -> str:
...
+    @property
     def _get_v12_migration_date(self) -> str:
-        """Get the 'from' date from the v12 schema in Loki config."""
+        """Get the 'from' date from the v12 schema in Loki config.
+
+        In 2024-05 we changed the storage backend the charm configures from v11/boltdb to v12/tsdb.
+
+        Returns:
+        str: If v12 schema is found, returns the 'from' date of the v12 schema
+                        in ISO 8601 format (YYYY-MM-DD), otherwise returns empty string.
```
sed-i, line 591:
```suggestion
        Returns:
            The 'from' date of the v12 schema, in YYYY-MM-DD format (ISO 8601), if it is found; otherwise empty string.
```

src/charm.py
```diff
@@ -578,13 +580,21 @@ def _alerting_config(self) -> str:
...
+        """
         if not self._loki_container.can_connect():
             return ""
         try:
             running_config = yaml.safe_load(
-                self._loki_container.pull(LOKI_CONFIG_PERSISTED, encoding="utf-8").read()
+                self._loki_container.pull(LOKI_CONFIG_BACKUP, encoding="utf-8").read()
             )
         except PathError:
```
sed-i, line 599: We may end up needing more exception types here: sometimes we get pebble errors after a can_connect guard. I think ConnectionError, APIError and maybe another one.

But a yaml parsing error we probably want to have the charm in error state?

src/config_builder.py
```diff
@@ -112,24 +118,28 @@ def _ruler(self) -> dict:
...
-                },
+        configs = [
+            {
+                "from": "2020-10-24",
+                "index": {"period": "24h", "prefix": "index_"},
+                "object_store": "filesystem",
+                "schema": "v11",
+                "store": "boltdb-shipper",
+            }
+        ]
```
sed-i, line 129: Side question: would it be easier to always render tsdb, and have boltdb rendered conditionally?

src/charm.py
```diff
@@ -208,7 +208,7 @@ def __init__(self, *args):
             path=f"{external_url.path}/loki/api/v1/push",
         )
 
-        self.v12_migration_date = self._get_v12_migration_date()
+        self.v12_migration_date = self._get_v12_migration_date
```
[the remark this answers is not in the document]
IbraAoad: I've marked some of the other comments as argument below is relevant to all

The reason for keeping this property separate is to determine the value based on the specific hook that fired. The `self._get_v12_migration_date()` job is to get the date from the persisted file. If the file isn't there (i.e., during an upgrade/install), then the event listeners will take care of setting the value. Dropping this in favor of 1 property can be achieved but felt rather complex as we won't have knowledge inside the `self._get_v12_migration_date()` of which hook caused it to execute which will leave us with a couple of options

- always set the migration date to tomorrow (This will lead to complications later on when boltdb is completely removed)
- query the loki api for how many chunks are inserted, 0 chunks meaning this is a fresh install and we can set the TSDB date as today, now the problem with this approach is that it will involve some unnecessary string manipulation for the api response and will enforce us to guard against HTTP and connection errors.

src/config_builder.py
IbraAoad: Loki itself wouldn't care but it will create some files in this dir that will be redundant with no actual schema

tests/integration/test_boltdb_v11_to_tsdb_v12_migration.py
IbraAoad: tbh I'd rather have it fail if it runs across midnight than have a dedicated check just for this 😅😅

src/charm.py
IbraAoad: so most of these are already taken care of within the [can_connect](https://github.com/canonical/operator/blob/6cbc264a3bba177a34d93e8422e7f9c0807a0961/ops/model.py#L2132) guard, the parsing error doesn't seem recoverable tho so it makes sense to have it in error state

src/config_builder.py
IbraAoad: I would rather drop it off completely in the next breaking change, just to avoid on having these many conditions that depend on life-cycle events

src/charm.py
sed-i: That is true but there is a subtle difference: it is only the `can_connect` call that is inside the try-except - the code that follows inside your if block is not. `can_connect` is a point-in-time check, i.e. if you call it twice, the second may return false. So the code after a can_connect guard can still fail for pebble operations. In in our experience it does (transient startup error; mentioned [here](https://discourse.charmhub.io/t/its-probably-ok-for-a-unit-to-go-into-error-state/13022)). 

Based on the current semantics of this function, it seems you'd want to return `""` on pebble.ConnectionError, etc.

Zooming out a bit, I wonder if returning `""` on error is the best way to approach this.

src/charm.py
```diff
@@ -238,9 +242,17 @@ def _on_collect_unit_status(self, event: CollectStatusEvent):
             event.add_status(to_status(status))
 
     def _on_config_changed(self, _):
+        # If migration date isn't present then this is a fresh loki installation, hence set the v12 schema date to today
+        if not self.v12_migration_date:
+            self.v12_migration_date = (datetime.date.today()).strftime("%Y-%m-%d")
         self._configure()
 
     def _on_upgrade_charm(self, _):
```
sed-i, line 250: In the [upgrade sequence](https://discourse.charmhub.io/t/event-upgrade-charm/6485), upgrade-charm always fires before config-changed. However, afaik, there are no guarantees about `can_connect` at this point. This means that the following scenario would be possible:

1. upgrade a charm that already uses v12/tsdb.
2. (stop)
3. upgrade-charm fires before pebble-ready, so `self.v12_migration_date` is `""`. Hook changes it to tomorrow. 
    -  `_configure` is called.
    - (meanwhile, the container becomes available and the next `can_connect` is true)
    - a config file with "tsdb from tomorrow" is written to disk
4. config-changed fires, doesn't modify `self.v12_migration_date`
5. loki uses boltdb on a database that is already tsdb.

If you agree with the above, then one possible solution is:
1. Move the logic under `_configure`, after the can_connect guard.
2. If the backup exists, but our file doesn't, then it's an upgrade. If neither exist then it's a fresh install.

Wdyt?

tests/integration/test_boltdb_v11_to_tsdb_v12_migration.py
```diff
@@ -0,0 +1,103 @@
...
+@pytest.mark.abort_on_fail
+async def test_deploy_from_charmhub_v11_and_upgrade_to_v12_locally(ops_test: OpsTest, loki_charm):
+    """Deploy from Charmhub (v11 schema) and upgrade to v12 locally."""
+    await deploy_charm_from_charmhub_v11(ops_test, V11_APP_NAME)
+    await upgrade_charm_with_local_charm_v12(ops_test, V11_APP_NAME, loki_charm)
+    await verify_upgrade_success(ops_test, V11_APP_NAME, False)
+
+    # Here we upgrade again to ensure config is persisted and won't be overwritten with any weird values
+    await upgrade_charm_with_local_charm_v12(ops_test, V11_APP_NAME, loki_charm)
+    await verify_upgrade_success(ops_test, V11_APP_NAME, False)
```
Abuelodelanada, line 46: Are these test going to be valid once this PR is merged and a brand new `edge` version released?

IbraAoad: The test is locked on deploying rev 140 from charmhub, so it should yes.

src/config_builder.py
sed-i: So `v12_migration_date` is not an optional arg.
Could you please take another look at this function? Perhaps there a better way to structure this?

src/charm.py
```diff
@@ -241,6 +244,7 @@ def _on_config_changed(self, _):
...
+        self._stored.is_upgrade = True
```
sed-i, line 247: If we don't reset this stored value to False on stop, then its name `is_upgrade` is misleading, because what it really tells you is "has undergone an upgrade in the past".

Would it make sense to call it `_stored.fresh_install` instead?

src/charm.py
```diff
@@ -127,6 +129,7 @@ def __init__(self, *args):
                 retention=to_tuple(ActiveStatus()),
             )
         )
+        self._stored.set_default(is_upgrade=False)
```
sed-i, line 132: Wdyt about adding a comment such as:
```suggestion
        self._stored.set_default(is_upgrade=False)  # Relies on controller-backed storage
```

src/charm.py
```diff
@@ -465,6 +469,25 @@ def _configure(self):  # noqa: C901
...
+        # If it's an upgrade scenario, we set the date to tomorrow to accommodate today's logs
+        # that might have been written in the previous v11 schema format.
+
+        # By default, we assume it's a fresh installation unless state explicitly set to True
+        # by the upgrade hook, indicating an upgrade
+
+        if not (v12_migration_date := self.get_v12_migration_date_from_backup()):
+            today = datetime.date.today()
+            tomorrow = today + datetime.timedelta(days=1)
+            v12_migration_date = (tomorrow if self._stored.is_upgrade else today).strftime(
```
sed-i, line 487: Just though of a scenario we probably don't want to address, but could document in a comment?

If the pod is rescheduled during the startup sequence (e.g. due to manual kubectl delete or resource limit patch) - would we get the `upgrade-charm` event too early?

sed-i (commented):
Good job, lgtm, only doc-level questions this time :)

lucabello (approved):

sed-i (approved):

PR #405, opened 2024-05-28
[WIP] Charm logging tracing integration

PietroPasotti:
## Logs --> traces

If a charm is using `charm_tracing` and `charm_logging` (and both are active and healthy), we should be able to link traces and logs generated by each charm execution. 

This PR links logs to traces by:
- tagging all logs with the trace ID obtained from the context
- configuring the loki datasource to link each log tagged with a trace ID to the appropriate tempo trace in grafana

Open questions/challenges
- the `charm_logging` decorator needs to be applied **before** the `charm_tracing` one, otherwise logging will be unaware of the tracing context and fail to acknowledge the root span id.

### DIFF
--- a/lib/charms/loki_k8s/v0/charm_logging.py
+++ b/lib/charms/loki_k8s/v0/charm_logging.py
@@ -33,27 +33,21 @@ the certificate file.
 """
 
 import functools
-import inspect
 import logging
 import os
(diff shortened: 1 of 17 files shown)


lucabello:
@PietroPasotti is this still relevant?

PR #412, opened 2024-05-31
Add deprecation warning for LogProxyConsumer

sed-i:
## Issue
We have an external dependency on promtail, which the charm lib (LogProxyConsumer) attempts to download from a hard-coded URL. This dependency is inconvenient to work around with in air-gapped environments.

## Solution
Encourage users to migrate to LogForwarder by deprecating LogProxyConsumer.
1. LogProxyConsumer will be omitted from v2 of loki_push_api, but v0 and v1 are still available. Users who are happy with having the promtail dependency, can keep using it without change, and future changes to the loki lib won't break backwards compatibility as far as it comes to promtail.
1. No deprecation warnings are added in this PR, to avoid unnecessary (and inaccurate) messaging.

Closes https://github.com/canonical/grafana-agent-k8s-operator/issues/300.

## Context
- Pebble log forwarding support was added in Juju 3.4.1.
- https://discourse.charmhub.io/t/promtail-based-log-streaming-to-loki-logproxyconsumer-is-scheduled-for-deprecation/14289

Y-statement: 
- I the context of streaming logs to loki,
- facing the concern that promtail isn't readily available in air-gapped environments,
- we decided to go forward only with LogForwarder, requiring Juju >= 3.4.1,
- and rejected potential solutions for making promtail available via PYDEPS or a fileserver,
- to achieve a coherent, uniform solution without external dependencies that relies on pebble log forwarding,
- accepting the downside that logging to loki won't be available in k8s models with Juju < 3.4.1.

## Testing Instructions
See example in https://github.com/canonical/postgresql-k8s-operator/pull/486.

## Upgrade Notes
If, in a model with Juju < 3.4.1, an admin refreshes a charm that migrated from LogProxyConsumer to LogForwarder, then logs won't be forwarded anymore. (There will be no effect if there aren't any logging relations to loki in place.)

### DIFF
--- a/lib/charms/loki_k8s/v1/loki_push_api.py
+++ b/lib/charms/loki_k8s/v1/loki_push_api.py
@@ -17,15 +17,14 @@ This library
 send log to Loki by implementing the consumer side of the `loki_push_api` relation interface.
 For instance, a Promtail or Grafana agent charm which needs to send logs to Loki.
 
-- `LogProxyConsumer`: SCHEDULED FOR DEPRECATION.
+- `LogProxyConsumer`: DEPRECATED.
 This object can be used by any Charmed Operator which needs to send telemetry, such as logs, to
 Loki through a Log Proxy by implementing the consumer side of the `loki_push_api` relation
 interface.
 In order to be able to control the labels on the logs pushed this object adds a Pebble layer
 that runs Promtail in the workload container, injecting Juju topology labels into the
 logs on the fly.
-This object is scheduled for deprecation with the release and adoption of Juju 3.6 LTS.
-Consider migrating to LogForwarder.
+This object is deprecated. Consider migrating to LogForwarder with the release of Juju 3.6 LTS.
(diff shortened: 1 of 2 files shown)


lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -11,25 +11,30 @@
 - `LokiPushApiProvider`: This object is meant to be used by any Charmed Operator that needs to
 implement the provider side of the `loki_push_api` relation interface. For instance, a Loki charm.
 The provider side of the relation represents the server side, to which logs are being pushed.
+This library
```
lucabello, line 14:
```suggestion
```

lucabello (approved):

lucabello:
Not entirely sure, but this might need a LIBPATCH bump if you want it to be released.

dstathis (approved):

PR #419, opened 2024-06-14
Stop using controller storage

sed-i:
## Issue
- Controller storage is unreliable: in a model migration we once saw `ops.storage.NoSnapshotError: no snapshot data found for LokiOperatorCharm/LokiPushApiProvider[logging]/on/loki_push_api_alert_rules_changed[250784] object`, which broke the upgrade.
- Controller storage is [deprecated in ops](https://github.com/canonical/operator/pull/882).

## Solution
Stop using controller storage.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -128,9 +128,9 @@ class LokiOperatorCharm(CharmBase):
                 config=to_tuple(ActiveStatus()),
                 rules=to_tuple(ActiveStatus()),
                 retention=to_tuple(ActiveStatus()),
-            )
+            ),
+            fresh_install=True,
         )
-        self._stored.set_default(fresh_install=True)  # Relies on controller-backed storage
 
         self._loki_container = self.unit.get_container(self._name)
         self._node_exporter_container = self.unit.get_container("node-exporter")
@@ -793,6 +793,4 @@ class LokiOperatorCharm(CharmBase):
 
 
 if __name__ == "__main__":
-    # We need use_juju_for_storage=True because ingress_per_unit
-    # requires it to keep track of the ingress address.
-    main(LokiOperatorCharm, use_juju_for_storage=True)
+    main(LokiOperatorCharm)


src/charm.py
```diff
@@ -128,9 +128,9 @@ def __init__(self, *args):
                 config=to_tuple(ActiveStatus()),
                 rules=to_tuple(ActiveStatus()),
                 retention=to_tuple(ActiveStatus()),
-            )
+            ),
+            fresh_install=True,
         )
-        self._stored.set_default(fresh_install=True)  # Relies on controller-backed storage
```
sed-i, line 134: @IbraAoad: we initially said this relies on controller storage, but now that I see we set `fresh_install` it to False and the config backup is saved to persistent storage, it seems that controller storage isn't necessary. Unless I'm missing something?

src/charm.py
```diff
@@ -793,6 +793,4 @@ def server_ca_cert_path(self) -> Optional[str]:
 
 
 if __name__ == "__main__":
-    # We need use_juju_for_storage=True because ingress_per_unit
-    # requires it to keep track of the ingress address.
-    main(LokiOperatorCharm, use_juju_for_storage=True)
```
sed-i, line 798: @PietroPasotti @dstathis Our other IPU charm, prometheus, doesn't use controller storage.
And in ingress_per_unit v1, the only use of stored state is for current_urls, which seems like we don't need to rely on controller storage. Wdyt?

PietroPasotti: I'd tend to agree. worst that can happen is you get two ready events instead of one, and given the hot reloading this won't even trigger a restart.

src/charm.py
```diff
@@ -128,9 +128,9 @@ def __init__(self, *args):
                 config=to_tuple(ActiveStatus()),
                 rules=to_tuple(ActiveStatus()),
                 retention=to_tuple(ActiveStatus()),
-            )
+            ),
+            fresh_install=True,
```
PietroPasotti, line 132: it's not clear to me what this does, and why this change was necessary. I realize this is hard to test as we'd need an integration test to redeploy the charm, or simulate pod churn? 
But at least let's add a detailed comment to explain what's going on.

PietroPasotti (changes requested):
we need a little bit more comments/ documentation to explain to our future selves what this is all about

src/charm.py
PietroPasotti: if you do think this is testable though, please let's write a test

lucabello (approved):

src/charm.py
sed-i: It was introduced in #393 and there's an itest.

src/charm.py
ca-scribner: yeah I think I agree with this.  Persisting data between pods due to upgrade, churn, etc, shouldn't be needed.  This flag I think is only meaningful within a given pod's lifecycle.  It says whether this pod started its life in an install or upgrade, and I think that's all that's necessary

src/charm.py
ca-scribner: Yeah this code is from a previous commit, but like @PietroPasotti I found it tough to understand the importance at first.  Maybe a comment or link to an issue explaining it could help.  Although it does get better explanation in the relevant part of `configure` further down

ca-scribner (approved):

src/charm.py
sed-i: Yep, @IbraAoad documented this in two places next to where it's used:
- https://github.com/canonical/loki-k8s-operator/blob/fd3e58407ef55c5150ac5b54bf42dd4fb8a7dadc/src/charm.py#L492
- https://github.com/canonical/loki-k8s-operator/blob/fd3e58407ef55c5150ac5b54bf42dd4fb8a7dadc/src/charm.py#L252

lucabello:
Failing integration tests after the latest commit :(

sed-i:
> Failing integration tests after the latest commit :(

Got rate limited

```
unknown container reason "ImagePullBackOff": Back-off pulling image "docker.io/jujusolutions/charm-base:ubuntu-20.04"
```

Issue #390, opened 2024-04-26
Pebble ConnectionError on logging relation joined


gruyaume:
### Bug Description

We are getting recurrent issues in our integration tests from the `loki_push_api` v1 library. 

CI failure example:
- https://github.com/canonical/sdcore-udr-k8s-operator/actions/runs/8849637877/job/24302465822#step:7:206

### To Reproduce

1. Deploy charm that uses the `loki_push_api` lib
2. Deploy grafana agent
3. Integrate the two charms

### Environment

- lib version: 1.9
- juju version: 3.4.2

### Relevant log output

```shell
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-sdcore-udr-k8s-0/charm/./src/charm.py", line 639, in <module>
    main(UDROperatorCharm)
  File "/var/lib/juju/agents/unit-sdcore-udr-k8s-0/charm/venv/ops/main.py", line 544, in main
    manager.run()
  File "/var/lib/juju/agents/unit-sdcore-udr-k8s-0/charm/venv/ops/main.py", line 520, in run
    self._emit()
  File "/var/lib/juju/agents/unit-sdcore-udr-k8s-0/charm/venv/ops/main.py", line 509, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name)
  File "/var/lib/juju/agents/unit-sdcore-udr-k8s-0/charm/venv/ops/main.py", line 143, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-sdcore-udr-k8s-0/charm/venv/ops/framework.py", line 352, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-sdcore-udr-k8s-0/charm/venv/ops/framework.py", line 851, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-sdcore-udr-k8s-0/charm/venv/ops/framework.py", line 941, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-sdcore-udr-k8s-0/charm/lib/charms/loki_k8s/v1/loki_push_api.py", line 2550, in _update_logging
    self._update_endpoints(container, loki_endpoints)
  File "/var/lib/juju/agents/unit-sdcore-udr-k8s-0/charm/lib/charms/loki_k8s/v1/loki_push_api.py", line 2562, in _update_endpoints
[...]

PR #433, opened 2024-07-17
Improve exceptions handling and add can_connect guards

Abuelodelanada:
## Issue

This PR add some missing `can_connect()` guards in loki_push_api lib, and improves exception handling in Loki charm.

See #390

## Testing Instructions

- Run the test suite (unit, scenario, integration) and verify everything is green.
- Alternatively you can manually:
  - `juju add-model apps`
  - `juju deploy grafana-agent-k8s agent --channel latest/edge`
  - `juju deploy sdcore-udr-k8s --channel 1.5/edge`
  - Copy the new version of the lib into the sdcore-udr-k8s repo
  - Sync repo with running charm: `jhack utils sync sdcore-udr-k8s/0`
  - Relate both charms: `juju relate sdcore-udr-k8s:logging agent:logging-provider`
  - Check there are no errors like [these ones ](https://github.com/canonical/loki-k8s-operator/issues/390) in juju debug-log like

### DIFF
--- a/lib/charms/loki_k8s/v1/loki_push_api.py
+++ b/lib/charms/loki_k8s/v1/loki_push_api.py
@@ -2562,10 +2562,9 @@ class LogForwarder(ConsumerBase):
             return
 
         for container in self._charm.unit.containers.values():
-            if not container.can_connect():
-                continue
-
-            self._update_endpoints(container, loki_endpoints)
+            if container.can_connect():
+                self._update_endpoints(container, loki_endpoints)
+            # else: `_update_endpoints` will be called on pebble-ready anyway.
 
     def _retrieve_endpoints_from_relation(self) -> dict:
         loki_endpoints = {}
--- a/lib/charms/loki_k8s/v1/loki_push_api.py
+++ b/lib/charms/loki_k8s/v1/loki_push_api.py
@@ -527,7 +527,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 11
+LIBPATCH = 12
 
 PYDEPS = ["cosl"]
 
@@ -2475,6 +2475,9 @@ class _PebbleLogClient:
         container: Container, active_endpoints: Dict[str, str], topology: JujuTopology
     ):
         """Disable forwarding for inactive endpoints by checking against the Pebble plan."""
+        if not container.can_connect():
+            return
+
         pebble_layer = container.get_plan().to_dict().get("log-targets", None)
         if not pebble_layer:
             return
@@ -2501,6 +2504,9 @@ class _PebbleLogClient:
         container: Container, active_endpoints: Dict[str, str], topology: JujuTopology
     ):
         """Enable forwarding for the specified Loki endpoints."""
+        if not container.can_connect():
+            return
+
         layer = Layer(
             {  # pyright: ignore
                 "log-targets": _PebbleLogClient._build_log_targets(
--- a/lib/charms/loki_k8s/v1/loki_push_api.py
+++ b/lib/charms/loki_k8s/v1/loki_push_api.py
@@ -2475,9 +2475,6 @@ class _PebbleLogClient:
         container: Container, active_endpoints: Dict[str, str], topology: JujuTopology
     ):
         """Disable forwarding for inactive endpoints by checking against the Pebble plan."""
-        if not container.can_connect():
-            return
-
         pebble_layer = container.get_plan().to_dict().get("log-targets", None)
         if not pebble_layer:
             return
@@ -2504,9 +2501,6 @@ class _PebbleLogClient:
         container: Container, active_endpoints: Dict[str, str], topology: JujuTopology
     ):
         """Enable forwarding for the specified Loki endpoints."""
-        if not container.can_connect():
-            return
-
         layer = Layer(
             {  # pyright: ignore
                 "log-targets": _PebbleLogClient._build_log_targets(
@@ -2568,6 +2562,9 @@ class LogForwarder(ConsumerBase):
             return
 
         for container in self._charm.unit.containers.values():
+            if not container.can_connect():
+                continue
+
             self._update_endpoints(container, loki_endpoints)
 
     def _retrieve_endpoints_from_relation(self) -> dict:
--- a/src/charm.py
+++ b/src/charm.py
@@ -66,7 +66,7 @@ from ops.model import (
     StatusBase,
     WaitingStatus,
 )
-from ops.pebble import ConnectionError, Error, ExecError, Layer, PathError, ProtocolError
+from ops.pebble import Error, Layer, PathError, ProtocolError
 
 # To keep a tidy debug-log, we suppress some DEBUG/INFO logs from some imported libs,
 # even when charm logging is set to a lower level.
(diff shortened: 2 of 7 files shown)


lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2501,6 +2504,9 @@ def enable_endpoints(
         container: Container, active_endpoints: Dict[str, str], topology: JujuTopology
     ):
         """Enable forwarding for the specified Loki endpoints."""
+        if not container.can_connect():
+            return
+
```
dstathis, line 2509: If we return here and skip this method, are we sure the method will run again after pebble is ready?

src/charm.py
```diff
@@ -589,21 +593,36 @@ def _update_cert(self):
...
+
+    def _remove_cert_files(self):
+        ca_cert_path = Path(self._ca_cert_path)
+        try:
             self._loki_container.remove_path(CERT_FILE, recursive=True)
             self._loki_container.remove_path(KEY_FILE, recursive=True)
             self._loki_container.remove_path(ca_cert_path, recursive=True)
+        except ConnectionError:
+            logger.error("Could not remove cert files. Pebble refused connection. Shutting down?")
+            return
```
dstathis, line 613: I think we should not return here but surface the error. If we are shutting down, then it doesn't really matter because the charm will go away soon. If we are not shutting down though, these files will stay on disk forever.

src/charm.py
```diff
@@ -589,21 +593,36 @@ def _update_cert(self):
...
-        self._loki_container.exec(["update-ca-certificates", "--fresh"]).wait()
-        subprocess.run(["update-ca-certificates", "--fresh"])
+    def _update_ca_certificates(self):
+        try:
+            self._loki_container.exec(["update-ca-certificates", "--fresh"]).wait()
+            subprocess.run(["update-ca-certificates", "--fresh"])
+        except (ConnectionError, ExecError) as e:
+            logger.error("Could not run update-ca-certificates in workload container. %s", e)
+        except subprocess.SubprocessError as e:
+            logger.error("Could not run update-ca-certificates in charm container. %s", e)
```
dstathis, line 625: This feels like we are just sweeping errors under the rug. Isn't it a problem if the `update-ca-certificates` command fails?

lib/charms/loki_k8s/v1/loki_push_api.py
Abuelodelanada: Yes, because this method: `_PebbleLogClient.enable_endpoints()` is called from `enable_endpoints()` and this method is called from: `_update_logging()` and `_on_pebble_ready()`

src/charm.py
Abuelodelanada: We can avoid returning here and let the workflow continue.
This exception is raised if pebble is not ready. 

We execute `_update_cert()` which conditionally executes `_remove_cert_files()` on:

- `__init__` method
- `server_cert.on.cert_changed` event handler
- `_configure` common exit hook method

In order to avoid this files stay on disk forever, we can execute this method before anything else:

```diff
     def _update_cert(self):
+        self._remove_cert_files()
+
         if self.server_cert.available:
             self._push_cert_files()
-        else:
-            self._remove_cert_files()
 
         self._update_ca_certificates()

```

WDYT?

src/charm.py
Abuelodelanada: You are right.

WDYT about this?

```diff
@@ -84,6 +84,7 @@ class CompositeStatus(TypedDict):
     config: Tuple[str, str]
     rules: Tuple[str, str]
     retention: Tuple[str, str]
+    certificates: Tuple[str, str]
 
 
 def to_tuple(status: StatusBase) -> Tuple[str, str]:
@@ -132,6 +133,7 @@ class LokiOperatorCharm(CharmBase):
                 config=to_tuple(ActiveStatus()),
                 rules=to_tuple(ActiveStatus()),
                 retention=to_tuple(ActiveStatus()),
+                certificates=to_tuple(ActiveStatus()),
             ),
             fresh_install=True,
         )
@@ -620,9 +622,13 @@ class LokiOperatorCharm(CharmBase):
             self._loki_container.exec(["update-ca-certificates", "--fresh"]).wait()
             subprocess.run(["update-ca-certificates", "--fresh"])
         except (ConnectionError, ExecError) as e:
-            logger.error("Could not run update-ca-certificates in workload container. %s", e)
+            msg = f"Could not run update-ca-certificates in workload container. {e}"
+            logger.error(msg)
+            self._stored.status["certificates"] = to_tuple(BlockedStatus(msg))
         except subprocess.SubprocessError as e:
-            logger.error("Could not run update-ca-certificates in charm container. %s", e)
+            msg = f"Could not run update-ca-certificates in charm container. {e}"
+            logger.error(msg)
+            self._stored.status["certificates"] = to_tuple(BlockedStatus(msg))
```

dstathis: I think this needs to be error status. Blocked status implies that the user should can fix it. We should just raise the error or raise a custom error like `CACertsError` or something. The bonus of making it an error status is Juju will retry the hook automatically and it might resolve itself.

src/charm.py
dstathis: I am not sure. I think I would still prefer to surface the exception because another hook might not fire. I don't feel strongly on this one though so it is in no way blocking my approval.

src/charm.py
Abuelodelanada: You are right about not setting Blocked status....  raising the error and letting the charm go into error state maybe is a good idea here.
Besides I thought about using other two options: `tenacity` or `.defer()`...

What do you think @sed-i @dstathis ?

dstathis: I think tenacity would likely be better. Retry for a few minutes, then raise an exception if that fails.

Abuelodelanada: This PR started because of a missing `can_connect()` guard in the lib, the changes in `charm.py` seemed trivial, but after this exchange of messages it is clear that we should think better about it. 

Let's follow this conversation in a separate issue: https://github.com/canonical/loki-k8s-operator/issues/434 and merge only the changes to the lib

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2475,6 +2475,9 @@ def disable_inactive_endpoints(
         container: Container, active_endpoints: Dict[str, str], topology: JujuTopology
     ):
         """Disable forwarding for inactive endpoints by checking against the Pebble plan."""
+        if not container.can_connect():
+            return
```
sed-i, line 2479: - `_update_endpoints` is called by `_on_pebble_ready` and `_update_logging`.
- `_update_logging` is called by a bunch of relation observers.

Since we don't want to have a can_connect guard in _on_pebble_ready, seems like the best place to move the guard to is `_update_logging`.

I.e. I would remove the can_connect guards from the `_enable_...`/`_disable_...` methods, and keep it "at the top", because `_update_logging` has better context to decide what to do if we can't connect.

Wdyt?

Abuelodelanada: Since `can_connect()` is a point-in-time check I tend to use that guard as close as possible to the action that connects through Pebble.
A few milliseconds after `can_connect()` returns True, Pebble may not be available... but, how likely is that possibility? 🤔 

While using `can_connect()` we are in a non-deterministic zone, that's why I believe we need to revisit the [interactions with Pebble](https://github.com/canonical/loki-k8s-operator/issues/434) in all our charms and libs, not just Loki.

BUT... in this specific situation we can move the `can_connect()` guard inside the `for` loop in `_update_logging`, and if it not possible to connect just `continue`... because when Pebble is ready again an event will be fired and we'll end up executing `_update_endpoints` anyway.

lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -2562,6 +2562,9 @@ def _update_logging(self, _):
             return
 
         for container in self._charm.unit.containers.values():
+            if not container.can_connect():
+                continue
+
             self._update_endpoints(container, loki_endpoints)
```
sed-i, line 2568:
```suggestion
            if container.can_connect():
                self._update_endpoints(container, loki_endpoints)
            # else: `_update_endpoints` will be called on pebble-ready anyway.
```

sed-i (approved):
Let's proceed with the assumption that pebble-ready is certainly coming after can-connect is false.

PR #442, opened 2024-08-09
Refactored restart logic with replan for workload containers

MichaelThamm:
## Issue
Clean up the restart logic in the charms' `_configure` method.

## Solution
The node_exporter workload container was successfully switched to pebble.replan() logic (although the pebble layer will not change throughout the charm's lifecycle). Since loki has `"startup": "disabled"` set in the pebble layer, there is still a requirement for some restart logic.

## Context
pebble.replan() will only restart services that have `"startup": "enabled"` set in the pebble layer. 

## Testing Instructions
* Integration tests pass

## Upgrade Notes
None

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -46,15 +46,6 @@ from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
 from charms.tempo_k8s.v1.charm_tracing import trace_charm
 from charms.tempo_k8s.v2.tracing import TracingEndpointRequirer
 from charms.traefik_k8s.v1.ingress_per_unit import IngressPerUnitRequirer
-from config_builder import (
-    CERT_FILE,
-    HTTP_LISTEN_PORT,
-    KEY_FILE,
-    LOKI_CONFIG,
-    LOKI_CONFIG_BACKUP,
-    RULES_DIR,
-    ConfigBuilder,
-)
 from ops import CollectStatusEvent, StoredState
 from ops.charm import CharmBase
 from ops.main import main
@@ -67,6 +58,16 @@ from ops.model import (
     WaitingStatus,
 )
 from ops.pebble import Error, Layer, PathError, ProtocolError
+
+from config_builder import (
+    CERT_FILE,
+    HTTP_LISTEN_PORT,
+    KEY_FILE,
+    LOKI_CONFIG,
+    LOKI_CONFIG_BACKUP,
+    RULES_DIR,
+    ConfigBuilder,
+)
 
 # To keep a tidy debug-log, we suppress some DEBUG/INFO logs from some imported libs,
 # even when charm logging is set to a lower level.
--- a/src/charm.py
+++ b/src/charm.py
@@ -273,14 +273,10 @@ class LokiOperatorCharm(CharmBase):
             logger.debug("Cannot set workload version at this time: could not get Loki version.")
 
     def _on_node_exporter_pebble_ready(self, _):
-        current_layer = self._node_exporter_container.get_plan()
         new_layer = self._node_exporter_pebble_layer
-        restart = current_layer.services != new_layer.services
-
-        if restart:
-            self._node_exporter_container.add_layer("node-exporter", new_layer, combine=True)
-            self._node_exporter_container.restart("node-exporter")
-            logger.info("Node Exporter (re)started")
+        self._node_exporter_container.add_layer("node-exporter", new_layer, combine=True)
+        self._node_exporter_container.replan()
+        logger.info("Node Exporter (re)started")
 
     def _on_alertmanager_change(self, _):
         self._configure()
@@ -323,7 +319,7 @@ class LokiOperatorCharm(CharmBase):
         return f"/usr/bin/loki -config.file={LOKI_CONFIG}"
 
     @property
-    def _build_pebble_layer(self) -> Layer:
+    def _loki_pebble_layer(self) -> Layer:
         """Construct the pebble layer.
 
         Returns:
@@ -485,10 +481,6 @@ class LokiOperatorCharm(CharmBase):
                 BlockedStatus("Please provide a non-negative retention duration")
             )
 
-        current_layer = self._loki_container.get_plan()
-        new_layer = self._build_pebble_layer
-        restart = current_layer.services != new_layer.services
-
         # If v12_migration_date isn't set (due to missing or failed retrieval),
         # we determine the migration date for v12 schema. This occurs once
         # during initial setup, as subsequent hooks will get the value from the persisted backup config.
@@ -525,17 +517,17 @@ class LokiOperatorCharm(CharmBase):
             v12_migration_date=v12_migration_date,
         ).build()
 
-        restart = self._update_config(config) or restart
+        restart = self._update_config(config)
+        new_layer = self._loki_pebble_layer
         self._loki_container.add_layer(self._name, new_layer, combine=True)
-        # Now that we for sure have a layer, we can check if the service is running
-        restart = not self._loki_container.get_service(self._service_name).is_running() or restart
 
         if self.server_cert.enabled and not self.server_cert.available:
             # We have a TLS relation in place, but the cert isn't there
             logger.info("Loki stopped")
             self._loki_container.stop(self._service_name)
 
-        elif restart:
+        # Now that we for sure have a layer, we can check if the service is running
+        elif not self._loki_container.get_service(self._service_name).is_running() or restart:
             self._loki_container.restart(self._name)
             logger.info("Loki (re)started")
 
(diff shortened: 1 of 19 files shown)


src/charm.py
```diff
@@ -326,7 +322,7 @@ def _loki_command(self):
         return f"/usr/bin/loki -config.file={LOKI_CONFIG}"
 
     @property
-    def _build_pebble_layer(self) -> Layer:
+    def _loki_pebble_layer(self) -> Layer:
```
MichaelThamm, line 325: I changed this because it is more clear than `build` since that makes it sound like a method not a property. Is there any other location that I should refactor this besides within the `src/charm.py` file?

PietroPasotti: other than test code I wouldn't expect this to be accessed from anywhere else. If the tests pass you're good.

src/charm.py
```diff
@@ -276,14 +276,10 @@ def _on_loki_pebble_ready(self, _):
...
         new_layer = self._node_exporter_pebble_layer
-        restart = current_layer.services != new_layer.services
-
-        if restart:
-            self._node_exporter_container.add_layer("node-exporter", new_layer, combine=True)
-            self._node_exporter_container.restart("node-exporter")
-            logger.info("Node Exporter (re)started")
+        self._node_exporter_container.add_layer("node-exporter", new_layer, combine=True)
+        self._node_exporter_container.replan()
+        logger.info("Node Exporter (re)started")
```
PietroPasotti, line 282: not sure if 'restarted' is correct here. We're adding a layer that didn't exist previously, so probably "started" is enough.

PietroPasotti:
the failing itest is likely due to a missing can-connect guard in the _update_certs path

PietroPasotti:
> @PietroPasotti There is already a con_connect guard in the [_update_cert](https://github.com/canonical/loki-k8s-operator/blob/a3bcf3b47526a8fef2e43e60aada38731920002a/src/charm.py#L567) method. Were you referring to somewhere else in the code?

huh. Interesting. Then I misread the message, or even can_connect errors for some reason

src/charm.py
```diff
@@ -276,14 +277,10 @@ def _on_loki_pebble_ready(self, _):
...
-        current_layer = self._node_exporter_container.get_plan()
         new_layer = self._node_exporter_pebble_layer
-        restart = current_layer.services != new_layer.services
-
-        if restart:
-            self._node_exporter_container.add_layer("node-exporter", new_layer, combine=True)
-            self._node_exporter_container.restart("node-exporter")
-            logger.info("Node Exporter (re)started")
+        self._node_exporter_container.add_layer("node-exporter", new_layer, combine=True)
+        self._node_exporter_container.replan()
```
PietroPasotti, line 282: I'm missing some context on this obviously, but why is it that we can't add startup=True to the layer definition instead?

src/charm.py
```diff
@@ -528,19 +521,19 @@ def _configure(self):  # noqa: C901
...
             # We have a TLS relation in place, but the cert isn't there
             logger.info("Loki stopped")
             self._loki_container.stop(self._service_name)
 
-        elif restart:
+        # Now that we for sure have a layer, we can check if the service is running
+        elif not self._loki_container.get_service(self._service_name).is_running() or restart:
             self._loki_container.restart(self._name)
-            logger.info("Loki (re)started")
+            logger.info("Loki started")
```
PietroPasotti, line 536: Imho it would help with debugging if we added here _why_ loki is being (re)started. is it because the service was not running, or because we just made a change to the config?

src/charm.py
```diff
@@ -528,19 +521,19 @@ def _configure(self):  # noqa: C901
             v12_migration_date=v12_migration_date,
         ).build()
 
-        restart = self._update_config(config) or restart
+        restart = self._update_config(config)
```
PietroPasotti, line 524: what does this restart variable represent? 
```suggestion
        has_config_changed = self._update_config(config)
```

src/charm.py
```diff
@@ -528,19 +521,19 @@ def _configure(self):  # noqa: C901
...
+        new_layer = self._loki_pebble_layer
```
PietroPasotti, line 525: we don't quite know or care if it's `new`, do we?
```suggestion
        self._loki_container.add_layer(self._name, self._loki_pebble_layer, combine=True)
```

PietroPasotti (commented):
I'm not sure I understand exactly the scope of the changes and what exactly they're trying to solve, so I'm hesitant to approve before getting some more context, but I have some low-level feedback on the implementation. let's discuss the higher level in person

src/charm.py
MichaelThamm: Yes your assumption is correct and I think the renaming of the variable is more clear now. Will include in the PR.

src/charm.py
PietroPasotti: can we have clarity on the reason, before we spend more time solving unknown issues?

src/charm.py
```diff
@@ -510,19 +502,22 @@ def _configure(self):  # noqa: C901
             tsdb_versions_migration_dates=self._tsdb_versions_migration_dates,
         ).build()
 
-        restart = self._update_config(config) or restart
-        self._loki_container.add_layer(self._name, new_layer, combine=True)
-        # Now that we for sure have a layer, we can check if the service is running
-        restart = not self._loki_container.get_service(self._service_name).is_running() or restart
+        has_config_changed = self._update_config(config)
+        self._loki_container.add_layer(self._name, self._loki_pebble_layer, combine=True)
```
PietroPasotti, line 506: Still a bit unclear to me what the purpose is of adding a layer which potentially is identical to the one we have already. Can you add a comment to explain what's going on?

src/charm.py
```diff
@@ -510,19 +502,22 @@ def _configure(self):  # noqa: C901
...
         if self.server_cert.enabled and not self.server_cert.available:
             # We have a TLS relation in place, but the cert isn't there
             logger.info("Loki stopped")
             self._loki_container.stop(self._service_name)
 
-        elif restart:
+        # Now that we for sure have a layer, we can check if the service is running
+        elif not self._loki_container.get_service(self._service_name).is_running():
+            self._loki_container.restart(self._name)
+            logger.info("Loki restarted. The service was not in the active state.")
```
PietroPasotti, line 516:
```suggestion
            logger.info("Loki service inactive. Restarting...")
```

src/charm.py
```diff
@@ -510,19 +502,22 @@ def _configure(self):  # noqa: C901
...
+
+        elif has_config_changed:
             self._loki_container.restart(self._name)
-            logger.info("Loki (re)started")
+            logger.info("Loki restarted. There was a change to the configuration.")
```
PietroPasotti, line 520:
```suggestion
            logger.info("Loki configuration changed. Restarting...")
```

src/charm.py
```diff
@@ -510,19 +502,22 @@ def _configure(self):  # noqa: C901
...
-        restart = self._update_config(config) or restart
-        self._loki_container.add_layer(self._name, new_layer, combine=True)
-        # Now that we for sure have a layer, we can check if the service is running
-        restart = not self._loki_container.get_service(self._service_name).is_running() or restart
+        has_config_changed = self._update_config(config)
+        self._loki_container.add_layer(self._name, self._loki_pebble_layer, combine=True)
 
         if self.server_cert.enabled and not self.server_cert.available:
             # We have a TLS relation in place, but the cert isn't there
             logger.info("Loki stopped")
```
PietroPasotti, line 510:
```suggestion
            logger.info("TLS relation present but no certificate is available: stopping Loki...")
```

src/charm.py
PietroPasotti: if the layer is unchanged, do we still want to restart?

MichaelThamm: Technically the pebble layer will always be the same because the command will always be:
`"command": "/usr/bin/loki -config.expand-env=true -config.file=LOKI_CONFIG_DIR/loki-local-config.yaml"`
and the rest of the layer are static strings. I could add code to check `if previous_layer != current_layer` then `replan`.

MichaelThamm: The main reason we add a layer here without checking for layer differences is due to the comment you see on line `515` "# Now that we for sure have a layer, we can check if the service is running". Which we check with `container_is_running()` a few lines later in the code.

Is there a better way to handle ensuring that we can check if the container is running without adding a layer?

PietroPasotti: adding a layer has, as far as I know, no effect on the running/restart behaviour unless the plan changes as a result.
To check the service is running I would `self._loki_container.get_service(self._service_name).is_running()` as it was before, you shouldn't need to re-add the layer before.
Not saying it won't work, I just think it's effectively a no-op

PietroPasotti (approved):
GJ! and thanks once more for the patience :)

PR #446, opened 2024-09-09
Chore: Pin Scenario to version 6

MichaelThamm:
## Issue
With the release of Scenario v7 there are a lot of breaking changes like [ExecOutput is now Exec](https://github.com/canonical/ops-scenario/blob/29a50b75893f199d83713d675588bdbb2efccf13/UPGRADING.md?plain=1#L292). Scenario is pinned to latest release of major version 6 (6.1.7).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -86,7 +86,7 @@ description = Scenario tests
 deps =
     pytest
     pydantic>=2
-    ops-scenario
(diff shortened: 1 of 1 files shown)


PietroPasotti (approved):
right, let's open an issue to bump though

PR #448, opened 2024-09-10
Initial Commit

MichaelThamm:
## Issue
As part of [OPENG-2677](https://warthogs.atlassian.net/browse/OPENG-2677), we want to _Investigate goss to see how much of the cos doctor capacities we could possibly fulfil with goss_. The [OB041 - Juju Doctor (Pebble vs. Goss)](https://docs.google.com/document/d/10EhRTWHAm7Oyog6j4iFkaAOP6tJLlqS8tKnS5AoCk4w/edit) doc was created and highlights a sample, external Goss architecture with charm health checks via Pebble.

## Solution
This PR aims to provide:

1. A `goss.yaml` in the root of the charm that is used with the [Goss binary](https://goss.readthedocs.io/en/stable/installation/) on the operator's host to provide predefined tests, specific to the charm repo. 
2. The charm author is expected to add necessary [Pebble health checks](https://juju.is/docs/sdk/interact-with-pebble#heading--perform-health-checks-on-the-workload-container) in the `charm.py` file for any necessary, internal checks.

This architecture is generic and can be applied to any charm.

## Context
* Internal tests are expected to be configured using Pebble health checks added to the Pebble layer. The Pebble binary offers two commands:
  * `health` -> Returns `healthy` if all checks pass
  * `checks` -> Returns a list of passing and failing checks
* External tests are expected to be configured in the `goss.yaml` file in the root of the charm repo which (potentially) has access to the host's `juju` and `k8s` context.
  * Therefore, we can run tests like `juju status --relations ...` and `kubectl describe ...`.

## Testing Instructions
1. Install the Goss binary on your host with `curl -fsSL https://goss.rocks/install | sh`
2. Deploy the charm
```
charmcraft pack
juju deploy ./*.charm loki-k8s --resource loki-image=docker.io/ubuntu/loki:2-22.04 --resource node-exporter-image=docker.io/prom/node-exporter:v1.7.0 --trust
```
3. Test with Goss using `goss v -f tap`

Expected results:
```
1..6
ok 1 - Command: charm-exit-codes: exit-status: matches expectation: 0
ok 2 - Command: charm-exit-codes: stdout: matches expectation: ["Running"]
ok 3 - Command: relation-cluster: exit-status: matches expectation: 0
ok 4 - Command: relation-cluster: stdout: matches expectation: ["grafana-source"]
ok 5 - Command: self-tests-health: exit-status: matches expectation: 0
ok 6 - Command: self-tests-health: stdout: matches expectation: ["healthy"]
```

### DIFF
--- a/goss.yaml
+++ b/goss.yaml
@@ -1,0 +1,16 @@
+command:
+  self-tests-health:
+    exec: juju ssh loki-k8s/0 /charm/bin/pebble health
+    exit-status: 0
+    stdout:
+      - healthy
+  relation-cluster:
+    exec: juju status --relations | grep grafana_datasource | grep regular
+    exit-status: 0
+    stdout:
+      - grafana-source
+  charm-exit-codes:
+    exec: kubectl describe -n welcome-k8s pod loki-0 | grep -A 15 "Containers:" | grep -A 15 "charm:" | grep -A 3 "State:"
+    exit-status: 0
+    stdout:
+      - Running
--- a/src/charm.py
+++ b/src/charm.py
@@ -66,7 +66,7 @@ from ops.model import (
     StatusBase,
     WaitingStatus,
 )
-from ops.pebble import Error, Layer, PathError, ProtocolError
+from ops.pebble import Error, Layer, PathError, ProtocolError, Check
 
 # To keep a tidy debug-log, we suppress some DEBUG/INFO logs from some imported libs,
 # even when charm logging is set to a lower level.
@@ -492,6 +492,8 @@ class LokiOperatorCharm(CharmBase):
         new_layer = self._build_pebble_layer
         restart = current_layer.services != new_layer.services
 
+        self._add_readiness_check(new_layer)
+
         # If v12_migration_date isn't set (due to missing or failed retrieval),
         # we determine the migration date for v12 schema. This occurs once
         # during initial setup, as subsequent hooks will get the value from the persisted backup config.
@@ -809,6 +811,11 @@ class LokiOperatorCharm(CharmBase):
             return self._ca_cert_path
         return None
 
+    def _add_readiness_check(self, new_layer: Layer):
+        """Add readiness check to a pebble layer."""
+        new_layer.checks["ready"] = Check(
+            "ready", {"override": "replace", "http": {"url": "http://localhost:3100/ready"}}
+        )
 
 if __name__ == "__main__":
     main(LokiOperatorCharm)


goss.yaml
```diff
@@ -0,0 +1,16 @@
+command:
```
sed-i, line 1: Can we use cli args here somehow, specifically model name, to pass into `juju -m {model_name}`?

goss.yaml
```diff
@@ -0,0 +1,16 @@
...
+  relation-cluster:
+    exec: juju status --relations | grep grafana_datasource | grep regular
+    exit-status: 0
+    stdout:
+      - grafana-source
+  charm-exit-codes:
+    exec: kubectl describe -n welcome-k8s pod loki-0 | grep -A 15 "Containers:" | grep -A 15 "charm:" | grep -A 3 "State:"
+    exit-status: 0
+    stdout:
+      - Running
```
sed-i, line 16: Does goss have json/yaml checks? E.g. we exec `kubectl -o json` and then the assertion is one something like '.containers.charm.state'.

goss.yaml
```diff
@@ -0,0 +1,16 @@
+command:
+  self-tests-health:
+    exec: juju ssh loki-k8s/0 /charm/bin/pebble health
```
sed-i, line 3: - Is the idea to have a `goss.yaml` per charm?
- Can we loop in goss over several charm names, if the checks are the same?

src/charm.py
```diff
@@ -809,6 +811,11 @@ def _charm_logging_ca_cert(self) -> Optional[str]:
             return self._ca_cert_path
         return None
 
+    def _add_readiness_check(self, new_layer: Layer):
+        """Add readiness check to a pebble layer."""
+        new_layer.checks["ready"] = Check(
+            "ready", {"override": "replace", "http": {"url": "http://localhost:3100/ready"}}
```
sed-i, line 817: The readiness check, when pebble will be able to trigger an event on success, would be very useful, and in some cases could replace the logic we run on pebble-ready.

With our current TLS design, would need to use the external url when TLS is enabled.

goss.yaml
MichaelThamm: In the case of a bundle we could have 1 `goss.yaml` which is responsible for checking each charm's pebble health and running some extra external checks like relations.

Technically there is no need to have the `goss.yaml` in the charm repo (these should be in a COS Doctor repo).

goss.yaml
```diff
@@ -0,0 +1,16 @@
...
+    exit-status: 0
+    stdout:
+      - healthy
+  relation-cluster:
+    exec: juju status --relations | grep grafana_datasource | grep regular
+    exit-status: 0
+    stdout:
+      - grafana-source
```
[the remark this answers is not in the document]
MichaelThamm: Yes, I think this is possible with `juju status --relations | grep grafana-source` with some templated variables for app names to ensure the correct apps are related over this interface

goss.yaml
MichaelThamm: Yes goss has the following output options when running `validate`:

(default: "rspecish")
* documentation
* json
* junit
* nagios
* prometheus
* rspecish
* silent
* structured
* tap

However, in the case that we just exec `kubectl -o json` we should probably require the operator's host to have JQ and YQ installed because we will likely have a lot of exec commands like `kubectl -o json | jq -r '.containers.charm.state'` and the stdout check will have to reflect the expected value (`active/idle` in the example above).

goss.yaml
MichaelThamm: You can loop over a list of variables defined in a `vars.yaml` like so:

```
# Loop over users
user:
{{range .Vars.users}}
  {{.}}:
    exists: true
    groups:
    - {{.}}
    home: /home/{{.}}
    shell: /bin/bash
{{end}}
```

MichaelThamm:
Closing due to stale PR, will re-open or create new one once initiative is active again

PR #455, opened 2024-09-13
add pipeline_stage to remove `filename` label

Abuelodelanada:
## Issue

This PR goes in tandem with https://github.com/canonical/grafana-agent-operator/pull/180

## Solution

We add `pipeline_stages` to `promtail` configuration in order to remove `filename` label an add it as `structured_metadata`

## Testing Instructions

Deploy this bundle (update lib in `flog` and pack `loki`)

```yaml
bundle: kubernetes
applications:
  flog:
    charm: flog-k8s
    channel: latest/edge
    revision: 8
    resources:
      workload-image: 2
    scale: 1
    constraints: arch=amd64
  grafana:
    charm: grafana-k8s
    channel: latest/edge
    revision: 118
    resources:
      grafana-image: 70
      litestream-image: 45
    scale: 1
    constraints: arch=amd64
    storage:
      database: kubernetes,1,1024M
    trust: true
  loki:
    charm: local:loki-k8s-0
    scale: 1
    constraints: arch=amd64
    storage:
      active-index-directory: kubernetes,1,1024M
      loki-chunks: kubernetes,1,1024M
    trust: true
  traefik:
    charm: traefik-k8s
    channel: latest/edge
    revision: 207
    resources:
      traefik-image: 160
    scale: 1
    constraints: arch=amd64
    storage:
      configurations: kubernetes,1,1024M
    trust: true
relations:
- - traefik:ingress-per-unit
  - loki:ingress
- - grafana:ingress
  - traefik:traefik-route
- - grafana:grafana-source
  - loki:grafana-source
- - flog:log-proxy
  - loki:logging
```

Check in Grafana that `filename` is not a regular label but part of `structured_metadata`

image

image

PR #459, opened 2024-09-18
use loki_push_api v1

Abuelodelanada:
## Issue

Now we use loki_push_api v1.... 

## Context

While working on https://github.com/canonical/loki-k8s-operator/pull/455 and https://github.com/canonical/grafana-agent-operator/pull/185 I've found that we were using v0 of the lib


lucabello (approved):
Approving, but please fix itests and maybe manually test things as well :)

PR #494, opened 2025-03-03
feat: add grafana_base_url to Grafana source

MichaelThamm:
## Issue
Loki needs to know the Grafana base URL for its `ruler` config which may have an ingress. This is needed to correctly render the Source button URL to Grafana when Loki alert rules are firing:
image
 
## Solution
Is this relation the best way to give Loki the ingress URL from Grafana? It makes sense since this scenario requires the `grafana-source` relation, otherwise Loki logs won't make it to Grafana and this scenario won't be possible.
- [x] Add tests for Loki consuming the grafana URL

## Context
- https://github.com/canonical/alertmanager-k8s-operator/issues/293
- https://github.com/canonical/grafana-k8s-operator/pull/398
- https://grafana.com/docs/loki/latest/configure/#ruler

## Testing Instructions
1. Create an always firing Loki alert rule in the [Zinc charm](https://github.com/jnsgruk/zinc-k8s-operator):
```yaml
alert: HostHighLogErrorRate
expr: count_over_time({juju_application="zinc"} [1h]) > -1
for: 0m
labels:
  severity: warning
annotations:
  summary: High error rate in logs (instance {{ $labels.instance }})
  description: "High error rate in logs\n  VALUE = {{ $value }}\n  LABELS = {{ $labels }}"
```
2. Pack the zinc charm
3. Deploy Zinc
4. Deploy cos-lite
```
curl -L https://raw.githubusercontent.com/canonical/cos-lite-bundle/main/overlays/offers-overlay.yaml -O
juju deploy cos-lite --trust --overlay ./offers-overlay.yaml
juju deploy 
```
5. Pack the Loki and Grafana charms with the changes in this PR and the tandem PR
6. `juju refresh` both charms with the packed charms
7. Relate Zinc to Loki for logging and generate some logs in the Zinc UI (log in or run a query)
8. Check the Alertmanager UI for the Source button link to Grafana ingress URL instead of Loki's ingress URL

- [x] Check that the config is empty until we receive a valid URL (how did this not error before?)
- [x] Check with multiple grafana's related
- [x] Check with multiple loki units to 1 Grafana

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -432,21 +432,14 @@ class GrafanaSourceProvider(Object):
     def get_grafana_base_urls(self) -> Dict[str, str]:
         """Get the grafana base URL (potentially ingressed) assigned by the remote end(s) to this datasource.
 
-        Returns a mapping from remote application UIDs to URL.
+        Returns a mapping from relation ID to URL.
         """
         urls = {}
         for rel in self._charm.model.relations.get(self._relation_name, []):
             if not rel:
                 continue
             app_databag = rel.data[rel.app]
-            grafana_uid = app_databag.get("grafana_uid")
-            if not grafana_uid:
-                logger.warning(
-                    "remote end is using an old grafana_datasource interface: "
-                    "`grafana_uid` field not found."
-                )
-                continue
-            urls[grafana_uid] = app_databag.get("grafana_base_url")
+            urls[rel.id] = app_databag.get("grafana_base_url")
         return urls
 
     def get_source_uids(self) -> Dict[str, Dict[str, str]]:
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -430,10 +430,9 @@ class GrafanaSourceProvider(Object):
             self._set_sources(rel)
 
     def get_grafana_base_urls(self) -> Dict[str, str]:
-        # TODO update docstring
-        """Get the grafana external URL assigned by the remote end(s) to this datasource.
+        """Get the grafana base URL (potentially ingressed) assigned by the remote end(s) to this datasource.
 
-        Returns a mapping from remote application UIDs to unit names to datasource uids.
+        Returns a mapping from remote application UIDs to URL.
         """
         urls = {}
         for rel in self._charm.model.relations.get(self._relation_name, []):
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -177,7 +177,7 @@ class GrafanaSourceData:
     """This class represents the data Grafana provides others about itself."""
 
     datasource_uids: Dict[str, str]
-    external_url: str
+    external_url: Optional[str]
 
     def get_unit_uid(self, unit: str):
         """Return the UID for a given unit."""
@@ -223,8 +223,7 @@ class RelationInterfaceMismatchError(Exception):
         self.expected_relation_interface = expected_relation_interface
         self.actual_relation_interface = actual_relation_interface
         self.message = (
-            "The '{}' relation has '{}' as "
-            "interface rather than the expected '{}'".format(
+            "The '{}' relation has '{}' as interface rather than the expected '{}'".format(
                 relation_name, actual_relation_interface, expected_relation_interface
             )
         )
@@ -465,7 +464,7 @@ class GrafanaSourceProvider(Object):
                 continue
             grafana_data = GrafanaSourceData(
                 datasource_uids=json.loads(app_databag.get("datasource_uids", "{}")),
-                external_url=app_databag.get("grafana_base_url", "")
+                external_url=app_databag.get("grafana_base_url"),
             )
             data[grafana_uid] = grafana_data
         return data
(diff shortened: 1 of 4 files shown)


lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -429,6 +429,26 @@ def update_source(self, source_url: Optional[str] = ""):
...
+                continue
+            app_databag = rel.data[rel.app]
+            grafana_uid = app_databag.get("grafana_uid")
+            if not grafana_uid:
+                logger.warning(
+                    "remote end is using an old grafana_datasource interface: "
+                    "`grafana_uid` field not found."
+                )
+                continue
+            urls[grafana_uid] = app_databag.get("grafana_base_url")
```
MichaelThamm, line 449: We could sort by the relation_id because the string sorting is not reliable

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -429,6 +429,19 @@ def update_source(self, source_url: Optional[str] = ""):
...
+        """Get the grafana base URL (potentially ingressed) assigned by the remote end(s) to this datasource.
+
+        Returns a mapping from relation ID to URL.
+        """
+        urls = {}
+        for rel in self._charm.model.relations.get(self._relation_name, []):
+            if not rel:
+                continue
+            app_databag = rel.data[rel.app]
+            urls[rel.id] = app_databag.get("grafana_base_url")
```
MichaelThamm, line 442: Here we set relation IDs as the keys so we can consistently sort by integers rather than the [strings from grafana_uid](https://github.com/canonical/loki-k8s-operator/pull/494/files#diff-7ed482b6184969c9603036d352837ddec840d76473e57a583c8b3c5bc49f4cbeL450) from `get_source_uids`.

We need to sort on both of these results so should we update this public method for matched functionality but risk breaking changes?

PietroPasotti, line 442: this looks good. So long as you're using `.get` instead of `[]` you should be backwards-compatible.

PietroPasotti: Why do we need to sort (with integers)? 
`get_source_uids` returns a mapping from remote application names to (remote unit names to grafana uids). Application names are just as sortable as integers, as they're guaranteed by juju not to contain duplicates.

Relation IDs are probably useless to the caller, so IMO we should avoid returning them.
Two proposals on how we could do this instead:

1. use the IDs to sort the output, but don't return them as part of the output
```
urls = []
for rel in self._charm.model.relations.get(self._relation_name, []):
    try:
        urls.append((rel.id, rel.data[rel.app].get("grafana_base_url"))
    except: pass
return [url[1] for url in sorted(urls, key=lambda url: url[0])]
```

2. consistently with `get_source_uids`, return a mapping from remote app names to urls instead
```
urls = {}
for rel in self._charm.model.relations.get(self._relation_name, []):
    try:
        urls[rel.app.name] = rel.data[rel.app].get("grafana_base_url")
    except: pass
return urls
```

src/charm.py
```diff
@@ -555,6 +556,7 @@ def _configure(self):  # noqa: C901
         if self.server_cert.available and not self._certs_on_disk:
             self._update_cert()
 
+        datasource_uid, grafana_external_url = self._sorted_ruler_cfg()
```
PietroPasotti, line 559: in general I'm wary of this pattern:
`apples, pears = self._get_potatoes()`

I'd recommend:
```
# good: some duplicated work perhaps, but clear API
apples = self._get_apples() 
pears = self._get_pears() 

# better: make the method return a dataclass/namedtuple/some namespace
fruit = self._get_fruit()
apples = fruit.apples
pears = fruit.pears

# bad: hard to extend if you want to add bananas one day
apples, pears = self._get_apples_and_pears()
```

PietroPasotti (commented):
my 2c

lib/charms/grafana_k8s/v0/grafana_source.py
MichaelThamm: You mean `rel.data.get(rel.app)` instead of `rel.data[rel.app]`? Im not sure which `[]` you are referring to

src/charm.py
MichaelThamm: If we pre-sort in the methods from the library, then we don't need any of this code.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -131,6 +131,7 @@ def __init__(self, *args):
 import logging
```
MichaelThamm, line 131: I left the changes here in Loki for the grafana_source lib just for CI and testing. I mirrored the changes in the Grafana PR and will delete this here (and run charmcraft fetch-lib) once the Grafana PR is merged.

lucabello (approved):

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -171,6 +172,22 @@ def __init__(self, *args):
...
+
+    datasource_uids: Dict[str, str]
+    external_url: str
+
+    def get_unit_uid(self, unit: str):
+        """Return the UID for a given unit."""
+        if unit in self.datasource_uids:
+            datasource_uid = self.datasource_uids[unit]
+        else:
+            datasource_uid = ""
```
Abuelodelanada, line 187: Just a minor style change in this Grafana lib:

```suggestion
        datasource_uid = ""
        if unit in self.datasource_uids:
            datasource_uid = self.datasource_uids[unit]            
```

Abuelodelanada (approved):

lib/charms/grafana_k8s/v0/grafana_source.py
MichaelThamm: Unfortunately, already merged the PR on Grafana lib. Do you think this is worth a new lib version bump?

PR #505, opened 2025-04-01
feat: move datasource refresh logic into a reconcile function

michaeldmitry:
## Issue
`test_datasource_exchange` tests started failing due to a missing handler for `relation_created` and `relation_joined` on both `datasource` and `send-datasource` relations. (no idea how did that work earlier) 

## Solution
Instead of observing 4 additional events, call `_update_datasource_exchange` inside a `reconcile` function.

## Context
https://github.com/canonical/tempo-coordinator-k8s-operator/blob/main/src/charm.py#L624
https://github.com/canonical/tempo-coordinator-k8s-operator/blob/main/src/charm.py#L624


PietroPasotti (commented):
all for adding a reconciler, but we've been thinking that the 'right' way to do it shouldn't be to call it directly from `__init__` as that's too 'hacky' cit. @ca-scribner.

We have a call in a couple of days to agree on a less-hacky pattern we can roll out to all our charms. Perhaps wait on that? (or join?)

lucabello:
Waiting is fine by me! Can I `@skip` the test for now, with a `TODO:` on putting it back?

michaeldmitry:
> Waiting is fine by me! Can I `@skip` the test for now, with a `TODO:` on putting it back?

You'll also need the same for prometheus

simskij:
As the PR is fairly trivial, but the branch has been dead for 8 months, I'll just go ahead and close this. We can open a new PR when/if we want to do the change.

PR #506, opened 2025-04-02
feat: migrate to ci v1

lucabello:
This PR migrates the repository to CI v1.

Notes:
- a unit test is being skipped until #505 is decided on
- `.charm_tracing_buffer.raw` was removed
- the now unused `doc` folder with one ADR from 2023 was removed


MichaelThamm (approved):
It would be nice (for future large repos, e.g. prometheus) to split these migration PRs into smaller PRs if possible (I understand some changes are dependent on one another) like:
- `migrate functionality` -> workflows, dep mgmt, charm code, ...
- `migrate misc` -> moving to tests/resources, removing doc/adr, unrelated linting/static fixes, ...

Seeing 52 changes files is intimidating and makes me not want to approve, but doing so anyways since we reviewed over a call together 🙂

lucabello:
I'm aware :) The problem unfortunately is that new CI would fail because we lint new things (i.e., tests and alert rules), which need to be changed together with `tox.ini`, which in turn needs to be changed with the `uv` changes 

`migrate misc` is a good idea, but it's hard to see static checks fail and only fix a part of it; CI wouldn't even succeed on the first PR

It's hard to split in two is what I'm trying to say 🙈

PR #510, opened 2025-04-24
feat(terraform): add terraform module

simskij:
## Issue
There is currently no terraform module available for deploying Loki in the context of COS Lite.

## Solution
Add a terraform module

## Testing Instructions
- Clone repo
- `terraform init`
- `terraform apply -var="model_name=<MODEL_NAME>"`

## Upgrade Notes
Add terraform module for COS Lite flavor of Loki

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -1,0 +1,40 @@
+# Terraform module for `loki-k8s`
+
+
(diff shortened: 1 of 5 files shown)


Abuelodelanada (approved):
Before running `terraform apply -var="model_name=<MODEL_NAME>"` we need to run `terraform init`

PR #513, opened 2025-05-07
chore: finalize track 1

lucabello:


sed-i (approved):
Plan:
- Merge with the xfail, after 24.04 removed from platforms
- (CI releases to track/1)
- Remove xfails and update itest to deploy from track 1 instead of latest

PR #530, opened 2025-08-14
feat: add extra_alert_labels functionality to log alerts

### LINKED ISSUE #529: Allow adding extra labels to log alerts in the loki_push_api library
### Enhancement Proposal

Similarly to what was done for metrics in https://github.com/canonical/prometheus-k8s-operator/pull/714 , we should enrich log alerts at the point of "sending them to Loki" with extra labels optionally passed by other charms.

lucabello:
## Issue
Closes #529 

Mimics what is done in https://github.com/canonical/prometheus-k8s-operator/pull/714

### DIFF
--- a/lib/charms/loki_k8s/v1/loki_push_api.py
+++ b/lib/charms/loki_k8s/v1/loki_push_api.py
@@ -498,6 +498,7 @@ class MyCharm(...):
 Do this, and all charm logs will be forwarded to Loki as soon as a relation is formed.
 """
 
+import copy
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -1562,6 +1565,15 @@ def __init__(
 
         self._recursive = recursive
 
+    @staticmethod
+    def _inject_extra_labels_to_alert_rules(rules: Dict, extra_alert_labels: Dict) -> Dict:
```
sed-i, line 1569: Consider opening an issue for moving this function into https://github.com/canonical/cos-lib/blob/6cd3769727419f50018994591a90c0024542389b/src/cosl/rules.py#L175

As-is we'd have negative impact on coverage, but at the same time we do not want to repeat the tests.

sed-i (approved):

PR #537, opened 2025-09-10
fix: deduplicate endpoints when Loki is scaled

sinapah:
## Issue

Currently, the way the `loki_push_api` lib is written, when something like Flog is related to Loki HA (which doesn't do ingress per unit), then, in Flog's `promtail_config`, the URL for `promtail` clients is duplicated, which causes Flog to go into error state. For example, if you scale Loki coordinator up to 3, then, the clients in `promtail_config` will look like something like:
```
clients:
- tls_config:
    insecure_skip_verify: true
  url: http://10.181.160.194/test-my-loki/loki/api/v1/push
- tls_config:
    insecure_skip_verify: true
  url: http://10.181.160.194/test-my-loki/loki/api/v1/push
- tls_config:
    insecure_skip_verify: true
  url: http://10.181.160.194/test-my-loki/loki/api/v1/push
```
The URL duplication will result in an error in Flog: `hook failed: log-proxy-relation changed`
## Solution

We need to deduplicate the endpoints list so we don't end up with duplicated URLs. After these changes, the config will be:
```
- tls_config:
    insecure_skip_verify: true
  url: http://10.181.160.194/test-my-loki/loki/api/v1/push
```

## Context

This change is necessary to resolve Flog in [this integration test](https://github.com/canonical/loki-coordinator-k8s-operator/blob/main/tests/integration/test_charm_ha_scaled.py) for Loki Coordinator. After this PR is merged, we need to update the library in Flog and push to stable for this `iTest` to pass. The updated config above is when the changes are locally applied to Flog's `loki_push_api`.

### DIFF
--- a/lib/charms/loki_k8s/v1/loki_push_api.py
+++ b/lib/charms/loki_k8s/v1/loki_push_api.py
@@ -1621,6 +1621,10 @@ class ConsumerBase(Object):
                 deserialized_endpoint = json.loads(endpoint)
                 url = deserialized_endpoint.get("url")
 
+                # It's necessary to deduplicte the endpoints because if we don't do this, in the event that Loki coordinator is related to Flog and is scaled,
+                # in the /etc/promtail/promtail_config.yaml of Flog, there will duplicate entries for each unit of the Loki coordinator, which will cause Flog to go into error due to the duplication.
+                # The deduplication applied here
+
                 if not url or url in seen_urls:
                     continue
 


lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -1601,28 +1601,39 @@ def loki_endpoints(self) -> List[dict]:
...
+
+                    """
+                    # It's necessary to deduplicte the endpoints because if we don't do this, in the event that Loki coordinator is related to Flog and is scaled,
+                    # in the /etc/promtail/promtail_config.yaml of Flog, there will duplicate entries for each unit of the Loki coordinator, which will cause Flog to go into error due to the duplication.
+                    # The deduplication applied here
+                    """
+
+                    if url and url not in seen_urls:
+                        seen_urls.add(url)
+                        endpoints.append({"url": url})
```
Abuelodelanada, line 1631: Let's use guard-clause to un-indent the code

```suggestion
                    if not url or url in seen_urls
                        continue
                    
                    seen_urls.add(url)
                    endpoints.append({"url": url})
```


Right now this property has 4 indentation levels.
We can even go further and make this property much more readable and easy to reason about with only 2 indentation levels:

```python
    @property
    def loki_endpoints(self) -> List[dict]:
        """Fetch Loki Push API endpoints sent from LokiPushApiProvider through relation data.

        Returns:
            A list of unique dictionaries with Loki Push API endpoints, for instance:
            [
                {"url": "http://loki1:3100/loki/api/v1/push"},
                {"url": "http://loki2:3100/loki/api/v1/push"},
            ]
        """
        endpoints = []
        seen_urls = set()

            for unit in relation.units:
                if unit.app == self._charm.app:
                    continue

                if not endpoint := relation.data[unit].get("endpoint"):
                    continue

                deserialized_endpoint = json.loads(endpoint)
                url = deserialized_endpoint.get("url")

                if not url or url in seen_urls:
                    continue

                seen_urls.add(url)
                endpoints.append({"url": url})

        return endpoints
```

Abuelodelanada (approved):
LGTM!

PR #559, opened 2025-12-05
feat(LokiPushApiProvider): deprecate `address` arg

### LINKED ISSUE #554: [bug] lib ignores address from constructor.
### Bug Description

In def _process_logging_relation_changed(self, relation: Relation)

The _update_endpoint(relation=relation) will ignore if the self.address is set. This effectively always pass the hostname on the relation for the endpoint, despite if the charm author wants the ip-address to be passed instead.

### To Reproduce

Create a LokiProvider charm and relate to opentelemetry-collector - the endpoint relation will always have the hostname on the relation even if I set the address to the instance IP.

### Environment

ubuntu24

### Relevant log output

```shell
No logs
```

### Additional context

_No response_

sed-i:
## Issue
We [added](https://github.com/canonical/loki-k8s-operator/pull/144/commits/957029d72f5eed17744824000720db532242bef5) an `address` arg (and `self.address`) in #144, but that value was never used.

## Solution
Deprecate the `address` arg.

Closes #554.

## Context
See #555.

### DIFF
--- a/lib/charms/loki_k8s/v1/loki_push_api.py
+++ b/lib/charms/loki_k8s/v1/loki_push_api.py
@@ -79,7 +79,6 @@ and three optional arguments.
               external_url = urlparse(self._external_url)
               self.loki_provider = LokiPushApiProvider(
                   self,
-                  address=external_url.hostname or self.hostname,
                   port=external_url.port or 80,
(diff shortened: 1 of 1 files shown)


sed-i (commented):
Good suggestions from @copilot.

PR #565, opened 2026-01-20
feat: Store updated PushAPI endpoint

MichaelThamm:
## Issue

This PR in otelcol:
- https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/173

Requires updating the `endpoint` field in the LokiPushApiProvider databag. This worked fine previously for non-reconciler charms, but now we need support for our common reconciler-pattern charms.

## Solution

Add a new attribute named `_custom_url` which is set when the public `update_endpoint` method is called. This defaults to `None` and the `endpoint` field will always revert to `self._url` if `self._custom_url` is not set.


Abuelodelanada (approved):

MichaelThamm:
@Abuelodelanada I went back and checked the Canonical repo for all usages of the public `update_endpoint` for extra confidence in this feature

PR #572, opened 2026-03-07
test: Migrate integration tests to Jubilant

### LINKED ISSUE #595: Tracking issue: migrate itests to Jubilant
We need to review PR #572. This is a tracking issue to keep this work on our radar.

tonyandrewmeyer:
## Issue

This PR migrates the integration tests from pytest-operator and python-libjuju to Jubilant and pytest-jubilant.

Note that this was originally written by AI as part of exploratory work by Charm Tech to see how AI (particularly copilot) can help get charms migrated to Jubilant. This PR is an example of that work (a post will appear on Discourse very soon referencing this PR and will more details). However, I have also reviewed it and made tweaks, so am putting myself as the human in the loop, so to speak.

However, what is needed is review from someone who is familiar with the charm and its tests, which I am not.

## Solution

Using Jubilant allows for simpler and more modern and maintainable tests. Most importantly, it allows running the integration tests against both Juju 3 and Juju 4.

## Context

N/A

## Testing Instructions

Check the tests pass, human review that they do the same as the old ones.


tests/integration/conftest.py
```diff
@@ -1,147 +1,129 @@
...
-        start_time = datetime.now()
-        if fname in store.keys():
-            ret = store[fname]
-        else:
-            logger.info("Return for {} not cached".format(fname))
-            ret = await func(*args, **kwargs)
-            store[fname] = ret
-        logger.info("Finished: {} in: {} seconds".format(fname, datetime.now() - start_time))
-        return ret
+    Copied from pytest-jubilant 1.x, which no longer provides pack() in 2.0.
```
tonyandrewmeyer, line 20: FYI, Charm Tech's recommendation is that packing is done outside of the test code. This should better suit using `charmcraft test` / spread, but we feel that it's generally better anyway. For now, this PR is already rather large and it seems simpler to just have the old pack function vendored in, and you can deal with it how you like later on.

tests/integration/conftest.py
```diff
@@ -1,147 +1,129 @@
...
-
-    def __setattr__(self, key, value):
-        """Override __setattr__ so dot syntax works on keys."""
-        self[key] = value
-
-
-store = Store()
-
-
-def timed_memoizer(func):
```
tonyandrewmeyer, line 38: Claude dropped this, and I'm not entirely sure why -- maybe just because of the move away from async? It seems like it shouldn't always be needed, but if it is then I can add it back.

tests/integration/conftest.py
```diff
@@ -1,147 +1,129 @@
...
 
 @pytest.fixture(scope="session")
 def cos_channel():
     return "2/edge"
 
-@pytest.fixture(scope="module", autouse=True)
-def copy_loki_library_into_test_charms(ops_test):
-    """Ensure that the tester charm uses the current Prometheus library."""
+
+@pytest.fixture(scope="session", autouse=True)
```
tonyandrewmeyer, line 58: Claude also switched these to session. I think that should be fine, because it's just the library / packed charm, not a deployed charm, so it seems like it should be safe to share throughout the whole session.

tests/integration/test_alert_rules_fire.py
```diff
@@ -19,74 +19,49 @@
 }
 
 
-@pytest.mark.abort_on_fail
```
tonyandrewmeyer, line 22: There's no abort_on_fail in pytest-jubilant 2.0 (see comments in https://github.com/canonical/jubilant/issues/122). For now, I've just used fail fast, but a true cross-fail might be what you want. If so, I think opening an issue in pytest-jubilant would be best (I don't think it was ruled out there).

cc @james-garner-canonical in case I have this wrong.

tests/integration/helpers.py
```diff
@@ -347,70 +323,28 @@ async def juju_show_unit(
...
-    app_data_arg = "--app" if app_data_only else ""
-    cmd = filter(
-        None,
-        f"juju show-unit {unit_name} {endpoint_arg} {related_unit_arg} {app_data_arg}".split(" "),
-    )
-
-    retcode, stdout, stderr = await ops_test.run(*cmd)
-    assert retcode == 0, f"`juju show-unit` failed: {(stderr or stdout).strip()}"
-
-    # Response looks like this:
```
tonyandrewmeyer, line 369: Claude dropped this, but maybe you want to keep it? Or maybe you'd want something more native in Jubilant than using `cli()` here? (If so, please do open a ticket).

tonyandrewmeyer:
Hi @canonical/observability ! I've bumped this to handle pytest-operator 2.0, and do some other minor fixes. It's up to you what you do with it now (it is a bit large, although a lot of the line count changes is from the lock file). Let me know if you need anything more from me, but otherwise it's over to you to review as people that know the repo way better (or to reject, if that's your preference).

tests/integration/conftest.py
MichaelThamm: I also removed this for another charm's jubilant tests. I think this is fine.

sinapah:
Hi @tonyandrewmeyer! Thanks for putting up this PR, it looks great! We'll keep in mind to review this PR as soon as we can.

lucabello:
Done in #625 , thanks!

PR #591, opened 2026-04-22
fix: prevent v13 schema date drift on pod churn and upgrades

### LINKED ISSUE #589: After upgrade-charm, the v13 schema's `from` date is set to tomorrow causing breakage
### Bug Description

When an `upgrade-charm` happens, the Loki config is regenerated by the charm and pushed to the workload container. Every time the charm attempts to rewrite the workload config file, it has to set the `from` dates for the V11 and V13 schemas under `schema_config`:
```yaml
schema_config:
  configs:
  - from: '2020-10-24'
    index:
      period: 24h
      prefix: index_
    object_store: filesystem
    schema: v11
    store: boltdb-shipper
  - from: '2026-04-21'
    index:
      period: 24h
      prefix: index_
    object_store: filesystem
    schema: v13
    store: tsdb
```

It  sets the date for the `from` key of the v13 schema to today **if** this is a fresh deployment. However, if we're upgrading, the `from` key for the v13 schema is set to tomorrow to "to accommodate today's logs
         might have been written in the previous v11/v12 schema formats". That means, the V13 schema will only be used to read data dated from the day after the charm is deployed. This has the potential of breaking the charm in the following scenario:

1. We deploy the charm today. Assume today is 2026-04-21. We ingest some logs dated 2026-04-21.
2. Tomorrow, the charm is upgraded. Because it's an upgrade, for v13, we set the `from` key to be 2026-04-22. However, the logs that were ingested on the 21st, should be accessed by the TSDB index (v13) but because we have moved the from date to tomorrow, Loki tries to read them using the v11 index, which results in an error.

The solution seems to be: if we're upgrading the charm, we should not change the V13 `from` date at all. To get the older v13 date, we can use _get_schema_config_version_migration_date_from_backup. 

### To Reproduce

TBA.

### Environment

Using Loki rev 220 (uses Loki version 3.7.1) on CK8s.

### Relevant log output

```shell
TBA.
```

### Additional context

_No response_

lucabello:
## Problem

The `_tsdb_versions_migration_dates` property in `src/charm.py` has a date drift bug: when `fresh_install=True`, it **always** returns `datetime.now()` ("today") as the v13 `from` date, **bypassing the backup config entirely**.

Any event that triggers `_configure()` on a subsequent day — pod churn, config-changed, upgrade-charm, etc. — silently shifts the v13 `from` date forward. This makes previously ingested logs inaccessible: Loki tries to look them up via the v11/boltdb index instead of v13/TSDB.

### Concrete failure scenario (no upgrade needed)

1. **Day 1**: Fresh deploy. v13 from = "2026-04-21". Config pushed & backed up.
2. **Day 2**: Pod churns. `pebble-ready` fires → `_configure()`.
   - `fresh_install=True` → early return with v13 from = "2026-04-22"
   - Backup is **never consulted**
   - Day 1 logs indexed under v13/TSDB are now looked up via v11/boltdb → **error**

### StoredState unreliability (K8s pod churn)

Per the [Ops docs](https://ops.readthedocs.io/en/stable/howto/manage-stored-state.html#storing-state-for-the-lifetime-of-the-charm-container-or-machine), `StoredState` persists only for the lifetime of the charm container. When a K8s resource patch applies (or any cause of pod churn), the container is recreated and StoredState is lost. Using StoredState for `fresh_install` and `v13_migration_date` is therefore unreliable as a fallback.

Additionally, a `PathError` on the backup file does not necessarily mean "fresh install" — it could be a transient filesystem issue. Relying on backup absence alone to determine fresh vs upgrade is insufficient.

## Solution

1. **Read backup config first**: The persisted backup on `/loki/chunks/` (persistent storage) is the primary source of truth for migration dates.
2. **Replace StoredState with peer relation databag**: Move `fresh_install` and `v13_migration_date` from `StoredState` to the `replicas` peer relation app databag, which persists for the **lifetime of the application** in the Juju controller DB — surviving pod churns, container recreation, and K8s resource patches.
3. **Dual fresh-vs-upgrade signal**: When computing from scratch, use both the peer databag `fresh_install` flag (set by `_on_upgrade_charm`) and the presence of v12 in the backup as a secondary signal.

### Fallback hierarchy

| Priority | Source | Survives pod churn? |
|----------|--------|-------------------|
| 1. Primary | Backup config on PVC | ✅ (persistent volume) |
| 2. Secondary | Peer relation app databag | ✅ (Juju controller DB) |
| 3. Last resort | Compute from scratch | N/A (first-time only) |

This makes the system **idempotent**: the v13 date is computed exactly once (on first config push) and frozen in both the backup and peer databag. All subsequent calls return the persisted date.

## Changes

- **`src/charm.py`**:
  - Add `_peers`, `_peer_data_get`, `_peer_data_set` helpers for the `replicas` peer relation
  - Move `fresh_install` and `v13_migration_date` from `StoredState` to peer databag
  - `_on_upgrade_charm` writes `fresh_install=false` to peer databag
  - Rewrite `_tsdb_versions_migration_dates` with layered fallback (backup → peer databag → compute)
  - Add INFO logging of schema migration dates for debugging
  - Add WARNING when backup config is missing but chunks directory has data
  - `StoredState` kept only for ephemeral `_stored.status` dict (recalculated every event)
- **`tests/unit/test_charm.py`**: 9 tests for date preservation across pod churns, upgrades, peer databag fallback, and v12 backup signal
- **`tests/unit/test_consumer.py`**: Fix `CosTool` discovery when `platform.processor()` returns empty string

Closes #589
Closes https://github.com/canonical/loki-operators/issues/49

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -319,12 +319,11 @@ class LokiOperatorCharm(CharmBase):
         self._configure()
 
     def _on_upgrade_charm(self, _):
-        # Note: If the pod is rescheduled during startup (e.g., by manual kubectl delete
-        # or resource limit patch), the `upgrade-charm` event might trigger prematurely,
-        # This could incorrectly set the TSDB date to the upgrade day, instead of the following day,
-        # potentially corrupting logs received on that day. Subsequent logs should be healthy.
-        # Documenting for troubleshooting; automatic handling may not be necessary due to the complexity
-        # of guarding against this rare edge case.
+        # Mark as upgrade so that the initial v13 schema date (if not yet persisted in the
+        # backup config) is set to tomorrow rather than today, preserving today's logs that
+        # were written under the previous v11/v12 schema.
+        # The backup config on persistent storage is the primary safeguard against date drift;
+        # this flag only matters on the very first config generation after an upgrade.
         self._stored.fresh_install = False
         self._configure()
 
@@ -898,41 +897,40 @@ class LokiOperatorCharm(CharmBase):
 
     @property
     def _tsdb_versions_migration_dates(self) -> List[Dict[str, str]]:
-        # If v13_migration_date isn't set (due to missing or failed retrieval),
-        # we determine the migration date for v13 schema. This occurs once
-        # during initial setup, as subsequent hooks will get the value from the persisted backup config.
-
-        # If it's a fresh Loki installation, it's safe to set the v13 schema date to today.
-        # This ensures that logs are correctly managed in v13 from the beginning of Loki's operation.
-        # If it's an upgrade scenario, we set the date to tomorrow to accommodate today's logs
-        # that might have been written in the previous v11/v12 schema formats.
-
-        # By default, we assume it's a fresh installation unless state explicitly set to False
-        # by the upgrade hook, indicating an upgrade
+        # Always read migration dates from the persisted backup config first.
+        # The backup (on persistent storage) is the source of truth — this ensures
+        # idempotency across pod churns, upgrades, and any event that triggers _configure().
+        #
+        # The v13 date is computed only once (when no backup exists yet):
+        # - Fresh install: today (no pre-existing logs)
+        # - Upgrade from older charm (v11/v12 only): tomorrow (preserves today's logs
+        #   written under the previous schema)
 
         date_format = "%Y-%m-%d"
         today = datetime.datetime.now(datetime.timezone.utc)
 
-        if self._stored.fresh_install:
-            return [{"version": "v13", "date": today.strftime(date_format)}]
-
         ret = []
 
-        if v12_migration_date := self._get_schema_config_version_migration_date_from_backup("v12"):
+        v12_migration_date = self._get_schema_config_version_migration_date_from_backup("v12")
+        if v12_migration_date:
             ret.append({"version": "v12", "date": v12_migration_date})
 
-        if v13_migration_date := self._get_schema_config_version_migration_date_from_backup("v13"):
+        v13_migration_date = self._get_schema_config_version_migration_date_from_backup("v13")
+        if v13_migration_date:
             ret.append({"version": "v13", "date": v13_migration_date})
             return ret
 
-        tomorrow = today + datetime.timedelta(days=1)
+        # No v13 in backup — first time determining the date.
+        if self._stored.fresh_install:
+            ret.append({"version": "v13", "date": today.strftime(date_format)})
+        else:
+            tomorrow = today + datetime.timedelta(days=1)
+            if v12_migration_date and tomorrow.strftime(date_format) == v12_migration_date:
+                after_tomorrow = tomorrow + datetime.timedelta(days=1)
+                ret.append({"version": "v13", "date": after_tomorrow.strftime(date_format)})
+            else:
+                ret.append({"version": "v13", "date": tomorrow.strftime(date_format)})
 
-        if tomorrow.strftime(date_format) == v12_migration_date:
-            after_tomorrow = tomorrow + datetime.timedelta(days=1)
-            ret.append({"version": "v13", "date": after_tomorrow.strftime(date_format)})
-            return ret
-
-        ret.append({"version": "v13", "date": tomorrow.strftime(date_format)})
         return ret
 
     def _update_datasource_exchange(self) -> None:
--- a/src/charm.py
+++ b/src/charm.py
@@ -149,6 +149,7 @@ class LokiOperatorCharm(CharmBase):
                 retention=to_tuple(ActiveStatus()),
             ),
             fresh_install=True,
+            v13_migration_date="",
         )
 
         self._loki_container = self.unit.get_container(self._name)
@@ -901,7 +902,11 @@ class LokiOperatorCharm(CharmBase):
         # The backup (on persistent storage) is the source of truth — this ensures
         # idempotency across pod churns, upgrades, and any event that triggers _configure().
         #
-        # The v13 date is computed only once (when no backup exists yet):
+        # Stored state (Juju controller DB) is used as a secondary fallback, because
+        # the backup file may be temporarily unavailable during pod recreation (e.g. PVC
+        # not yet remounted when pebble is already connectable).
+        #
+        # The v13 date is computed only once (when neither backup nor stored state has it):
         # - Fresh install: today (no pre-existing logs)
         # - Upgrade from older charm (v11/v12 only): tomorrow (preserves today's logs
         #   written under the previous schema)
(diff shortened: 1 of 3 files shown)


MichaelThamm:
Maybe add that it fixes this issue as well:
- https://github.com/canonical/loki-operators/issues/49

src/charm.py
```diff
@@ -898,41 +898,52 @@ def _charm_logging_ca_cert(self) -> Optional[str]:
 
     @property
     def _tsdb_versions_migration_dates(self) -> List[Dict[str, str]]:
```
MichaelThamm, line 937: Can you convert these comments into a docstring?

tests/unit/test_charm.py
MichaelThamm, line 1: It would be nice to have some GIVEN, WHEN, THEN statements so we can migrate this away from Harness to Scenario easier. We said we should only feature new tests as Scenario not Harness, but given the time constraint I think we can compromise and just add some BDD statements.

tests/unit/test_consumer.py
```diff
@@ -342,6 +343,17 @@ def setUp(self):
         self.sandbox = TempFS("consumer_rule_files", auto_clean=True)
         self.addCleanup(self.sandbox.close)
 
+        # CosTool uses platform.processor() to locate the cos-tool binary.
```
MichaelThamm, line 346: ping @Abuelodelanada I think you saw something similar. Can you comment on this?

src/charm.py
```diff
@@ -319,12 +320,11 @@ def _on_config_changed(self, _):
         self._configure()
 
     def _on_upgrade_charm(self, _):
```
MichaelThamm, line 322: docstring instead of comments?

MichaelThamm (approved):

src/charm.py
```diff
@@ -898,41 +902,54 @@ def _charm_logging_ca_cert(self) -> Optional[str]:
...
             return ret
 
-        tomorrow = today + datetime.timedelta(days=1)
-
-        if tomorrow.strftime(date_format) == v12_migration_date:
-            after_tomorrow = tomorrow + datetime.timedelta(days=1)
-            ret.append({"version": "v13", "date": after_tomorrow.strftime(date_format)})
+        # Backup unavailable or has no v13 — fall back to stored state.
+        if self._stored.v13_migration_date:
+            ret.append({"version": "v13", "date": self._stored.v13_migration_date})
```
sinapah: According to the Ops docs: https://ops.readthedocs.io/en/stable/howto/manage-stored-state.html#storing-state-for-the-lifetime-of-the-charm-container-or-machine, stored state persists for the lifetime of a charm container. If a K8s resource patch applies, the pod churns and stored state is then  lost. Could we still rely on this stored state then?

tests/unit/test_charm.py
```diff
@@ -523,3 +523,198 @@ def test__loki_connection_errors_on_lifecycle_events_appropriately_clear(self):
         self.harness.charm.on.config_changed.emit()
         self.harness.evaluate_status()
         self.assertIsInstance(self.harness.charm.unit.status, ActiveStatus)
+
+
+class TestTsdbVersionsMigrationDates(unittest.TestCase):
```
sinapah, line 529: Should we re-write using Scenario?

lucabello: This PR has an urgent bug fix, so I will push back on this one; let's leave test improvements for after

sed-i:
Thanks @lucabello!

I think there's something I'm missing, but:
- fresh_install seems redundant. Isn't the absence of the config in `/loki/chunks` equivalent to fresh install?
- Isn't a an empty `/loki/chunks` folder equivalent to fresh install?
- If an admin manually deletes the file from the chunks folder, seems like we can't really be confident on the schema dates we need to render. Should we block on "no config and non-empty chunks"?
- Just for safety, should we start logging the schema dates `on.start`?

lucabello:
@sed-i `fresh_install` is not something I added, I'm not sure it's a good approach or not, but I don't understand how that is related to this PR. Things like *Should we block on "no config and non-empty chunks"* are scope explosion.

lucabello:
Given that as @sinapah commented stored state is unreliable, and given the extra comments from @sed-i, I'm removing stored state (which we don't use in other charms, btw) in favor of a peer relation.

src/charm.py
```diff
@@ -372,6 +377,32 @@ def _on_workload_tracing_endpoint_removed(self, _) -> None:
...
 
+    @property
+    def _peers(self) -> Optional[Relation]:
+        """Return the peer relation, or None if not yet available."""
+        return self.model.get_relation("replicas")
+
+    def _peer_data_get(self, key: str) -> str:
+        """Read a value from the peer relation app databag.
+
+        Returns empty string if the peer relation is unavailable or the key is absent.
```
Abuelodelanada: Let's explain first the porpoise of this method, then the edge cases.

```suggestion
        Returns the value of a given key from the peer data or empty string if the peer relation is unavailable or the key is absent.
```

tests/unit/test_charm.py
```diff
@@ -523,3 +523,252 @@ def test__loki_connection_errors_on_lifecycle_events_appropriately_clear(self):
...
+
+        with patch.object(
+            type(charm),
+            "_get_schema_config_version_migration_date_from_backup",
+            return_value="",
+        ):
+            dates = charm._tsdb_versions_migration_dates  # type: ignore
+            v13_entries = [d for d in dates if d["version"] == "v13"]
+            self.assertEqual(len(v13_entries), 1)
+            import datetime
```
Abuelodelanada: Let's remove import from within the method

tests/unit/test_charm.py
```diff
@@ -523,3 +523,252 @@ def test__loki_connection_errors_on_lifecycle_events_appropriately_clear(self):
...
+        """GIVEN a fresh install where the backup config already has a v13 date.
+
+        WHEN _tsdb_versions_migration_dates is accessed after a pod churn
+        THEN the backup date takes precedence over today.
+        """
+        charm = self.harness.charm
+        original_date = "2026-01-15"
+
+        def mock_backup(self_inner, sc_version):
+            return {"v13": original_date, "v12": ""}.get(sc_version, "")
```
Abuelodelanada: Can we extract this mock outside the test method?

tests/unit/test_charm.py
```diff
@@ -523,3 +523,252 @@ def test__loki_connection_errors_on_lifecycle_events_appropriately_clear(self):
...
+
+        WHEN _tsdb_versions_migration_dates is accessed
+        THEN the persisted backup date must be used.
+        """
+        charm = self.harness.charm
+        self._set_peer_data("fresh_install", "false")
+        original_date = "2026-01-15"
+
+        def mock_backup(self_inner, sc_version):
+            return {"v13": original_date, "v12": ""}.get(sc_version, "")
```
Abuelodelanada: Can we extract the mock outside the test method?

tests/unit/test_charm.py
```diff
@@ -523,3 +523,252 @@ def test__loki_connection_errors_on_lifecycle_events_appropriately_clear(self):
...
+
+        with patch.object(
+            type(charm),
+            "_get_schema_config_version_migration_date_from_backup",
+            return_value="",
+        ):
+            dates = charm._tsdb_versions_migration_dates  # type: ignore
+            v13_entries = [d for d in dates if d["version"] == "v13"]
+            self.assertEqual(len(v13_entries), 1)
+            import datetime
```
Abuelodelanada: Let's remove the import

tests/unit/test_charm.py
```diff
@@ -523,3 +523,252 @@ def test__loki_connection_errors_on_lifecycle_events_appropriately_clear(self):
...
+        """GIVEN a v13 date in the backup config but not yet in peer databag.
+
+        WHEN _tsdb_versions_migration_dates is accessed
+        THEN the backup date is also persisted into the peer databag for future fallback.
+        """
+        charm = self.harness.charm
+        original_date = "2026-01-15"
+
+        def mock_backup(self_inner, sc_version):
+            return {"v13": original_date, "v12": ""}.get(sc_version, "")
```
Abuelodelanada: Can we extract the mock outside?

tests/unit/test_charm.py
```diff
@@ -523,3 +523,252 @@ def test__loki_connection_errors_on_lifecycle_events_appropriately_clear(self):
...
+
+        WHEN _tsdb_versions_migration_dates is accessed
+        THEN v12 presence signals an upgrade → v13 = tomorrow (not today).
+        """
+        charm = self.harness.charm
+        # Clear any v13 date that may have been set during setUp hooks.
+        self._set_peer_data("v13_migration_date", "")
+
+        def mock_backup(self_inner, sc_version):
+            return {"v12": "2025-06-01", "v13": ""}.get(sc_version, "")
```
Abuelodelanada: Let's extract the mock outside

tests/unit/test_charm.py
```diff
@@ -523,3 +523,252 @@ def test__loki_connection_errors_on_lifecycle_events_appropriately_clear(self):
...
+
+        with patch.object(
+            type(charm),
+            "_get_schema_config_version_migration_date_from_backup",
+            mock_backup,
+        ):
+            dates = charm._tsdb_versions_migration_dates  # type: ignore
+            v13_entries = [d for d in dates if d["version"] == "v13"]
+            self.assertEqual(len(v13_entries), 1)
+            import datetime
```
Abuelodelanada: Let's remove the import from the test function

Abuelodelanada (changes requested):

Abuelodelanada (approved):

PR #593, opened 2026-04-23
feat: remove fresh_install

sinapah:
The idea is to build on #591 and remove the need for fresh_install. This PR refactors how Loki's schema migration dates are determined and simplifies related logic, removing reliance on peer databag state and focusing on the presence of log chunks and backup config. The goal is to make migration date selection more robust, predictable, and easier to reason about, especially during upgrades, fresh installs, and pod churns. Tests are updated to reflect the new logic and to remove now-unnecessary peer databag handling.

**Schema migration logic changes:**

* The `_tsdb_versions_migration_dates` property in `src/charm.py` is rewritten to base its logic solely on whether log chunks exist and the presence of migration dates in the backup config, removing the use of peer databag flags like `fresh_install` and `v13_migration_date`. The migration date for v13 is now computed as follows:
  - If no chunks exist: use v13 from backup if present, otherwise today.
  - If chunks exist: use v13 from backup if present; else, if v12 is present, set v13 to one day after v12; otherwise, set v13 to tomorrow.
* The `_on_upgrade_charm` handler no longer sets a `fresh_install` flag in the peer relation databag, as this is now obsolete.
* A new helper method `_chunks_non_empty` is added to detect whether Loki has written any log chunks, which is now central to determining schema migration logic.

**Testing and code cleanup:**

* All test cases in `tests/unit/test_charm.py` are updated to use the new chunk-based logic, removing all peer databag helpers and references. Test names and docstrings are updated for clarity, and tests now patch `_chunks_non_empty` to control the test scenario.
* Tests for fallback to peer databag and for syncing backup dates to the peer databag are removed, as this fallback is no longer used.
* Unused imports and obsolete code related to the peer databag are removed for clarity and maintainability.

These changes make the schema migration process more transparent and less error-prone by relying only on persistent backup and observable chunk state, and they significantly simplify both the charm logic and its test suite.## Issue

## Upgrade Notes

Note there is a chance that the `fresh_install` KV remains in peer data after an upgrade from the previous rev of Loki to the one featuring this change occurs.


sinapah:
Note: I haven't migrated any of the tests that came from #591 to Scenario and I stuck with Harness for this PR as well.

sinapah:
And the unit test has been moved to Scenario in #594!

lucabello:
We had problems prior with `/loki/chunks` giving a `PathError` sometimes;  I think removing the fallback on `v13_migration_date` would cause the same problem to happen, would it not?

src/charm.py
```diff
@@ -843,6 +832,22 @@ def _remove_alert_rules_files(self) -> None:
...
+
+        In single-tenant mode, Loki stores chunks exclusively under CHUNKS_DIR/fake/.
+        Other files (e.g. the backup config) live at the top level of CHUNKS_DIR and
+        are not chunks, so we check the tenant subdirectory directly.
+        Ref: https://grafana.com/docs/loki/latest/operations/storage/filesystem/
+        """
+        # "fake" is the fixed tenant ID used when auth_enabled=False (single-tenant mode).
+        tenant_chunks_dir = os.path.join(CHUNKS_DIR, "fake")
+        try:
+            files = self._loki_container.list_files(tenant_chunks_dir)
```
sed-i: We sometimes see units go into error state due to timeouts on the pebble socket.
I wonder if we'd be better off going via the mount point directly instead of using pebble, e.g.
```python
self.meta.storages["loki-chunks"].location
# or
self.model.storages["loki-chunks"]
```
Ref: https://documentation.ubuntu.com/ops/latest/howto/manage-storage/

src/charm.py
```diff
@@ -935,69 +940,55 @@ def _charm_logging_ca_cert(self) -> Optional[str]:
...
-        The peer relation app databag is used as a secondary fallback, because
-        the backup file may be temporarily unavailable (e.g. PathError during pod
-        recreation). Unlike StoredState, the peer databag persists for the lifetime
-        of the application and survives pod churns.
+        # Migration date selection is driven by whether Loki has already stored log chunks:
+        #
+        # No chunks  → v13 = value from backup config if present, else today
+        # Chunks exist → v13 from backup config if present;
+        #                else v13 = v12 + 1 day if v12 known;
+        #                else v13 = tomorrow
```
sed-i, line 1008: It's a bit difficult to grok the cases here and below. For future readings, can we add a feature file with a table covering the various cases? E.g.

| `num_chunk_file` | `v12_date` | `v13_date`
| --- | --- | --- |
| 0 | None | None |
| 0 | (something) | None |
| etc. | | |

That could give document the feature well and add confidence to the implementation.

src/charm.py
lucabello: Addressed in [`945755d` (this PR)](https://github.com/canonical/loki-k8s-operator/pull/593/commits/945755d92a6859dfd7f3494fe9e61ab0c038abd7)

lucabello (approved):

PR #594, opened 2026-04-23
chore: rewrite migration unit tests in scenario

sinapah:
## Issue

Tandem to #591 and #593; it rewrites the Harness tests needed in those PRs using Scenario.


sinapah:
Note: there are some new tests that should be moved to Scenario and removed. (DONE)

lucabello (approved):

PR #601, opened 2026-04-29
feat: migrate charm-tracing to ops[tracing]

lucabello:
## Summary

Migrate charm tracing from the old `charm_tracing` charm library (`@trace_charm` decorator) to the new `ops[tracing]` approach using `ops.tracing.Tracing`.

### Changes
- Replaced `@trace_charm` decorator with `ops.tracing.Tracing` instantiation
- Updated dependencies to use `ops[tracing]`, removed pinned opentelemetry deps
- Added `cargo` to build-packages (pydantic-core dependency)
- Added `receive-ca-cert` relation (interface: `certificate_transfer`)
- Removed old `charm_tracing.py` library (kept `tracing.py` for workload-tracing)
- Updated test fixtures to remove `charm_tracing_disabled()` usage

### Motivation
The `ops[tracing]` approach is the recommended way to add charm tracing as of ops 2.20.0. It replaces the third-party charm_tracing charm library with first-party support built into ops.


pyproject.toml
```diff
@@ -75,7 +68,7 @@ lint.pydocstyle.convention = "google"
 # Static analysis tools configuration
 [tool.pyright]
 extraPaths = ["lib"]
-pythonVersion = "3.8"
+pythonVersion = "3.10"
```
mmkay: On purpose?

src/charm.py
```diff
@@ -242,15 +230,14 @@ def __init__(self, *args):
         self.dashboard_provider = GrafanaDashboardProvider(self)
 
         self.catalogue = CatalogueConsumer(charm=self, item=self._catalogue_item)
-        self.charm_tracing = TracingEndpointRequirer(
-            self, relation_name="charm-tracing", protocols=["otlp_http"]
+        self.tracing = ops.tracing.Tracing(  # type: ignore[union-attr]
```
mmkay: I'd go with `self.charm_tracing` again, to distinguish with workload tracing.

mmkay (approved):

pyproject.toml
lucabello: Yes, but I'm actually taking this back I think :)

lucabello: I'm splitting the static checks into two, one for charm code and tests, and one for the libraries, so those can stay on 3.8.

src/charm.py
lucabello: Done!

charmcraft.yaml
```diff
@@ -50,7 +50,7 @@ parts:
   charm:
     source: .
     plugin: uv
-    build-packages: [git]
+    build-packages: [cargo, git]
```
sed-i: Curious why we need this when the charm is built with the uv plugin?

charmcraft.yaml
```diff
@@ -151,6 +151,12 @@ requires:
     limit: 1
     description: |
       Enables sending workload traces to a distributed tracing backend such as Tempo.
+  receive-ca-cert:
```
sed-i, line 154: Fetch-lib, etc. for this?

Abuelodelanada (approved):

charmcraft.yaml
lucabello: This is a leftover, good catch!

charmcraft.yaml
lucabello: On it!

PR #616, opened 2026-05-28
fix: set `promtail_binary_zip_url` on `relation_changed` as fallback for reconcile-pattern charms

### LINKED ISSUE #615: `LokiPushApiProvider` does not set `promtail_binary_zip_url` when used with the reconcile pattern
## Bug Description

`LokiPushApiProvider` only sets `promtail_binary_zip_url` in application relation data during the [`_on_logging_relation_joined`](https://github.com/canonical/loki-k8s-operator/blob/0075d8f2072d7c47ae72cfc9e99d281d3e4076bc/lib/charms/loki_k8s/v1/loki_push_api.py#L969) event handler. This works for charms using the traditional event-handler pattern, but fails for charms using the **reconcile pattern** (all logic gated behind `container.can_connect()`).

In the reconcile pattern, if `relation_joined` fires **before** the container is ready, the charm returns early without instantiating `LokiPushApiProvider`. The [observer for `relation_joined`](https://github.com/canonical/loki-k8s-operator/blob/main/lib/charms/loki_k8s/v1/loki_push_api.py#L976) is never registered, the handler never runs, and `promtail_binary_zip_url` is never written to app data. Since `relation_joined` only fires once, the URL is permanently missing.

As a result, any `LogProxyConsumer` client on the other side of the relation cannot download promtail and never starts pushing logs.

## To Reproduce

1. Deploy a charm that uses `LokiPushApiProvider` with the reconcile pattern:
   ```python
   class MyCharm(CharmBase):
       def __init__(self, *args):
           super().__init__(*args)
           if not self.unit.get_container("workload").can_connect():
               return  # early exit
           self._reconcile()

       def _reconcile(self):
           self.loki_provider = LokiPushApiProvider(self, relation_name="logging", ...)
           self.loki_provider.update_endpoint(...)
   ```

2. Deploy a charm that uses `LogProxyConsumer` (e.g. `flog-k8s`) and integrate them.

3. Observe that `promtail_binary_zip_url` is missing from the provider's application data:
   ```
   --- provider endpoint: logging ---
     application data:
       <empty>
     unit data:
       provider/0:
         endpoint: {"url": "http://..."}
   ```

4. The `LogProxyConsumer` client [finds no binary URL and exits without installing promtail](https://github.com/canonical/loki-k8s-operator/blob/main/lib/charms/loki_k8s/v1/loki_push_api.py#L2194-L2198):
   ```python
   promtail_binaries = json.loads(
       relation.data[relation.app].get("promtail_binary_zip_url", "{}")
   )
   if not promtail_binaries:
       return  # never downloads promtail
   ```

## Root Cause

The sequence of events that triggers the bug:

1. `relation-joined` fires on the provider charm.
2. The workload container is not yet ready (`can_connect()` returns `False`).
3. The charm exits early without creating `LokiPushApiProvider`.
4. The `_on_logging_relation_joined` handler is never registered, never runs.
5. `promtail_binary_zip_url` is never set in app data.
6. Later, `pebble-ready` fires, the container is connectable, `_reconcile()` runs.
7. `LokiPushApiProvider` is created and `update_endpoint()` sets unit data (the endpoint URL).
8. But `_on_logging_relation_joined` does not fire again (it's a one-time event).
9. `promtail_binary_zip_url` remains permanently unset.

The relevant code is in [`LokiPushApiProvider._on_logging_relation_joined`](https://github.com/canonical/loki-k8s-operator/blob/0075d8f2072d7c47ae72cfc9e99d281d3e4076bc/lib/charms/loki_k8s/v1/loki_push_api.py#L969):

```python
# Only called on relation_joined (one-time event)
def _on_logging_relation_joined(self, event: RelationJoinedEvent):
        event.relation.data[self._charm.app].update(self._promtail_binary_url)
```

## Environment

- Juju 3.6 on Kubernetes
- `loki_push_api` library v1, LIBPATCH 22

## Workaround

Provider charms using the reconcile pattern can work around this by manually setting the app data after calling `update_endpoint()`:

```python
if charm.unit.is_leader():
    for relation in charm.model.relations.get("logging", []):
        if not relation.data[charm.app].get("promtail_binary_zip_url"):
            relation.data[charm.app].update(loki_provider._promtail_binary_url)
```

Note: this accesses the private `_promtail_binary_url` property, which is not ideal.

## References

- https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/172
- https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/120

Abuelodelanada:
# Issue

This PR fixes #615

Charms using the reconcile pattern (the way `opentelemetry-collector-k8s-operator` implements it) never receive `promtail_binary_zip_url` in relation data. This prevents `LogProxyConsumer` clients from downloading promtail, breaking the entire log forwarding pipeline.

Root cause: `promtail_binary_zip_url` is only set in `_on_logging_relation_joined`, but reconcile-pattern charms guard their `__init__` with `can_connect()` checks. When `relation_joined` fires before the workload container is ready, the charm returns early, `LokiPushApiProvider` is never instantiated, and the handler never runs. Subsequent `relation_changed` events do invoke the provider, but never set the URL.

Ref: https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/172

# Solution

Add a fallback in `_process_logging_relation_changed` (called on every `relation_changed` and from `update_endpoint` loops) that sets `promtail_binary_zip_url` in app data if it is missing:

```python
    if not relation.data[self._charm.app].get("promtail_binary_zip_url"):
```
This is idempotent — it only writes when the key is absent, so it has no effect on charms where relation_joined already set it.

# Context

The "reconcile pattern" (opentelemetry-collector-operator flavour) defers all logic to a single `_reconcile()` method and exits early from event handlers if the workload container is not yet connectable. This means one-shot lifecycle events like `relation_joined` are effectively lost. Libraries that set relation data exclusively in `relation_joined` are incompatible with this pattern.

# Testing Instructions

Run unit tests

```shell
tox -e unit -- tests/unit/test_charm.py::TestAppRelationData -v
```

The new test `test_promtail_url_set_on_relation_changed_if_missing` verifies that:

1. After promtail_binary_zip_url is cleared from app data (simulating missed relation_joined)
2. A subsequent relation_changed event restores it with a valid HTTP URL
To verify the fix prevents regression, remove lines 1050-1055 from `lib/charms/loki_k8s/v1/loki_push_api.py` and observe the test fail with:

AssertionError: 'promtail_binary_zip_url' not found in {}

# Upgrade Notes

No upgrade steps required. The fix is purely additive — it fills missing relation data on the next relation_changed event. Existing deployments where promtail_binary_zip_url is already set are unaffected (the conditional check skips the write). Charms consuming the library will pick up the fix on the next charmcraft fetch-lib / update-libs run.

### DIFF
--- a/lib/charms/loki_k8s/v1/loki_push_api.py
+++ b/lib/charms/loki_k8s/v1/loki_push_api.py
@@ -544,7 +544,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 24
+LIBPATCH = 25
 
 PYDEPS = ["cosl"]
 
@@ -1046,6 +1046,14 @@ class LokiPushApiProvider(Object):
         """
         relation.data[self._charm.unit]["public_address"] = socket.getfqdn() or ""
         self.update_endpoint(relation=relation)
+
+        # Ensure promtail binary URL is set in app data. This is normally done on
+        # relation_joined, but charms using the reconcile pattern may miss that event
+        # if the workload container is not yet ready when the relation is first established.
+        if self._charm.unit.is_leader():
+            if not relation.data[self._charm.app].get("promtail_binary_zip_url"):
+                relation.data[self._charm.app].update(self._promtail_binary_url)
+
         return self._should_update_alert_rules(relation)
 
     @property
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -441,6 +441,37 @@ class TestAppRelationData(unittest.TestCase):
         url = promtail_binaries["amd64"]["url"]
         self.assertTrue(url.startswith("http"))
 
+    def test_promtail_url_set_on_relation_changed_if_missing(self):
+        """Scenario: promtail_binary_zip_url is set on relation_changed as a fallback.
+
+        Charms using the reconcile pattern may miss the relation_joined event if the
+        workload container is not yet ready. In that case, promtail_binary_zip_url
+        must be set on the next relation_changed event.
+
+        Ref: https://github.com/canonical/loki-k8s-operator/issues/615
+        """
+        # GIVEN promtail_binary_zip_url was removed from app data (simulating missed relation_joined)
+        rel_data = self.harness.get_relation_data(self.rel_id, self.harness.charm.app)
+        self.assertIn("promtail_binary_zip_url", rel_data)
+        self.harness.update_relation_data(
+            self.rel_id, self.harness.charm.app.name, {"promtail_binary_zip_url": ""}
+        )
+        rel_data = self.harness.get_relation_data(self.rel_id, self.harness.charm.app)
+        self.assertNotIn("promtail_binary_zip_url", rel_data)
+
+        # WHEN a relation_changed event fires (e.g. consumer updates its metadata)
+        self.harness.update_relation_data(
+            self.rel_id, "consumer", {"metadata": json.dumps(METADATA)}
+        )
+
+        # THEN promtail_binary_zip_url is set again in app data
+        rel_data = self.harness.get_relation_data(self.rel_id, self.harness.charm.app)
+        self.assertIn("promtail_binary_zip_url", rel_data)
+        self.assertNotEqual(rel_data["promtail_binary_zip_url"], "")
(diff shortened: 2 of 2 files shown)


MichaelThamm (approved):
Harness 🤮, but I will approve out of kindness.

PR #622, opened 2026-06-05
feat: filter out invalid rule groups and save valid rules to disk

### LINKED ISSUE #623: Loki Push API should correctly filter out invalid rules instead of removing all alerts due to one invalid rule
### Enhancement Proposal

Currently, this charm relies on the Loki Push API lib to read alert rules. The library writes is supposed to validate alert rules. If there are any invalid rule groups, the charm refuses to write ANY groups to disk, even those groups that are valid.

The behaviour should be that the Loki Push API lib loads all rules, validates them, and only returns the valid ones.

Then, the charm can save those valid rules to disk.

However, there should still be a status message as well as a status of blocked in that case.

sinapah:
## Issue

Fixes #623.
Related to: https://github.com/canonical/loki-operators/issues/80. Once this PR goes in, the consequent `fetch-lib` action will fix this issue as well. However, let's not close the issue until we've added a unit test in that repo too.

## Solution

We remove the `_has_alert_rule_errors` method. This method used to check whether ANY logging relations had validation errors. If even a single one did, it would set status=blocked. Instead, the desired behaviour is that we first read all rule groups coming from relation data.  Then, the Loki Push API library iterates over each available relation and validates the rule groups. If a group is invalid, it should `continue`, without writing it to the `alerts` object it will eventually return to the charm. 

Now, we filter out invalid rules, save valid rules to disk, and still set blocked to capture the attention of the admin.
## Context

Now, both the validation of rules on disk and in databag contribute to setting Blocked status in the charm. The `Blocked` status in this context is purely cosmetic: allowing the charm to operate normally, avoiding a charm (with invalid alert rules) side-effecting in Loki.

### DIFF
--- a/lib/charms/loki_k8s/v1/loki_push_api.py
+++ b/lib/charms/loki_k8s/v1/loki_push_api.py
@@ -544,7 +544,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 25
+LIBPATCH = 26
 
 PYDEPS = ["cosl"]
 
@@ -1174,7 +1174,6 @@ class LokiPushApiProvider(Object):
                 try:
                     metadata = json.loads(relation.data[relation.app]["metadata"])
                     identifier = JujuTopology.from_dict(metadata).identifier
-                    alerts[identifier] = self._tool.apply_label_matchers(alert_rules)  # type: ignore
 
                 except KeyError as e:
                     logger.debug(
@@ -1195,7 +1194,7 @@ class LokiPushApiProvider(Object):
                     relation.data[self._charm.app]["event"] = json.dumps({"errors": errmsg})
                 continue
 
-            alerts[identifier] = alert_rules
+            alerts[identifier] = self._tool.apply_label_matchers(alert_rules)  # type: ignore
 
         return alerts
 
--- a/lib/charms/loki_k8s/v1/loki_push_api.py
+++ b/lib/charms/loki_k8s/v1/loki_push_api.py
@@ -1188,13 +1188,18 @@ class LokiPushApiProvider(Object):
                 )
                 continue
 
+            alerts[identifier] = self._tool.apply_label_matchers(alert_rules)  # type: ignore
+
             _, errmsg = self._tool.validate_alert_rules(cast(OfficialRuleFileFormat, alert_rules))
             if errmsg:
+                logger.error(f"Invalid alert rule file: {errmsg}")
+                if alerts[identifier]:
+                    del alerts[identifier]
                 if self._charm.unit.is_leader():
                     relation.data[self._charm.app]["event"] = json.dumps({"errors": errmsg})
                 continue
 
-            alerts[identifier] = self._tool.apply_label_matchers(alert_rules)  # type: ignore
+            alerts[identifier] = alert_rules
 
         return alerts
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -854,6 +854,13 @@ class LokiOperatorCharm(CharmBase):
 
         alerts = self.loki_provider.alerts
 
+        # If there aren't any alerts, we can just clean it and move on
+        # The alerts at this point are guaranteed to be valid,
+        # since library filters out invalid rules before returning the dictionary.
+        if alerts:
+            self._generate_alert_rules_files()
+            self._check_alert_rules()
+
         # Check if any relations reported alert rule validation errors.
         # The provider's alerts property writes {"errors": ...} to relation data
         # when cos-tool rejects a rule. The charm should go blocked in that case.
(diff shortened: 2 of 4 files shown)


lib/charms/loki_k8s/v1/loki_push_api.py
```diff
@@ -1174,7 +1174,6 @@ def alerts(self) -> dict:  # noqa: C901
                 try:
                     metadata = json.loads(relation.data[relation.app]["metadata"])
                     identifier = JujuTopology.from_dict(metadata).identifier
-                    alerts[identifier] = self._tool.apply_label_matchers(alert_rules)  # type: ignore
```
sinapah, line 1177: The fact that the alert rules were written to the dict before validation had taken place meant that the invalid rules would be saved to disk, resulting in https://github.com/canonical/loki-operators/issues/80.

This is because after a failed validation, we wouldn't remove the rules from the dict.

src/charm.py
sinapah, line 856: Small note that this can _potentially_ be improved further. For example, why should we remove all alert rules and "regenerate them"? We don't do this in the HA charm. I've left it unchanged for now.

tests/unit/test_alert_rule_filtering.py
```diff
@@ -0,0 +1,102 @@
...
+    "rel1_invalid, rel2_invalid, expected_group_names",
+    [
+        (False, False, {"rel1-group", "rel2-group"}),
+        (False, True, {"rel1-group"}),
+        (True, True, set()),
+    ],
+)
+@patch("charms.loki_k8s.v1.loki_push_api.CosTool.validate_alert_rules", side_effect=_validate_rule_file)
+@patch("charm.LokiOperatorCharm._check_alert_rules", return_value=None)
+def test_relation_level_filtering(_, __, context, loki_container, rel1_invalid, rel2_invalid, expected_group_names):
```
MichaelThamm: The signature is quite smelly. Could you try something a bit cleaner like this:
```python
@pytest.mark.parametrize(
    "rel1_invalid, rel2_invalid, expected_group_names",
    [
        (False, False, {"rel1-group", "rel2-group"}),
        (False, True, {"rel1-group"}),
        (True, True, set()),
    ],
)
def test_relation_level_filtering(
    mocker, context, loki_container, rel1_invalid, rel2_invalid, expected_group_names
):
    # Setup patches cleanly inside the test body without polluting the signature
    mocker.patch(
        "charms.loki_k8s.v1.loki_push_api.CosTool.validate_alert_rules", 
        side_effect=_validate_rule_file
    )
    mocker.patch(
        "charm.LokiOperatorCharm._check_alert_rules", 
        return_value=None
    )

    # Act & Assert ...
```

MichaelThamm: Requires https://pypi.org/project/pytest-mock/

MichaelThamm: Note that this is a nit, and if you find a better python-native way to mock this, that works as well

tests/unit/test_alert_rule_filtering.py
sinapah, line 1: The structure of this unit test is by my design. I wanted to have both charm level and Loki Push API level logic tested in the same file and tests. The separation of the tests is by design, since this is one of those features that would be confusing to the reader if implemented via parametrization.

MichaelThamm (approved):

sinapah:
To manually test this change, first deploy the following bundle:
```yaml
bundle: kubernetes
applications:
  cos-config:
    charm: cos-configuration-k8s
    channel: dev/edge
    revision: 102
    resources:
      git-sync-image: 36
    scale: 1
    options:
      git_branch: main
      git_repo: https://github.com/sinapah/cos-config
      git_ssh_key_secret: secret://d8k306fmp25c76c0n6ug/key
      prometheus_alert_rules_path: prometheus_alert_rules
    constraints: arch=amd64
    storage:
      content-from-git: kubernetes,1,1024M
    trust: true
  flog:
    charm: flog-k8s
    channel: latest/edge
    revision: 13
    base: ubuntu@20.04/stable
    resources:
      workload-image: 4
    scale: 1
    constraints: arch=amd64
    trust: true
  loki:
    charm: local:loki-k8s-1
    scale: 1
    constraints: arch=amd64
    storage:
      active-index-directory: kubernetes,1,1024M
      loki-chunks: kubernetes,1,1024M
    trust: true
relations:
- - flog:log-forwarder
  - loki:logging

```

The Loki deployed should be the locally packed charm from this branch. The bundle deploys Flog, which contains valid rules, and a COS Config charm preconfiged to point to a repo with two rules: a valid one and an invalid one, forming a group.

Initially, Loki is only related to Flog, so the charm must be `active/idle`. Also, the rule file should be present at `/loki/rules/fake/` in the `loki` container.
```
root@loki-0:/# ls /loki/rules/fake/
test_1fa6431e_flog_alert.rules
```
Now, relate Loki and COS Config:
```
juju integrate loki cos-config:loki-config
```

Since COS Config contains invalid rules, the charm should now be blocked. 
```
loki/0*        blocked   idle   10.1.0.125         Invalid alert rules. See debug-log
```
Also, the only rule that should show up on the Loki FS must be the valid group from Flog.

```
ls /loki/rules/fake/
test_1fa6431e_flog_alert.rules
```

If you now `juju remove-relation loki cos-config`, the charm should become active/idle again.

MichaelThamm (changes requested):
We need to be certain that if a charm with a bad alert rule is related to Loki, that it doesn't side-effect in Loki and causing the charm to no longer operate effectively. In other words: if we set Blocked status when Loki notices an invalid alert rule, that is fine, BUT we should not `return` early in the charm's operations. The Blocked status should be purely cosmetic

MichaelThamm (approved):

PR #625, opened 2026-06-08
test(integration): migrate from pytest-operator to jubilant

lucabello:
## Summary

Migrate integration tests from pytest-operator to jubilant and pytest-jubilant (>=2.0). Tests are now synchronous and use `sh` for charm packing.

## Changes

- Replace `juju` and `pytest-operator` with `jubilant` and `pytest-jubilant>=2.0`
- Convert all async tests to sync using jubilant's `Juju` class
- Use `sh.charmcraft.pack()` for charm building instead of `ops_test.build_charm()`
- Wait for both `all_active` and `all_agents_idle` on all waits for robustness
- Replace `fs.TempFS` in unit tests with stdlib `tempfile.TemporaryDirectory` wrapper
- Remove `fs` and `setuptools` dependencies

## Bug Fix: Add `job` label to LogForwarder logs

During the migration, we discovered that alert rule tests were **false positives** that never actually worked correctly.

### Root Cause

The `cosl` library replaces `%%juju_topology%%` placeholders in alert rules with `job=~".+"` for LogQL expressions (see `cosl/rules.py:459`). This regex **requires** logs to have a `job` label present.

However:
- `LogProxyConsumer` (deprecated) correctly adds a `job` label to logs
- `LogForwarder` (via `_PebbleLogClient`) did **not** include a `job` label
- The `loki-tester` charm (using `LokiPushApiConsumer` directly) also did not send a `job` label

The tests appeared to pass because the test charm `uv.lock` files were outdated (pinned to `cosl==1.0.0` instead of `1.9.1`), masking the issue.

### Fix

1. **Library fix**: Add `job` label to `_PebbleLogClient._build_log_target` using the same format as `LogProxyConsumer`: `juju_{topology.identifier}` (e.g., `juju_test-model_6ae23f62_loki-tester`)

2. **Test charm fix**: Add `job` label to `loki-tester` charm's log tags

3. **Dependency updates**: Refresh `uv.lock` files for all three test charms (`loki-tester`, `log-proxy-tester`, `log-forwarder-tester`)

4. **LIBPATCH**: Bumped from 26 to 27


lucabello:
This should be merged after #626 as it modifies a unit test to consider the `job` label

tests/integration/conftest.py
```diff
@@ -1,152 +1,149 @@
...
-        if fname in store.keys():
-            ret = store[fname]
-        else:
-            logger.info("Return for {} not cached".format(fname))
-            ret = await func(*args, **kwargs)
-            store[fname] = ret
-        logger.info("Finished: {} in: {} seconds".format(fname, datetime.now() - start_time))
-        return ret
-
-    return wrapper
```
sinapah, line 53: Why remove this? Can be useful for local testing, no?

tests/integration/helpers.py
```diff
@@ -64,41 +60,14 @@ async def loki_rules(ops_test, app_name) -> dict:
         return {}
```
sinapah, line 65: I'd say this one can you a `retry` as well, but it's not too critical.

tests/integration/test_workload_tracing.py
```diff
@@ -25,40 +26,37 @@
...
-        "tempo-worker-k8s", application_name="tempo-worker", channel=cos_channel, trust=True
-    )
-    await ops_test.model.deploy(
-        "tempo-coordinator-k8s",
-        application_name=TEMPO_APP_NAME,
-        channel=cos_channel,
+    juju.deploy("tempo-worker-k8s", "tempo-worker", channel=cos_channel, trust=True)
+    juju.deploy("tempo-coordinator-k8s", TEMPO_APP_NAME, channel=cos_channel, trust=True)
+    # Set up minio and s3-integrator
+    juju.deploy(
```
sinapah, line 42: Reminder to myself: https://github.com/canonical/loki-k8s-operator/issues/635

sinapah (approved):
LGTM! I think another set of 👀 will also be valuable here.

tests/integration/conftest.py
lucabello: Feel free to re-add it if needed

lucabello: The caching in `timed_memoizer` was load-bearing with pytest-operator because async fixtures could be called multiple times and the global `Store` dict provided session-level caching across modules. With jubilant, pytest's built-in `scope="module"` already ensures each fixture runs once per module, and since the `juju` fixture creates a fresh `temp_model()` per module anyway, cross-module caching wouldn't be meaningful. So the caching is a no-op here — bringing it back would add code that never actually saves any work.

PR #628, opened 2026-06-09
refactor: Jubilant tests

MichaelThamm:
## Solution

**Strengthen alert assertions to avoid vacuous passes**
- `tests/integration/test_alert_rules_fire.py`: Added an explicit `assert alerts, "no alerts fired in Loki after triggering log-error"` before the label-key check. Previously the `all(... for alert in alerts)` check was vacuously `True` when no alerts had fired, so the test could pass without Loki ever producing an alert. The scale-up case (`test_loki_scales_up`) gained an equivalent non-empty guard.
- `tests/integration/test_upgrade_charm_retains_alerts.py`: Added `assert rules_before_upgrade, "no alert rules present in Loki before upgrade"` immediately after fetching the pre-upgrade rules. Without it, a `{} == {}` comparison would pass even if no rules were ever present, defeating the purpose of the retention test.

**Surface swallowed errors in alert-polling helpers**
- `tests/integration/helpers.py`: In `loki_alerts` and `get_alertmanager_alerts`, replaced the silent `except Exception: pass` with `except Exception as e: logger.debug(...)` (including the URL and error). Transient fetch failures are now visible in the logs instead of silently masquerading as "no alerts."

**Fix file-handle leak in `oci_image`**
- `tests/integration/helpers.py`: `oci_image` now reads `charmcraft.yaml` via a `with open(...) as file:` context manager instead of `open(charmcraft_file).read()`, so the file is closed deterministically. No behavior change.

## Testing Instructions

See CI runs

### DIFF
--- a/tests/integration/test_alert_rules_fire.py
+++ b/tests/integration/test_alert_rules_fire.py
@@ -70,8 +70,8 @@ def test_loki_scales_up(juju: jubilant.Juju):
         loki_alerts(juju, "loki", unit_num=i) for i in range(3)
     ]
 
-    for unit_num, unit_alerts in enumerate(alerts_per_unit):
-        assert unit_alerts, f"no alerts fired in Loki on unit {unit_num} after triggering log-error"
+    for unit_alerts in alerts_per_unit:
+        assert unit_alerts, f"no alerts fired in Loki units after triggering log-error"
         assert all(
             key in alert["labels"].keys()
             for key in ["juju_application", "juju_model", "juju_model_uuid"]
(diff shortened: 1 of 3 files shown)


tests/integration/test_alert_rules_fire.py
```diff
@@ -69,7 +70,8 @@ def test_loki_scales_up(juju: jubilant.Juju):
         loki_alerts(juju, "loki", unit_num=i) for i in range(3)
     ]
 
-    for unit_alerts in alerts_per_unit:
+    for unit_num, unit_alerts in enumerate(alerts_per_unit):
+        assert unit_alerts, f"no alerts fired in Loki on unit {unit_num} after triggering log-error"
```
MichaelThamm:
```suggestion
    for unit_alerts in alerts_per_unit:
        assert unit_alerts, f"no alerts fired in Loki units after triggering log-error"
```

PR #664, opened 2026-08-10
feat!: set a default retention period of 15 days

sinapah:
## Issue

This charm (and the related Loki HA coordinator charm) currently disabled deletion by default, causing logs to be retained indefinitely by default. This has led to multiple deployments where users have had to patch their PVCs to remediate exhausted disk space.

## Solution

Similar to [Prometheus](https://github.com/canonical/prometheus-k8s-operator/blob/main/charmcraft.yaml#L206-L212), we'll set the default to 15 days.

## Context

Previously we had no retention so by default logs would be retained for ever. Since we're changing, I'm marking this as a breaking change.

## Testing Instructions

Pack the charm and deploy. The workload config file should be:
```yaml
...
compactor:
  delete_request_store: filesystem
  retention_enabled: true
  working_directory: /loki/compactor
...
limits_config:
  allow_structured_metadata: true
  ingestion_burst_size_mb: 15.0
  ingestion_rate_mb: 4.0
  per_stream_rate_limit: 4MB
  per_stream_rate_limit_burst: 15MB
  retention_period: 15d
  split_queries_by_interval: '0'
...
```

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -212,15 +212,13 @@ config:
       default: 15
     retention-period:
       description: |
-        Sets a global retention period, in days, for log streams in Loki.
-        The minimum retention period is 1 day, and a value of 0 (default) means "infinity" (disables retention).
-        Loki will not be cleaning up logs if duration is set to 0.
-        This config maps directly to the loki `compactor.retention_enabled` configuration option, which is set to `false` when no retention period is defined.
-        Specifying retention periods for individual streams is not currently supported.
-
-        Ref: https://grafana.com/docs/loki/latest/operations/storage/retention/
+        Sets the global log retention period in Loki (in days).
+        A value of `0` disables deletion, causing logs to be retained indefinitely.
+        The minimum supported non-zero retention period is `1d` (24 hours).
+        Individual/per-stream retention rules are not supported by Loki at this time.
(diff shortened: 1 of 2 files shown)


charmcraft.yaml
```diff
@@ -212,15 +212,13 @@ config:
...
-
-        Ref: https://grafana.com/docs/loki/latest/operations/storage/retention/
+        Sets the global log retention period in Loki (in days).
+        A value of `0` disables deletion, causing logs to be retained indefinitely.
+        The minimum supported non-zero retention period is `1d` (24 hours).
+        Individual/per-stream retention rules are not supported by Loki at this time.
+        For more info, see the Loki docs: https://grafana.com/docs/loki/latest/operations/storage/retention/
       type: int
-      default: 0
+      default: 15
```
sinapah, line 221: Hi @sed-i,

Initially I wanted to set this to 30 as we discussed. I looked at Prometheus and saw that it set the default to 15 days. Also, in the recent examples where users have reached out with retention-related questions, they always set retention to either two weeks or just 15 days.

I think this should be a reasonable default, but happy to change it to 30 if you prefer that.

lucabello (approved):
Approving, but please double check with Leon the default is what we want; also, do we need to replicate this for the Loki HA charms as well? For COS, I'd rather Loki HA match the behavior of Mimir, so can you double check what's the behavior there?

